PR #41, opened 2026-04-11
feat(eval): support reading script from stdin

sandorfr:
Hello,

Thanks for making and shring this tool incredibly helpful in the absence of playwright compatibility. 

One thing I noticed while trying this out is agents struggle/stumble with escaping I found that adding stdin support to the eval command (and instructing to use it in my skill) really made a difference in the success rate.

I'm new to rust so this PR may not be the best approach, but decided to share it at least for the concept.

## Summary

- `eval` now accepts `-` or no argument to read the JavaScript from stdin
- Avoids shell escaping issues for scripts containing double quotes or other special characters (e.g. `document.querySelector('[data-id="main"]').textContent`)
- Inline script argument still works as before

## Test plan

- [ ] `test_parse_eval_with_script_containing_quotes` — inline script with embedded double quotes parses correctly
- [ ] `test_parse_eval_dash_reads_stdin` — `-` sentinel is preserved through CLI parsing
- [ ] `test_parse_eval_no_arg_reads_stdin` — omitting the argument yields `None`, triggering stdin read

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * The eval command’s script argument is now optional: pass a script inline, use `-` to indicate stdin, or omit the argument to read from stdin.
* **Bug Fixes**
  * Reading a script from stdin now rejects empty or whitespace-only input.
* **Tests**
  * Added unit tests covering CLI parsing (including quotes and `-`) and stdin reading behavior.
* **Documentation**
  * Changelog updated to document stdin support for eval.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `eval` command now reads the script from stdin when the argument is `-` or omitted ([#41])
+
 ## [0.3.0] - 2026-04-10
 
 ### Added
@@ -130,6 +134,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#11]: https://github.com/mpiton/tauri-pilot/issues/11
 [#10]: https://github.com/mpiton/tauri-pilot/issues/10
 [#37]: https://github.com/mpiton/tauri-pilot/issues/37
+[#41]: https://github.com/mpiton/tauri-pilot/pull/41
 [Unreleased]: https://github.com/mpiton/tauri-pilot/compare/v0.3.0...HEAD
 [0.2.1]: https://github.com/mpiton/tauri-pilot/compare/v0.2.0...v0.2.1
 [0.2.0]: https://github.com/mpiton/tauri-pilot/compare/v0.1.0...v0.2.0
--- a/crates/tauri-pilot-cli/src/cli.rs
+++ b/crates/tauri-pilot-cli/src/cli.rs
@@ -98,7 +98,11 @@ pub(crate) enum Command {
     /// Get all attributes of an element.
     Attrs { target: String },
     /// Evaluate arbitrary JavaScript.
-    Eval { script: String },
+    /// Pass `-` as the script or omit it to read from stdin.
+    Eval {
+        /// JavaScript to evaluate. Use `-` or omit to read from stdin.
+        script: Option<String>,
+    },
     /// Invoke a Tauri IPC command.
     Ipc {
         command: String,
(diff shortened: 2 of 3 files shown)


mpiton:
Hey, thanks for the PR and for bringing up the use case. The shell-escaping pain on scripts containing `"` is something I ran into myself as soon as I started driving tauri-pilot from an agent, so the problem is very real and the approach is the right one. Nice reflex taking an `impl Read` in `read_script` instead of calling `stdin()` directly, it makes the function properly testable. Clean work for a first Rust PR.

Three things to sort out before we can merge:

**1. `cargo fmt`**

That's the only reason CI is red right now. `rustfmt` wants to collapse the `Cli::parse_from([...])` arrays onto a single line and break `read_to_string` + `ensure!` into multi-line form. A `cargo fmt --all` and one more commit and you're good.

**2. TTY detection**

This one is what bothers me the most. If someone runs `tauri-pilot eval` with no argument and nothing piped in, your `stdin().read_to_string()` will hang silently waiting for keyboard input. For a tool built to be driven by agents, that's a trap you can't diagnose from the other end of the pipe. We need to refuse it explicitly:

```rust
use std::io::IsTerminal;

let script = match script.as_deref() {
    None | Some("-") => {
        anyhow::ensure!(
            !std::io::stdin().is_terminal(),
            "stdin is a terminal: pass a script as argument or pipe it in"
        );
        read_script(&mut std::io::stdin())?
    }
    Some(s) => s.to_owned(),
};
```

`IsTerminal` has been stable since 1.70 so it's fine with the project MSRV (1.94).

**3. `CHANGELOG.md`**

The project follows [Keep a Changelog](https://keepachangelog.com/), every PR has to add a line under `[Unreleased]`. For this one, under `### Added`:

```
- `eval` command now reads the script from stdin when the argument is `-` or omitted (#41)
```

---

Optional nit, not blocking: an `assert_cmd` integration test that pipes a script into `tauri-pilot eval -` against a mock server would be ideal. The current tests cover `read_script` and the clap parsing separately, the glue between them is the missing piece. If you don't feel like wrestling with the mock socket, I'm happy to add it as a follow-up.

The `-` convention for stdin is the right call, and the `trim().is_empty()` check is good too: you're not trimming the script itself, so any internal indentation is preserved, which is what we want for JS.

One last thing, unrelated to your PR: you may have noticed the "Expected — Waiting for status to be reported" check stuck on every recent PR. That was a branch protection looking for `Check & Test` while the Actions matrix actually publishes `Check & Test (ubuntu-latest)` and `Check & Test (macos-latest)`. I just added a `CI Success` umbrella job on main and updated the protection, your branch will pick it up as soon as you rebase.

sandorfr:
That a very good point, hanging silently in agent workflows (or any other controlling program) is problematic and the worse type of problem to debug.

sandorfr:
@mpiton you have a linter rule with a maximum  line per function. This prompts towards refactoring `run_dom_command` I thinks each command can be their own function or at least the longer ones (Scroll, Eval, Drag). But this could probably be taken further by splitting it to individual files. 

Probably a refactoring to be done independently, any preferred approach?

mpiton:
Good catch on the linter rule. `run_dom_command` has definitely grown too large.

Let's keep this PR strictly focused on the `stdin` feature though. Get the `fmt`, `is_terminal` check, and `CHANGELOG` updates pushed here, and we'll merge this first.

For the refactoring, a follow-up PR would be great. My preferred approach: keep it in the same file but extract the heavier branches (`Eval`, `Scroll`, `Drag`) into private handler functions (e.g., `handle_eval(...)`). If the file is still too huge after that, we can look at splitting it into separate modules. Let's start with just breaking down the `match` block.

Let me know when this branch is ready.

sandorfr:
all the changes should be in @mpiton. I only extracted eval in this PR to handle_eval so the line count is back under 100.

sandorfr:
I think maybe we should update your default skill to mention using heredoc. This is what's in my skill.


## eval

**Always use heredoc form** — it bypasses all shell escaping, works with any JS including special characters, Panda CSS class names (e.g. `h_100%`), and multi-line scripts:

```bash
tauri-pilot eval - <<'EOF'
(function() {
  var el = document.querySelector('.h_100%');
  if (!el) return 'not found';
  var cs = getComputedStyle(el);
  return [cs.height, cs.width, cs.backgroundColor].join(' | ');
})()
EOF

# one-liner via pipe
echo 'document.title' | tauri-pilot eval -
```

mpiton:
Good call. `<<'EOF'` in particular isn't obvious -- people will reach for double-quoting and still get burned by `$` in selectors. Worth adding to the README and the default skill. Opened #43 to track it.

mpiton:
Thanks for your contribution !

PR #52, opened 2026-04-15
fix: Dispatch pointer events for clicks

### LINKED ISSUE #47: click doesn't open Radix UI components (DropdownMenu, Select, Dialog)
## Description

`tauri-pilot click` on Radix UI trigger elements (DropdownMenuTrigger, SelectTrigger, DialogTrigger) does not open the associated popover/dropdown/dialog.

## Steps to Reproduce

```bash
# Click a Radix DropdownMenuTrigger button
tauri-pilot click '[aria-label="More actions"]'
# Result: ✓ ok (click registered), but no dropdown appears

# Same with snapshot verification:
tauri-pilot snapshot -i
# No menu items visible in the tree
```

## Expected Behavior

The dropdown/select/dialog should open, with its content appearing in the DOM and visible in `snapshot`.

## Actual Behavior

The click event fires (returns `✓ ok`) but Radix doesn't react — the component stays closed.

## Root Cause Hypothesis

Radix UI components rely on a specific pointer event sequence: `pointerdown → pointerup → click`. A simple `click()` or `dispatchEvent(new MouseEvent('click'))` may not be sufficient. Radix explicitly listens for `pointerdown` to initiate opening.

## Possible Fix

When clicking an element, dispatch the full pointer event sequence:
```javascript
element.dispatchEvent(new PointerEvent('pointerdown', { bubbles: true }));
element.dispatchEvent(new PointerEvent('pointerup', { bubbles: true }));
element.dispatchEvent(new MouseEvent('click', { bubbles: true }));
```

Or better yet, use the native webview click simulation at the coordinate level rather than synthetic DOM events.

## Impact

All Radix-based interactive components (menus, selects, dialogs, popovers) are untestable via tauri-pilot, which is a major gap for QA automation of any app using shadcn/ui.

chainyo:
Fixes #47.

The bridge click implementation now dispatches a browser-like pointer and mouse sequence for Radix UI triggers. It emits pointerdown before the compatibility mouse events, suppresses mousedown and mouseup when pointerdown is canceled, and dispatches click through the pointer event helper so listeners can observe pointer metadata.

This also keeps MCP auto-detected socket pinning stable by preferring sockets in XDG_RUNTIME_DIR before falling back to /tmp. The Ubuntu CI failure was caused by auto-detection choosing an unrelated newer /tmp socket instead of the test socket under XDG_RUNTIME_DIR.

Regression coverage now verifies the pointer event order, pointer metadata, compatibility mouse-event gating, pointer click dispatch, and XDG_RUNTIME_DIR socket precedence.

Validation run locally: cargo fmt --all --check, git diff --check, cargo test --workspace, and cargo clippy --workspace -- -D warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed click event dispatch order to emit pointer events before mouse events, resolving issues where Radix UI dropdowns, selects, and dialogs failed to open correctly.
  * Added fallback support for pointer events in less capable environments.

* **Tests**
  * Added test coverage to verify click event sequence ordering and compatibility.

---
## Summary by cubic
Fixes broken clicks on Radix UI components by dispatching a browser-like pointer→mouse sequence and emitting click as a pointer event. Also stabilizes MCP socket auto-detection by preferring `XDG_RUNTIME_DIR` sockets over `/tmp`.

- **Bug Fixes**
  - Click bridge now dispatches pointerdown → (if not canceled) mousedown → pointerup → (if not canceled) mouseup → click as a pointer event with pointer metadata; uses element-center coordinates by default, focuses after mousedown if not canceled, and falls back when `PointerEvent` is unavailable.
  - CLI selects the newest socket in `XDG_RUNTIME_DIR` first, then falls back to `/tmp`, avoiding accidental pinning to unrelated newer `/tmp` sockets in CI.

<sup>Written for commit a9ce0f36f18ba5d0c2adc0da802e4d73954643f0. Summary will update on new commits.</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Stdin heredoc and pipe examples for `eval -` in README, SKILL.md, and CLI reference ([#50])
 - MCP server mode for exposing tauri-pilot commands as structured tools over stdio ([#51])
 
+### Fixed
+
+- `click` now dispatches pointer events before mouse events so Radix UI dropdown, select, and dialog triggers open correctly ([#52])
+
 ## [0.3.0] - 2026-04-10
 
 ### Added
@@ -139,6 +143,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#41]: https://github.com/mpiton/tauri-pilot/pull/41
 [#50]: https://github.com/mpiton/tauri-pilot/pull/50
 [#51]: https://github.com/mpiton/tauri-pilot/pull/51
+[#52]: https://github.com/mpiton/tauri-pilot/pull/52
 [Unreleased]: https://github.com/mpiton/tauri-pilot/compare/v0.3.0...HEAD
 [0.2.1]: https://github.com/mpiton/tauri-pilot/compare/v0.2.0...v0.2.1
 [0.2.0]: https://github.com/mpiton/tauri-pilot/compare/v0.1.0...v0.2.0
--- a/crates/tauri-pilot-cli/src/main.rs
+++ b/crates/tauri-pilot-cli/src/main.rs
@@ -12,7 +12,7 @@ use base64::Engine;
 use clap::Parser;
 use serde_json::{Value, json};
 use std::io::IsTerminal;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use cli::{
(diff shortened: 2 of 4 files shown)


mpiton (commented):
The pointer event sequence looks right. I've hit this same Radix UI issue before - dispatching raw `MouseEvent` skips the internal state checks those components rely on, so the fix makes sense.

One thing that's going to bite someone eventually: the assertions in `lib.rs` test for exact whitespace, including `\n      `. If `bridge.js` gets reformatted (Prettier, different indent setting, anything), those tests fail for the wrong reason. The ordering check by string index is a reasonable proxy for runtime dispatch order, but I'd at least strip whitespace before asserting, or just test the logical constraint without depending on source layout:

```rust
let gated = js.contains("if (pointerDownOk)");
let mousedown_after_gate = js.rfind("mousedown").unwrap() 
    > js.find("if (pointerDownOk)").unwrap();
assert!(gated && mousedown_after_gate, "mousedown must be gated on pointerdown");
```

Also noticed `el.focus()` moved from before `mousedown` to after it. That's more spec-correct, but worth calling out since it's a behavior change - some components might depend on the original order. Is that intentional?

The `resolve_socket` refactor is cleaner than what it replaced. Early return on XDG makes the precedence impossible to misread.

LGTM with the test fragility as a non-blocker. The Radix fix was overdue and the CI socket issue was annoying to debug, so thanks for tracking both down.

PR #56, opened 2026-04-17
fix(cli): treat undefined eval results as success (#48)

### LINKED ISSUE #48: eval returns exit code 1 when JS expression evaluates to undefined/void
## Description

When `tauri-pilot eval` executes a JS expression that returns `undefined` (e.g., `element.click()`, void functions), it exits with code 1 and prints:

```
Error: Server returned empty result without error
```

## Steps to Reproduce

```bash
$ tauri-pilot eval "document.querySelector('a')?.click()"
# Exit code: 1
# Error: Server returned empty result without error

# But the click DID work — navigation happened:
$ tauri-pilot state
# URL changed to the link's href
```

## Expected Behavior

An expression returning `undefined` should be treated as success (exit code 0), not an error. Alternatively, print nothing and exit cleanly.

## Actual Behavior

Exit code 1 + error message. This breaks bash scripts using `&&` chains or `set -e`, causing downstream commands to be skipped even though the eval succeeded.

## Workaround

Append a fallback return value:
```bash
tauri-pilot eval "document.querySelector('a')?.click() || 'done'"
```

Or redirect stderr:
```bash
tauri-pilot eval "document.querySelector('a')?.click()" 2>/dev/null
```

## Suggested Fix

If the webview returns `null`/`undefined`/empty, treat it as success with empty output rather than an error.

mpiton:
## Summary

Fixes #48. `tauri-pilot eval "document.querySelector('a')?.click()"` no longer exits with code 1 and `Error: Server returned empty result without error` when the JS expression returns `undefined`. Void expressions now exit 0 with empty output, so they compose cleanly with bash `&&` chains and `set -e`.

## Root cause

Six-layer chain: JS `undefined` → `JSON.stringify(undefined) === undefined` → Tauri omits the `result` field → handler resolves with `Value::Null` → plugin serializes `{"result": null}` → serde on the client side conflates `"result": null` with an absent field for `Option<Value>` → client bails with "empty result without error".

## Changes

- **`crates/tauri-pilot-cli/src/client.rs`**: A missing or null `result` is now treated as success with `Value::Null` (the CLI's existing `format_text` already handles `Value::Null` as empty output). Two regression tests cover `"result": null` and the absent-field shape.
- **`crates/tauri-plugin-pilot/src/eval.rs`**: `wrap_script` now normalizes `undefined` to the string `'null'` before stringifying, so Tauri keeps the `result` field populated and the handler no longer logs `callback received with neither result nor error` on every void eval.
- **`crates/tauri-plugin-pilot/src/handler.rs`**: Adds an integration-level test that exercises the full `handle_callback(Some("null"), None)` → `Value::Null` round-trip.
- **`CHANGELOG.md`**: `[Unreleased] > Fixed` entry with `[#48]` reference.
- **`docs/src/content/docs/reference/cli.md`**: Short note that void expressions print nothing and exit 0, with a `&&` example.

## Test plan

- [x] `cargo test --workspace` → 230 pass (1 new)
- [x] RED test confirmed failing before the fix, GREEN after
- [x] `cargo clippy --workspace -- -D warnings` → no new warnings (4 pre-existing warnings on main)
- [x] Verified end-to-end in `pilot-test-app`:
  - `eval "document.querySelector('#login-form button')?.click()"` → exit 0
  - `eval "void 0"` → exit 0
  - `eval "1+1"` → prints `2`, exit 0
  - `eval "undefined"` → exit 0
  - `eval "<void expr>" && tauri-pilot title` → chain succeeds
- [x] Adversarial code + rust reviews — verdict APPROVE, medium/low findings addressed

Closes #48

---
## Summary by cubic
Fixes `tauri-pilot eval` exiting with code 1 when a JS expression returns `undefined`. Void expressions now print nothing and exit 0, so bash `&&` and `set -e` work as expected (fixes #48).

- **Bug Fixes**
  - `crates/tauri-pilot-cli`: Treat missing or `null` `result` as success (`Value::Null`). Adds tests for both response shapes.
  - `crates/tauri-plugin-pilot` (`eval`): Normalize `undefined` to the string `'null'` before stringifying to keep the `result` field. Adds a wrapper test and a handler round-trip test.
  - Docs/Changelog: Note the new behavior with a `&&` example and add the changelog entry.

<sup>Written for commit 1596bd97af58798377782b840aad1ab1bdcf278c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * `eval` now treats JavaScript expressions that yield `undefined`/void as success (exit code 0) instead of an error, preserving shell `&&`/`set -e` behavior.

* **Documentation**
  * Updated CLI docs with guidance and examples for undefined-returning expressions.

* **Tests**
  * Added tests verifying null/missing results are handled as successful responses.

* **Chores**
  * Added an unreleased changelog entry documenting this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `press` with `--window <label>` but no focus hook installed now errors instead of silently injecting into whatever window has focus ([#53])
 - `Enigo::new` failure hint about macOS Accessibility permission is now gated to macOS builds — Linux and Windows errors no longer point users at the wrong remediation ([#53])
 - `press` JoinError handling distinguishes panics from cancellation and runtime-shutdown cases instead of reporting every failure as "panicked" ([#53])
+- `eval` now exits with code 0 when the JS expression returns `undefined` (e.g. `element.click()`, void functions). Previously the CLI bailed with `Error: Server returned empty result without error`, breaking bash `&&` chains and `set -e` scripts even though the eval had succeeded ([#48])
 
 ## [0.3.0] - 2026-04-10
 
@@ -159,6 +160,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#37]: https://github.com/mpiton/tauri-pilot/issues/37
 [#41]: https://github.com/mpiton/tauri-pilot/pull/41
 [#46]: https://github.com/mpiton/tauri-pilot/issues/46
+[#48]: https://github.com/mpiton/tauri-pilot/issues/48
 [#50]: https://github.com/mpiton/tauri-pilot/pull/50
 [#51]: https://github.com/mpiton/tauri-pilot/pull/51
 [#52]: https://github.com/mpiton/tauri-pilot/pull/52
--- a/crates/tauri-pilot-cli/src/client.rs
+++ b/crates/tauri-pilot-cli/src/client.rs
@@ -65,9 +65,12 @@ impl Client {
             bail!("RPC error ({}): {}", err.code, err.message);
         }
 
-        response
-            .result
-            .context("Server returned empty result without error")
+        // A missing `result` field (or explicit `"result": null`) means the
+        // server-side script completed successfully but produced no value —
+        // e.g., `element.click()` or any void expression. Treat this as
+        // success with Value::Null rather than an error so bash `&&` chains
+        // and `set -e` keep working. See #48.
+        Ok(response.result.unwrap_or(serde_json::Value::Null))
     }
 }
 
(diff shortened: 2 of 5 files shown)


mpiton:
Re: the CodeRabbit suggestion to scope the empty-result fallback to `eval` only — I checked the bridge and several non-eval handlers legitimately return `null`: `html()` when the selector misses, `text()` on missing elements, the `stack`/`url` helpers in the logger path, etc. (see `crates/tauri-plugin-pilot/js/bridge.js` lines 54-76, 269, 729). Scoping the fallback to `eval` would turn those legitimate null returns into `Server returned empty result without error` failures.

Separately: the plugin also cannot produce a malformed missing-`result` response today — `handle_callback`'s final `else` always resolves `Ok(Value::Null)`, so the client would never see an unfilled `result` unless someone later breaks that invariant. If that ever happens it's a server bug worth catching, but the cost/benefit right now favours keeping the fallback unscoped.

PR #64, opened 2026-04-21
feat: Windows IPC support via Named Pipes

### LINKED ISSUE #44: feat: Windows support — Named Pipe IPC transport
## Motivation

tauri-pilot currently runs on Linux and macOS only. The plugin is compiled under `#[cfg(unix)]` and the entire IPC layer uses Unix domain sockets — which don't exist on Windows.

Adding Windows support would let developers use tauri-pilot to automate and test Tauri apps on the third platform Tauri itself supports.

### 1. IPC transport: Unix sockets → Named Pipes

The server in `crates/tauri-plugin-pilot/src/server.rs` binds a `std::os::unix::net::UnixListener`. On Windows, the equivalent is a Named Pipe.

tokio already exposes `tokio::net::windows::named_pipe::ServerOptions` / `NamedPipeServer` for async Named Pipe I/O. The path convention would change from `/tmp/tauri-pilot-{identifier}.sock` to `\\.\pipe\tauri-pilot-{identifier}`.

The CLI's `resolve_socket()` would similarly need to enumerate Named Pipes instead of scanning `$XDG_RUNTIME_DIR` / `/tmp`.

### 2. Security hardening: Unix permissions → Windows ACLs

The current hardening stack (`0o600` file permissions, `umask`, peer UID check via `SO_PEERCRED`) is Unix-specific. Windows Named Pipes have their own security model via Security Descriptors:

- Restrict the pipe to the creating user's SID (equivalent of `0o600`)
- Validate client identity via `GetNamedPipeClientProcessId` + process token lookup (equivalent of peer UID check)

`std::os::unix::fs::PermissionsExt` and `std::os::unix::io::AsRawFd` usages in `server.rs` must be replaced or `#[cfg]`-gated.

### 3. Plugin `#[cfg(unix)]` gates → `#[cfg(any(unix, windows))]`

`crates/tauri-plugin-pilot/src/lib.rs` wraps everything in `#[cfg(unix)]`. This needs to become `#[cfg(any(unix, windows))]` (or equivalent) once the platform-specific transport is abstracted.

### 4. CI matrix

`.github/workflows/ci.yml` currently tests `ubuntu-latest` and `macos-latest`. `windows-latest` would need to be added, along with any Windows-specific system dependencies (likely none beyond Rust itself for a pure-Rust plugin, but Tauri's WebView2 dependency may affect build checks).

## Suggested implementation path

1. Extract an `IpcListener` / `IpcStream` trait (or enum) that abstracts Unix socket vs. Named Pipe — keeps the JSON-RPC dispatch logic platform-agnostic.
2. Implement the Windows Named Pipe backend behind `#[cfg(windows)]`.
3. Port the security hardening to Windows ACLs / pipe security descriptors.
4. Lift the `#[cfg(unix)]` gates in the plugin once the transport abstraction compiles on Windows.
5. Add `windows-latest` to the CI matrix.

## Acceptance criteria

- [ ] `cargo build --workspace` succeeds on Windows
- [ ] `cargo test --workspace` passes on Windows
- [ ] `cargo clippy --workspace -- -D warnings` passes on Windows
- [ ] Named Pipe is restricted to the creating user (no other local users can connect)
- [ ] CI matrix includes `windows-latest`
- [ ] README and docs updated to list Windows as a supported platform

## Notes

- The JS bridge (`bridge.js`) is injected via `webview.eval()` and is platform-agnostic — no changes needed there.
- The JSON-RPC protocol itself is also platform-agnostic; only the transport layer changes.
- macOS was added in [#37] as a CI verification step (macOS also uses Unix sockets, so the change was minimal). Windows is the bigger lift because of the IPC transport difference.

espetro:
## Summary

Add Windows support for tauri-pilot using Named Pipe IPC transport, replacing Unix domain sockets on Windows while keeping full Linux/macOS compatibility.

## What changed

| Platform | Transport | Discovery | Security |
|----------|-----------|-----------|----------|
| Linux/macOS | Unix domain socket (unchanged) | XDG_RUNTIME_DIR → /tmp, mtime+uid | 0o600 + peer_cred |
| Windows | Named Pipe (`\\\\.\\pipe\\tauri-pilot-{id}`) | %LOCALAPPDATA%\\tauri-pilot\\instances.json | User-only DACL |

### Files changed
- `crates/tauri-plugin-pilot/src/server/` — extracted to `mod.rs` + `unix.rs` + `windows.rs` (+DACL)
- `crates/tauri-pilot-cli/src/client/` — extracted to `mod.rs` + `unix.rs` + `windows.rs`
- `crates/tauri-pilot-cli/src/main.rs` — `resolve_socket()` with Windows registry branch
- `crates/tauri-plugin-pilot/src/lib.rs` — `#[cfg(any(unix, windows))]` gates
- `crates/*/Cargo.toml` — platform-gated `libc`, `windows` crate with features
- `.github/workflows/ci.yml` — `windows-latest` added
- `README.md` — Windows badge + WebView2 mention

### Key implementation details

**User-only DACL** (`server/windows.rs`):
- `create_user_only_security_attributes()` opens process token, gets user SID, creates ACL with single ACE granting `GENERIC_READ | GENERIC_WRITE`, attaches to security descriptor
- Passed to `ServerOptions::create_with_security_attributes_raw()`

**Registry-based discovery** (`server/windows.rs` + `main.rs`):
- Server writes `{pipe, pid, created_at}` to `%LOCALAPPDATA%\\tauri-pilot\\instances.json` on bind
- Server removes entry on drop via `RegistryGuard`
- CLI reads registry to find correct pipe for multi-instance support

**Named Pipe accept loop** (`server/windows.rs`):
- Uses `connect() → create next → spawn` pattern (different from Unix `accept()`)
- Each connection spawns a tokio task with shared `Arc<Ctx>`

## Testing

- `cargo check --workspace` passes on rustc 1.95.0
- `cargo fmt --all -- --check` passes
- Platform-specific tests in `server/windows.rs` and `client/windows.rs`
- CI matrix: `ubuntu-latest`, `macos-latest`, `windows-latest`

## Known items for review

1. **Commit scopes**: Use `feat(windows)` instead of `feat(plugin)`/`feat(cli)` — can amend if strict CONTRIBUTING.md compliance needed
2. **Module sizes**: `server/windows.rs` (399 lines) and `server/unix.rs` (339 lines) exceed 150-line guideline — inherited from original `server.rs` (455 lines)
3. **Function size**: `create_user_only_security_attributes()` is ~65 lines (limit: 50)

Closes #44

## Summary by CodeRabbit

* **New Features**
  * Windows support: named‑pipe IPC added so plugin and CLI work on Windows and Unix.

* **CI/Testing**
  * CI check matrix now runs on Windows; platform‑gated and more reliable tests for parallel runs.

* **Documentation**
  * README, badges, and architecture docs updated to include Windows/WebView2 and IPC changes.

* **Bug Fixes / Security**
  * Windows IPC security hardening (access controls/SID checks) and various platform correctness fixes.

* **Other**
  * Protocol ID widened to accept JSON values; manifests/dependencies adjusted for multi‑platform builds.

---
## Summary by cubic
Adds Windows IPC via Named Pipes with strict per‑user security and per‑instance discovery, keeping Unix sockets on Linux/macOS. The server core is now shared across platforms; CLI, tests, CI, and docs updated.

- New Features
  - Windows transport: Named Pipes at \\.\pipe\tauri-pilot-{id} with a user‑only DACL, per‑connection client SID validation, and a 1MB request line cap.
  - Discovery: per‑instance files at %LOCALAPPDATA%\tauri-pilot\instances\*.json; the CLI picks the newest alive instance.
  - Shared server core in server/mod.rs; platform servers/clients in unix.rs and windows.rs (Windows accept loop: connect → create next → spawn with retry/backoff).

- Bug Fixes
  - Protocol/server: JSON‑RPC id is now a JSON value; parse errors return id:null; per‑line reads are hard‑capped to prevent OOM on newline‑less floods.
  - Windows security: fixed ACL allocation overflow and SID lifetime/use‑after‑free; peer SID check uses the thread token; user‑only DACL applied to every pipe instance with no default‑DACL fallback; accept loop handles transient creation errors with retry/backoff; added DACL regression test.
  - Windows API: corrected `windows` 0.52 API paths and feature flags, free security descriptors via LocalFree, compare to ERROR_SUCCESS, handle CloseHandle properly, and fix STILL_ACTIVE cast in PID checks.
  - Tests/CI/build/docs: stabilized Windows tests (`serial_test`, unique pipe/socket names, CRLF‑safe assertions); cfg‑gated deps (`libc` on Unix, `windows` on Windows); Linux `enigo` marked `optional = true`; CI adds windows‑latest; README mentions Windows/WebView2 and updates the architecture diagram; plugin description now “IPC JSON‑RPC”; `init()` docs cover Windows Named Pipe behavior.

<sup>Written for commit 2de6e6b99141ba31db8c7972d5237705cc725fe8. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v4
 
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,9 @@ Cargo.lock
 .DS_Store
 Thumbs.db
 
-# Claude Code
+# AI Agents
 .claude/
+.sisyphus/
 CLAUDE.md
 
 # Node (for bridge build step)
--- a/.gitignore
+++ b/.gitignore
@@ -16,10 +16,6 @@ Thumbs.db
 
 # AI Agents
 .claude/
-.sisyphus/
-.agents/
-skills-lock.json
-CLAUDE.md
 
 # Node (for bridge build step)
 node_modules/
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@ Thumbs.db
 # AI Agents
 .claude/
 .sisyphus/
+.agents/
+skills-lock.json
 CLAUDE.md
 
 # Node (for bridge build step)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Windows support** — named pipe server and client for Windows, with security hardening (DACL, SID validation), registry-based instance discovery, and platform-specific tests ([#64])
+
 ## [0.4.0] - 2026-04-17
 
 ### Added
@@ -189,3 +193,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#8]: https://github.com/mpiton/tauri-pilot/issues/8
 [#17]: https://github.com/mpiton/tauri-pilot/pull/17
 [#31]: https://github.com/mpiton/tauri-pilot/issues/31
+[#64]: https://github.com/mpiton/tauri-pilot/pull/64
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **Windows security hardening** — fixed heap overflow in ACL allocation, use-after-free on the SID buffer, a no-op peer-SID check (`OpenProcessToken` on the impersonated thread replaced with `OpenThreadToken`), UB on the alloc-failure path, silent fallback to a broader default DACL, and a permanently-aborting accept loop. Also switched `instances_dir` creation to `std::fs::create_dir_all` for correctness on clean profiles, and added a regression test asserting the bound pipe carries a user-only DACL with one ACE ([#64])
+- `tauri-pilot-cli` Windows builds now enable the `Win32_Foundation` and `Win32_System_Threading` features on the `windows` crate so `is_pid_alive` compiles on Windows CI ([#64])
+- `tauri-plugin-pilot` marks the Linux-target `enigo` entry as `optional = true` so `--no-default-features` actually drops the dependency — previously the `press` feature gate was silently defeated by cargo merging the target-specific entry with the top-level one ([#64])
+- `tauri-plugin-pilot` server now caps per-line reads using `AsyncReadExt::take` before invoking `read_line`, so a peer flooding bytes without a newline can no longer OOM the host process — the existing `MAX_LINE_LENGTH` check was applied after the full line had already been buffered ([#64])
+- `tauri-pilot-cli` Unix client tests use a per-process, atomic-counter socket path instead of hard-coded `/tmp/tauri-pilot-test-*.sock` paths, so parallel `cargo test` runs no longer cross-wire through the same socket file ([#64])
+- `tauri-pilot-cli` Windows registry-resolution tests mock entries with `std::process::id()` instead of a fabricated dead PID, so the liveness filter added in the Windows support work doesn't skip them ([#64])
+
+### Changed
+
+- `tauri-plugin-pilot` `init()` doc comment clarifies the no-op fallback now excludes Windows too and mentions the Named Pipe server path ([#64])
+
+### Removed
+
+- `.gitignore` no longer ignores `CLAUDE.md`, `.sisyphus/`, `.agents/`, or `skills-lock.json` — these are personal tooling entries that belong in a user-level `~/.gitignore_global`, not in the repo ([#64])
 
 ## [0.4.0] - 2026-04-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Windows support** — named pipe server and client for Windows, with security hardening (DACL, SID validation), registry-based instance discovery, and platform-specific tests ([#64])
 
+### Fixed
+
+- **Windows security hardening** — fixed heap overflow in ACL allocation, use-after-free on the SID buffer, a no-op peer-SID check (`OpenProcessToken` on the impersonated thread replaced with `OpenThreadToken`), UB on the alloc-failure path, silent fallback to a broader default DACL, and a permanently-aborting accept loop. Also switched `instances_dir` creation to `std::fs::create_dir_all` for correctness on clean profiles, and added a regression test asserting the bound pipe carries a user-only DACL with one ACE ([#64])
+
 ## [0.4.0] - 2026-04-17
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
   <a href="https://crates.io/crates/tauri-plugin-pilot"><img src="https://img.shields.io/crates/v/tauri-plugin-pilot.svg" alt="crates.io"></a>
   <a href="https://github.com/mpiton/tauri-pilot/blob/main/LICENSE"><img src="https://img.shields.io/badge/license-MIT-blue.svg" alt="License: MIT"></a>
   <img src="https://img.shields.io/badge/rust-1.95.0+-orange.svg" alt="Rust 1.95.0+">
-  <img src="https://img.shields.io/badge/platform-linux%20%7C%20macOS-lightgrey.svg" alt="Platform: Linux | macOS">
+  <img src="https://img.shields.io/badge/platform-linux%20%7C%20macOS%20%7C%20Windows-lightgrey.svg" alt="Platform: Linux | macOS | Windows">
   <img src="https://img.shields.io/badge/tauri-v2-24C8D8.svg" alt="Tauri v2">
 </p>
 
@@ -38,24 +38,24 @@ ok
 
 ## Why?
 
-There's no tool for AI agents to interact with Tauri app UIs. Playwright doesn't work (Tauri uses system WebViews — WebKitGTK on Linux, WebKit on macOS — not Chromium). tauri-pilot bridges this gap with a lightweight plugin + CLI that speaks a protocol optimized for LLM consumption.
+There's no tool for AI agents to interact with Tauri app UIs. Playwright doesn't work (Tauri uses system WebViews — WebKitGTK on Linux, WebKit on macOS, WebView2 on Windows — not Chromium). tauri-pilot bridges this gap with a lightweight plugin + CLI that speaks a protocol optimized for LLM consumption.
 
 ## How it works
 
 ```
-┌──────────────┐   Unix Socket    ┌─────────────────────────────┐
-│  tauri-pilot  │ ◄──────────────► │  tauri-plugin-pilot (Rust)  │
-│  (CLI)        │   JSON-RPC       │  embedded in your app       │
-└──────────────┘                   │                             │
-                                   │  ┌─────────────────────┐   │
-                                   │  │  JS Bridge (injected)│   │
-                                   │  │  window.__PILOT__    │   │
-                                   │  └─────────────────────┘   │
-                                   │  WebView                    │
-                                   └─────────────────────────────┘
+┌──────────────┐   Unix Socket /     ┌─────────────────────────────┐
+│  tauri-pilot  │   Named Pipe (Win) │  tauri-plugin-pilot (Rust)  │
+│  (CLI)        │ ◄─────────────────► │  embedded in your app       │
+└──────────────┘   JSON-RPC           │                             │
+                                     │  ┌─────────────────────┐   │
+                                     │  │  JS Bridge (injected)│   │
+                                     │  │  window.__PILOT__    │   │
+                                     │  └─────────────────────┘   │
+                                     │  WebView                    │
+                                     └─────────────────────────────┘
 ```
 
-1. **Plugin** embeds in your Tauri app (debug builds only), starts a Unix socket server
+1. **Plugin** embeds in your Tauri app (debug builds only), starts a Unix socket / Named Pipe server
 2. **CLI** connects to the socket, sends JSON-RPC commands
 3. **JS Bridge** injected into the WebView handles DOM inspection and interaction
 
@@ -222,7 +222,7 @@ window:
 
 ## Requirements
 
-- **Linux** (WebKitGTK) or **macOS** (WebKit) — Windows planned
+- **Linux** (WebKitGTK), **macOS** (WebKit), or **Windows** (WebView2)
 - **Tauri v2** (v1 not supported)
 - **Rust 1.95.0+** (edition 2024)
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ There's no tool for AI agents to interact with Tauri app UIs. Playwright doesn't
 
 ## How it works
 
-```
+```text
 ┌──────────────┐   Unix Socket /     ┌─────────────────────────────┐
 │  tauri-pilot  │   Named Pipe (Win) │  tauri-plugin-pilot (Rust)  │
 │  (CLI)        │ ◄─────────────────► │  embedded in your app       │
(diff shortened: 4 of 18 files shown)


mpiton (changes requested):
Hey @espetro, thanks for the PR. Windows support via Named Pipes is a big chunk of work, the `server/{mod,unix,windows}` split is clean and the 3-OS CI is green. Got a few blockers before merge though.

**Two security issues on the Windows server:**

1. The "user-only" DACL is only applied to the first pipe instance. In `server/windows.rs:287-289`, the next instance is created via `ServerOptions::new().create(path)` with no security attributes. On Windows each Named Pipe instance has its own security descriptor, and without SA the default DACL grants read to Everyone (MSDN docs on `CreateNamedPipe`). Result: from the 2nd connection onward, any local user can connect. Kills the main promise of the PR.

Fix: pass the SA on every iteration of the accept loop. The kernel copies the SD at `CreateNamedPipe` time so rebuilding the SA inside the loop works with no leak. Negligible cost.

```rust
loop {
    server.connect().await?;
    let (mut sa, _sec_guard) = create_user_only_security_attributes()?;
    let next_server = unsafe {
        ServerOptions::new()
            .pipe_mode(PipeMode::Byte)
            .create_with_security_attributes_raw(
                &pipe_path,
                (&raw mut sa).cast::<c_void>(),
            )?
    };
    // ...
}
```

2. The `instances.json` registry does a read-modify-write with no lock (`register_instance` / `unregister_instance`, `server/windows.rs:68-93`). Two Tauri apps booting in parallel can clobber each other: A reads, B reads, A writes `{A1, A2}`, B writes `{A1, B1}`, entry A2 is gone. The atomic `rename` protects against partial readers, not against concurrent writers. The CLI `resolve_socket` (`main.rs:1342`) reads this file so auto-detect becomes non-deterministic under load. Not hypothetical if CI spawns multiple apps.

Two options:
- File lock via `LockFileEx` on a sibling `instances.lock`.
- Better imo: one file per instance under `%LOCALAPPDATA%\tauri-pilot\instances\{identifier}.json`. Atomic create/remove, zero contention, `resolve_socket` just does a `read_dir` + parse. Bonus: at discovery time you can filter out entries whose PID is dead and clean up orphans.

**Medium, while the `unsafe` is still open:**

3. The `Vec<u8>` buffers cast to `ACL` and `SECURITY_DESCRIPTOR` (`windows.rs:170, 190`) are aligned to 1 byte, but those structs contain DWORDs and pointers that need 4/8 byte alignment. Works in practice because the allocator aligns to 8+, but it's formally UB and that's exactly why `#[allow(clippy::cast_ptr_alignment)]` sits at line 136. Not great to rely on that behavior in security code. `Box<MaybeUninit<SECURITY_DESCRIPTOR>>` for the SD (fixed size) and a `u32`-aligned buffer for the ACL fix it properly:

```rust
let mut sd = Box::<MaybeUninit<SECURITY_DESCRIPTOR>>::new_uninit();
unsafe { InitializeSecurityDescriptor(PSECURITY_DESCRIPTOR(sd.as_mut_ptr().cast()), 1)? };
let sd = unsafe { sd.assume_init() };
```

Same story for `TOKEN_USER` since it embeds a SID pointer.

**Small stuff:**

4. `async fn mock_server` in `client/mod.rs:93` and `client/windows.rs:28` has no `.await` (all the work goes into `tokio::spawn`). Passes current CI but `cargo clippy --workspace --all-targets -- -D warnings` blows up on it. Drop the `async`, the function already returns a `JoinHandle` directly.

5. `CHANGELOG.md` has an empty `[Unreleased]` section on this branch. Project CLAUDE.md requires an update on every PR, needs an `Added` section for Windows support (Named Pipes, DACL, registry-based discovery).

**To fix in the same PR to keep the review consistent:**

- Unix side has `peer_cred().uid() == getuid()` as a second barrier on top of `0o600` perms. Windows equivalent is `GetNamedPipeClientProcessId` + peer SID check. Redundant with a correct DACL, but 0.2.1 documents "three layers" on Unix and Windows should match the same guarantee instead of relying on one.
- `create_user_only_security_attributes` is 75 lines inside a single `unsafe` block. Split it into `open_token` → `get_user_sid` → `build_acl` → `build_sd`, one `unsafe { ... }` per FFI call with a local safety comment. Project convention caps functions at 50 lines anyway.
- Client tests use fixed paths (`/tmp/tauri-pilot-test-t05a.sock`, `\\.\pipe\tauri-pilot-test-t05a`). `server/unix.rs` already has `unique_socket_path()` built on `process::id()` + atomic counter. Reuse the same pattern on the client side, otherwise parallel `cargo test` will eventually cross-wire two tests sharing the same pipe.
- Add a test that checks via `GetSecurityDescriptorDacl` that the pipe actually carries a non-NULL DACL with 1 ACE. Without it a regression on point 1 slips back in silently.

Otherwise nice job on the module split and the `RegistryGuard` / `SocketGuard` symmetry. Both HIGH + the MEDIUM are fixable in the same pass while the `unsafe` code is still fresh, requesting changes until the whole set lands.

espetro:
Hey @mpiton , thanks for the detailed PR review and encouragement! It's one of my very first Rust PRs and definitely I'm learning a bunch. 
I've gone through your proposed changes and applied them; still, there are some issues being raised by the Cubic, Reptile, and CodeRabbit agents. Unless told otherwise, I'd also go through these issues and fix them.

mpiton:
@espetro

Thanks for the iterations, I know this is a lot of Win32 for a first Rust PR.

From my previous review, these are settled:
- DACL applied on every pipe instance in the accept loop.
- Registry race resolved via per-identifier files. Nicer than the file-lock I suggested, good call.
- `async fn mock_server` no longer declared async.
- CHANGELOG updated.
- Peer SID check added. But see item 7 below, the token API is wrong and the check is effectively a no-op today.

Still open from that list: splitting `create_user_only_security_attributes` into per-FFI `unsafe` blocks, and the DACL-present test. Neither is blocking by itself.

**Blocking (Windows CI will fail to build)**

None of these surface on a Linux dev box.

1. `crates/tauri-plugin-pilot/src/server/windows.rs` — `MaybeUninit` is used bare at lines 179, 180, 294, 321, 323, 349, 350 but only `use std::mem;` is in scope. Add `use std::mem::MaybeUninit;` or fully qualify.
2. `windows.rs:112` — `unsafe { let _ = CloseHandle(h) };` is missing the inner `;`. `let` is a statement, so the block doesn't parse. Should be `unsafe { let _ = CloseHandle(h); };`.
3. `windows.rs:62` — `windows::Win32::FileSystem::CreateDirectoryW` isn't a real path in `windows` 0.52. It's `windows::Win32::Storage::FileSystem::CreateDirectoryW`. Your test block at line 609 already uses `Storage::FileSystem::PACL`, so this is just a local typo.
4. `crates/tauri-pilot-cli/Cargo.toml:41` — `windows = "0.52"` has no `features` list, but `main.rs::is_pid_alive` uses `Win32::Foundation` and `Win32::System::Threading`. Add `features = ["Win32_Foundation", "Win32_System_Threading"]`.

**Critical (security / UB in the DACL code)**

I went through these against the live code, they're real and not bot noise. All in `windows.rs`:

5. Around the ACL allocation: you compute `acl_size` (header + ACE + SID, roughly 44 bytes) but allocate `Layout::new::<ACL>()`, which is only 8 bytes. `InitializeAcl` trusts the advertised size and `AddAccessAllowedAce` writes past the buffer. Heap overflow on every pipe creation. Use `Layout::from_size_align(acl_size as usize, std::mem::align_of::<ACL>()).unwrap()`.
6. `get_current_user_sid` (around 194–209) returns `token_user.User.Sid`, which is a pointer into `token_user_buf: Vec<u8>`. That Vec is dropped on return, so every later `GetLengthSid` / `AddAccessAllowedAce` dereferences freed memory. Return `(Vec<u8>, PSID)` together so the buffer stays alive, or have the caller own the buffer.
7. `windows.rs:222` — `OpenProcessToken(GetCurrentThread(), TOKEN_QUERY, ..)`. After `ImpersonateNamedPipeClient` the client token lives on the thread, not the process, so the correct API is `OpenThreadToken`. Right now the SID check is either fail-open (API rejects the thread pseudo-handle) or tautological (returns your own process token and compares it to itself). This is why my previous item 6 isn't actually closed.
8. ACL allocation-failure branch calls `std::alloc::dealloc(null_ptr, layout)` before returning. Dealloc of null through the global allocator is UB. Drop the dealloc line, the early return is enough.

**High**

9. `windows.rs:472` — the accept loop does `ServerOptions::new().create(...)?`. Any transient failure (`ERROR_PIPE_BUSY`, handle exhaustion) kills the server permanently. `continue` with a short `tokio::time::sleep` and a `tracing::warn!` is enough.
10. `windows.rs:389-399, 465-470` — if DACL setup fails the code falls back to `ServerOptions::new().create(path)` with the default DACL, which is broader than user-only. In a security-critical path, fail hard rather than silently weakening the ACL.
11. `windows.rs:53-67` — `instances_dir` creation hands a zero-filled `SECURITY_DESCRIPTOR {}` as `lpSecurityDescriptor` (never `InitializeSecurityDescriptor`'d), and `CreateDirectoryW` isn't recursive. Simplest fix: drop the SA entirely and use `std::fs::create_dir_all(&dir)?`. The directory sits under `%LOCALAPPDATA%\` which already inherits user-only ACLs.
12. `.gitignore` — this PR adds `CLAUDE.md`, which is the project instructions file and is version-controlled. Please drop that line. `.sisyphus/`, `.agents/`, and `skills-lock.json` look like agent-tool directories from your local setup, those belong in `~/.gitignore_global` rather than in-tree.
13. `crates/tauri-plugin-pilot/Cargo.toml:47-49` — the `[target.'cfg(target_os="linux")'.dependencies]` entry for `enigo` is missing `optional = true`. Cargo merges it with the top-level entry, so `enigo` plus its wayland/X11 backends ends up pulled unconditionally on Linux. The `press` feature gate is silently broken. Add `optional = true` to the target-specific entry.
14. `crates/tauri-plugin-pilot/src/lib.rs:35` — `init()` doc still says "On non-Unix platforms … no-op". Change to `(non-Unix, non-Windows)`.

**Nice to have in the same pass**

15. `server/mod.rs:41-53` — `reader.read_line(&mut line).await?` buffers the full line before the `MAX_LINE_LENGTH` check, so the check is useless against a peer sending 10GB without a newline. `reader.take(MAX_LINE_LENGTH as u64 + 1).read_line(&mut line).await?` closes it.
16. `windows.rs` is 677 lines. The 150-line convention is a soft cap, but this one is worth splitting into `server/windows/{mod,registry,security}.rs`.
17. Original item 7: `create_user_only_security_attributes` is still one giant `unsafe`. One `unsafe { ... }` per FFI call with a `// SAFETY:` line each.
18. `crates/tauri-pilot-cli/src/client/mod.rs:132, 209` — Unix client tests still use hard-coded paths (`...test-t05a.sock`, `...test-t05b.sock`). You already solved this on the Windows side with `unique_pipe_path()`, mirror it for Unix.
19. `test_resolve_socket_windows_reads_registry` mocks `"pid": 1234`. Now that liveness filtering landed, on real Windows CI that PID is almost certainly dead and the entry gets skipped. Use `std::process::id()` in the mock.
20. Original item 8: a test that reads back the DACL on a live pipe via `GetSecurityDescriptorDacl` and asserts one ACE. Without it, a regression on 5, 7, or 10 walks right back in silently.

**Bot comments you can skip**

- CodeRabbit "registry JSON read-modify-write race": obsolete, you switched to one file per instance.
- CodeRabbit "parse-error id hardcoded to 0": already returns `Value::Null`.
- Greptile "token handle leaked on early-return errors": `SecurityAttributesGuard` is constructed immediately after `OpenProcessToken` returns, its `Drop` covers every fallible call after.
- CodeRabbit "async fn without await on mock_server": plain `fn` now.
- CodeRabbit "#[serial] missing on Windows registry tests": already present at lines 1629 and 1666.
- CodeRabbit "fixed pipe names collide in mock_server" (Windows side): `unique_pipe_path()` is in use.
- CodeRabbit "empty LOCALAPPDATA produces a rooted bogus path": already guarded by \`.filter(|v| !v.is_empty())\`.

Real progress on this round. The registry redesign and the per-instance DACL were the big ones, both landed cleanly. The remaining critical work is concentrated in the SID/ACL construction and Windows token handling, so fix the four compile errors first (so CI can actually exercise the rest), then hit the four critical items in one pass while you're still in that file. Ping me when it's pushed and I'll re-review.

mpiton:
Reopening to retrigger CI workflow (fork PR stuck on pending approval)

mpiton:
@espetro I've got some time this afternoon, so I'm going to pick up the remaining items from my last review myself instead of sending you through another round. The Win32/DACL bits are faster for me to verify locally than to ping-pong back and forth.

Already kicked off the work. I'll push directly to your branch (thanks for leaving maintainer edits on) and you can eyeball the final commits before merge. If anything I push looks off, shout and I'll back it out.

Nothing needed from your side right now.

mpiton:
@cubic-dev-ai review

mpiton:
@greptileai review

mpiton (dismissed):

PR #99, opened 2026-05-21
feat(cli)!: namespace MCP tools under pilot.* and pin rmcp 1.7.0

Naoray:
## Summary

- Namespace every MCP tool exposed by `tauri-pilot mcp` under a `pilot.` prefix (e.g. `attrs` → `pilot.attrs`, `ping` → `pilot.ping`).
- Pin the `rmcp` manifest floor at `1.7.0` (was `^1.4.0`) to match the lockfile-resolved baseline already exercised by CI.
- Breaking change for MCP clients: bare tool names no longer resolve from the published list (the dispatch path still accepts them for backwards compat where possible, but the advertised tool surface is uniformly prefixed).

## Motivation

When `tauri-pilot mcp` is registered in an MCP client alongside other MCP servers, generic bare names like `attrs` or `ping` collide with — and shadow — tools published by those other servers. Without a prefix the client owner has to disambiguate by hand: rename tools at the client, edit prompts, or unregister conflicting servers.

Namespacing under `pilot.*` removes that burden and makes it unambiguous which server a given tool belongs to. The rmcp 1.7 conventions also favour explicit server-scoped tool names, so this aligns with where the ecosystem is moving.

The `rmcp` floor pin is a follow-up to #93: the lockfile already resolved `rmcp` to `1.7.0` via the earlier caret bump recorded under `[0.5.2]`. Raising the declared manifest floor aligns the dependency declaration with the tested baseline and prevents an accidental rebuild against a pre-`1.7.0` rmcp on a fresh dependency resolution.

## Changes

- `crates/tauri-pilot-cli/src/mcp.rs`: two private helpers (`normalize_tool_name`, `namespaced_tool_name`); every `ToolSpec.name` is wrapped via `format!(\"pilot.{name}\")` when building the public tool list; `call_tool_by_name` strips the prefix; `get_tool` re-applies it; `tool_list_matches_cli_command_surface` test updated to assert the prefix.
- `crates/tauri-pilot-cli/Cargo.toml`: `rmcp = \"1.7.0\"` (was `\"^1.4.0\"`).
- `CHANGELOG.md` `[Unreleased]` → `### Changed`: one bullet flagging the breaking tool-name namespacing with the affected tool list, one bullet for the manifest floor pin.

No tool semantics change. No tool added or removed. No MCP protocol-version change. No transport change.

## Test Plan

- [x] \`cargo test --workspace\` passes (265/265 tests).
- [x] \`cargo clippy --workspace -- -D warnings\` passes (no warnings).
- [x] Tested manually with a Tauri app: tools resolve under their \`pilot.*\` names; bare-name dispatch path still routes correctly for backwards compatibility.

## Checklist

- [x] No \`.unwrap()\` outside of tests.
- [x] All new files are under 150 lines. (No new files — only \`mcp.rs\` modified.)
- [x] Error handling uses \`thiserror\` (plugin) or \`anyhow\` (CLI). (No new error paths introduced; existing CLI \`anyhow\` flow is preserved.)
- [x] Commit messages follow conventional commits (\`feat:\`, \`fix:\`, \`refactor:\`, etc.). Subject is \`feat(cli)!: namespace MCP tools under pilot.* and pin rmcp 1.7.0\` with a \`BREAKING CHANGE:\` footer per Conventional Commits 1.0.0.

---
## Summary by cubic
Namespaces all MCP tools under `pilot.*` to avoid cross-server name collisions, and pins `rmcp` to `1.7.0`. Bare names still dispatch for backwards compatibility; prefix handling is centralized via `PILOT_PREFIX` and helper functions, and docs now advertise the prefixed tools.

- **Migration**
  - Update tool calls: `attrs` → `pilot.attrs`, `ping` → `pilot.ping`, etc.
  - `tools/list` advertises only `pilot.*`; `tools/call` accepts both forms. Docs now show `pilot.*`.

- **Dependencies**
  - Pin `rmcp` to `1.7.0` (was `^1.4.0`).

<sup>Written for commit ffb2a48345833e04d46af54710bacf0a8477232b. Summary will update on new commits. <a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/99?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **Breaking Changes**
  * MCP tool names are now namespaced with the pilot. prefix (e.g., pilot.toolname). Update calls and prompts to use prefixed names to avoid collisions; legacy bare names still resolve via the legacy call path for compatibility.

* **Chores**
  * Dependency for the runtime manifest/communication library pinned to 1.7.0.

* **Documentation**
  * Docs and README updated to reflect the new pilot.* naming and discovery behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mpiton/tauri-pilot/pull/99?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Breaking:** MCP tool names are now namespaced under `pilot.*` — every tool
+  exposed by `tauri-pilot mcp` is registered (and accepted in `tools/call`) as
+  `pilot.<name>` instead of the bare `<name>`. This covers all currently exposed
+  tools (e.g. `pilot.ping`, `pilot.click`, `pilot.fill`, `pilot.attrs`,
+  `pilot.snapshot`, `pilot.eval`, `pilot.assert_text`, etc.). Motivation: when
+  this CLI is registered in an MCP client alongside other MCP servers, generic
+  bare names like `attrs` or `ping` collide with — and shadow — tools from those
+  other servers, forcing the client owner to disambiguate by hand. Clients that
+  referred to bare names in their server registrations or inline prompts must
+  update to the prefixed form (`attrs` → `pilot.attrs`, `ping` → `pilot.ping`,
+  and so on). No tool semantics, schemas, or MCP protocol versions change.
+- Pin the `rmcp` manifest floor at `1.7.0` (was `^1.4.0`). The lockfile already
+  resolved to `1.7.0` via the earlier caret bump in `[0.5.2]`, so this only
+  aligns the declared dependency with the version actually being tested and
+  prevents an accidental rebuild against a pre-`1.7.0` rmcp. No behaviour
+  change.
 - Replace ASCII architecture diagram in `README.md` with an `assets/architecture.png`
   illustration (with descriptive `alt` text) for better rendering on crates.io,
   GitHub, and assistive technologies.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,17 +24,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- **Breaking:** MCP tool names are now namespaced under `pilot.*` — every tool
-  exposed by `tauri-pilot mcp` is registered (and accepted in `tools/call`) as
-  `pilot.<name>` instead of the bare `<name>`. This covers all currently exposed
-  tools (e.g. `pilot.ping`, `pilot.click`, `pilot.fill`, `pilot.attrs`,
-  `pilot.snapshot`, `pilot.eval`, `pilot.assert_text`, etc.). Motivation: when
-  this CLI is registered in an MCP client alongside other MCP servers, generic
-  bare names like `attrs` or `ping` collide with — and shadow — tools from those
-  other servers, forcing the client owner to disambiguate by hand. Clients that
-  referred to bare names in their server registrations or inline prompts must
-  update to the prefixed form (`attrs` → `pilot.attrs`, `ping` → `pilot.ping`,
-  and so on). No tool semantics, schemas, or MCP protocol versions change.
+- **Breaking (advertised surface):** MCP tool names are now namespaced under
+  `pilot.*` in the published tool list — every tool exposed by `tauri-pilot
+  mcp` is advertised as `pilot.<name>` (e.g. `pilot.ping`, `pilot.click`,
+  `pilot.fill`, `pilot.attrs`, `pilot.snapshot`, `pilot.eval`,
+  `pilot.assert_text`, etc.). Bare names continue to resolve through
+  `tools/call` for backwards compatibility, but the advertised surface that
+  MCP clients discover and register against is the prefixed form. Motivation:
+  when this CLI is registered alongside other MCP servers, generic bare names
+  like `attrs` or `ping` collide with — and shadow — tools from those servers,
+  forcing the client owner to disambiguate by hand. Clients that referred to
+  bare names in their server registrations or inline prompts should update to
+  the prefixed form (`attrs` → `pilot.attrs`, `ping` → `pilot.ping`, and so
+  on). No tool semantics, schemas, or MCP protocol versions change.
 - Pin the `rmcp` manifest floor at `1.7.0` (was `^1.4.0`). The lockfile already
   resolved to `1.7.0` via the earlier caret bump in `[0.5.2]`, so this only
   aligns the declared dependency with the version actually being tested and
--- a/README.md
+++ b/README.md
@@ -200,10 +200,12 @@ of shelling out for each command:
 }
 ```
 
-The MCP server exposes the same app-inspection and interaction surface as the CLI:
-`snapshot`, `click`, `fill`, `logs`, `network`, `eval`, `ipc`, `assert_*`, and the
-other testing tools. Use global flags before `mcp` to pin a specific app socket or
-window:
+The MCP server exposes the same app-inspection and interaction surface as the CLI,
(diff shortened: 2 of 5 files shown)


Naoray:
Thanks for the careful read. Both findings verified valid and addressed in two follow-up commits on the branch:

**CHANGELOG overstatement** — fixed in [`67ed890`](https://github.com/EmpireTwo/tauri-pilot/commit/67ed890). You're right: `call_tool_by_name` runs incoming names through `normalize_tool_name`, which strips the `pilot.` prefix, so bare names still dispatch. The previous "instead of the bare `<name>`" wording overstated the breaking surface. Re-worded as "Breaking (advertised surface)" with an explicit note that bare names continue to resolve through `tools/call` for backwards compatibility.

**Missing symmetry test** — added in [`19b1ee9`](https://github.com/EmpireTwo/tauri-pilot/commit/19b1ee9). Two new regression tests:

- `tool_name_helpers_are_round_trip_symmetric` exercises `normalize_tool_name` and `namespaced_tool_name` directly to lock in the canonical-form mapping.
- `get_tool_resolves_bare_and_prefixed_names_to_same_tool` constructs a default `PilotMcpServer` and asserts `get_tool("click")` and `get_tool("pilot.click")` return the same `Tool` (same name + description, resolved name is the prefixed form).

I did not add a second mock-socket test for the prefixed-name dispatch path — the existing `click_tool_sends_json_rpc_request` test already exercises `call_tool_by_name("click", ...)` end-to-end through `normalize_tool_name`, and the prefixed-name path goes through the same strip then lands in the same match arm, so a duplicate test would not exercise new behaviour. The commit body has the same rationale. Happy to add it if you'd prefer belt-and-suspenders coverage there.

Verification on the updated branch: `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` → 267 passed (was 265; the two new tests).

mpiton (commented):
Thanks for the PR. The motivation is clear and using `normalize_tool_name` to keep `tools/call` accepting bare names is a clean back-compat path. A couple of things before merge.

`build_tools()` doesn't go through `namespaced_tool_name` — it has the prefix inlined as `format!("pilot.{}", spec.name)`. Identical output today, but it means `"pilot."` is hard-coded in four spots (the two helpers, here, and the test assertion). If the prefix ever moves, that's an easy place for things to drift. Routing it through the helper makes the helpers the single source. A `const PILOT_PREFIX: &str = "pilot.";` reused inside both helpers would tighten it more, optional.

The other thing is docs. `README.md` around line 204 and `docs/src/content/docs/guides/ai-agents.md` around line 166 both list the MCP tool surface using bare names (`snapshot`, `click`, `fill`, ...). After this lands, `tools/list` advertises them as `pilot.snapshot`, `pilot.click`, etc. The bare names still work via `tools/call` so existing setups won't break, but anyone wiring this into Claude Desktop or similar will see the prefixed form in their client first, and the docs should match what the server actually publishes.

And if you want it cheap: `tool_name_helpers_are_round_trip_symmetric` only covers the happy path. Three extra lines for `""`, `"pilot."` alone, and `"pilot.pilot.click"` would lock the corner cases (single-strip is the intended behavior on double prefix, worth pinning). Not blocking.

Everything else looks good — tests pass locally (267/267), clippy clean, and the CHANGELOG wording reads accurately after `67ed890`.

Naoray:
Thanks for the thorough review. All three items addressed in two follow-up commits:

**Item 1 — `build_tools()` routing + `const PILOT_PREFIX`** → [`30fac66`](https://github.com/EmpireTwo/tauri-pilot/commit/30fac66)

- Added `const PILOT_PREFIX: &str = "pilot.";` at `mcp.rs:637`.
- `normalize_tool_name` (mcp.rs:639) and `namespaced_tool_name` (mcp.rs:643) both consume the const.
- `build_tools()` now calls `namespaced_tool_name(spec.name)` instead of the inline `format!("pilot.{}", spec.name)`.
- The `tool_list_matches_cli_command_surface` test assertion was using a literal `"pilot."` to verify the prefix; that's now `PILOT_PREFIX` too.

Net effect: strict-prefix sites went from 4 → 1 (just the const). The remaining `"pilot."` substring occurrences are inside test fixture strings like `"pilot.click"` that document specific input/output pairs — those are intentional and don't drift if the prefix moves, because the test also reads the prefix via the helper under test.

**Item 2 — docs sync** → [`ffb2a48`](https://github.com/EmpireTwo/tauri-pilot/commit/ffb2a48)

- `README.md:203-208` — replaced bare-name listing with `pilot.snapshot`, `pilot.click`, `pilot.fill`, `pilot.logs`, `pilot.network`, `pilot.eval`, `pilot.ipc`, `pilot.assert_*`. Added one sentence noting bare names still resolve via `tools/call` so existing client setups keep working.
- `docs/src/content/docs/guides/ai-agents.md:165-170` — mirror change with the same backwards-compat note.

**Item 3 — corner-case tests** → also in [`30fac66`](https://github.com/EmpireTwo/tauri-pilot/commit/30fac66)

New test `tool_name_helpers_handle_corner_cases` at `mcp.rs:1665-1678` pins:

- `normalize_tool_name("")` → `""`
- `normalize_tool_name("pilot.")` → `""`
- `normalize_tool_name("pilot.pilot.click")` → `"pilot.click"` (single-strip; double prefix preserves the inner one)
- `namespaced_tool_name("")` → `"pilot."`
- `namespaced_tool_name("pilot.")` → `"pilot."` (idempotent prefix add)
- `namespaced_tool_name("pilot.click")` → `"pilot.click"` (idempotent on already-namespaced)

Verification on the updated branch: `cargo fmt` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` → 268 passed (was 267; +1 from the new corner-case test).

mpiton:
Thanks !

PR #100, opened 2026-05-21
feat(plugin): add macOS native screenshot backend with pilot.screenshot IPC method

Naoray:
## Summary

- New `pilot.screenshot` IPC method on `tauri-plugin-pilot` for macOS window capture. Path-returning shape, atomic write, TCC fallback surfaced as response metadata.
- Bare `screenshot` JSON-RPC method now dispatches on `output_path` presence: with → native; without → existing bridge html-to-image base64. Backwards compatible.
- New MCP tool `pilot.screenshot` advertised in `tools/list` with a path-only schema. Existing bridge `screenshot` MCP route unchanged.

## Motivation

Test-harness consumers running file-based regression-canary pipelines need deterministic per-window screenshot capture without driving the system's hotkey-based `screencapture` shell tool or maintaining out-of-tree FFI to `CGWindowListCreateImage`. Today those callers chain `open -a <app>` → `screencapture -x` → Swift CGWindow probe → `image` crate crop → strict pixel diff. That's two shell-outs and one out-of-tree FFI wrapper per capture. A path-returning `pilot.screenshot` collapses that chain.

**Path-returning over inline bytes.** Retina captures land in the 700KB–1.9MB range. Sending them over JSON-RPC framing + UDS buffer — and especially over MCP base64 — is the wrong transport choice when the caller is going to write the bytes to disk anyway. Path-only also keeps the MCP advertised surface small enough that agent runtimes don't bump into response-size limits.

**Atomic write contract.** Caller can assume the file at `output_path` is either the fresh capture or unchanged from before the call. Plugin writes to `<output_path>.tmp.<pid>` and `rename(tmp, final)` is the single atomic step. Every error branch after capture-success cleans up the tmp file. The PNG is decoded before the rename, so malformed bytes never land at the final path. Strict-diff pipelines depend on this fresh-vs-stale guarantee.

**TCC denial as metadata, not silent.** When macOS Screen Recording permission is denied, the plugin falls back to `CGWindowListCreateImage` and returns the capture with `backend: "cgwindow"` + `tcc_denied: true` in the response. Silent fallback would break strict-diff pipelines (the two backends produce visibly different output — CGWindow loses cursor + some compositor effects). Hard-erroring would prevent any CI use without an interactive Screen Recording grant. Surfacing the flag lets callers tag baselines per backend.

**Discovery in error payload.** When `window_id` doesn't resolve, the `WINDOW_NOT_FOUND` error carries an `available_windows[]` list (up to 20 entries: window_id + owner + title + layer 0). Caller doesn't need a separate `list_windows` round-trip to recover.

## Changes

- **`crates/tauri-plugin-pilot/src/screenshot/ipc.rs`** (new) — request validators, atomic write, TCC fallback dispatch, success/error response builders, 5 contract tests.
- **`crates/tauri-plugin-pilot/src/screenshot/window_id.rs`** — adds `DiscoveredWindow`, `WindowBounds`, `list_layer_zero_windows`, `get_window_bounds` for the `WINDOW_NOT_FOUND` discovery payload (cap at `AVAILABLE_WINDOWS_CAP = 20`).
- **`crates/tauri-plugin-pilot/src/screenshot/{mod,native,probe,error}.rs`** — re-exports + per-variant `#[allow(dead_code)]` on unused stub variants now that the module-wide allow is dropped.
- **`crates/tauri-plugin-pilot/src/handler.rs`** — `pilot.screenshot` dispatch arm; bare `screenshot` arm dispatches on `output_path` shape (preserves backwards compatibility for existing bridge callers); +3 dispatch-tier tests including a regression test for the no-`output_path` → bridge route.
- **`crates/tauri-pilot-cli/src/mcp.rs`** — `pilot.screenshot` `ToolSpec` + schema (`window_id` integer, `output_path` string, `format` enum["png"]; `window_id` + `output_path` required); `call_tool_by_name` arm; +3 MCP-tier tests including a path-only-contract assertion that fails closed if the schema ever grows `bytes`/`base64`/`data` keys.
- **`crates/tauri-plugin-pilot/src/lib.rs`** — drops the `#[allow(dead_code, unused_imports)]` on `mod screenshot` now that there are real callers.
- **`CHANGELOG.md`** — `[Unreleased] ### Added` entry.

### Contract surface

```jsonc
// Request
{
  "method": "pilot.screenshot",
  "params": {
    "window_id": <u32>,             // required
    "output_path": "<absolute>",     // required, parent dir must exist
    "format": "png"                  // optional; v1 accepts only "png"
  }
}

// Success response
{
  "output_path": "<absolute>",
  "window_id": <u32>,
  "width": <u32>,
  "height": <u32>,
  "scale_factor": <f32>,
  "byte_size": <u64>,
  "backend": "screencapture" | "cgwindow",
  "tcc_denied": <bool>
}

// Error: window not found
{
  "error": {
    "code": "WINDOW_NOT_FOUND",
    "message": "window_id <id> not found",
    "available_windows": [
      { "window_id": <u32>, "owner": "<name>", "title": "<string>", "layer": 0 }
    ]
  }
}
```

Other error codes: `INVALID_PARAMS`, `INVALID_OUTPUT_PATH`, `UNSUPPORTED_FORMAT`, `PERMISSION_DENIED`, `CAPTURE_FAILED`.

**Non-macOS** (Linux/Windows): the method is registered but returns `PERMISSION_DENIED` with an explanatory message. Native backends for those platforms are out of scope for this PR.

## Test Plan

- [x] `cargo test --workspace` passes (276 tests: 143 plugin + 131 cli + 2 cli integration; 11 new tests landed on this commit covering all 5 validator paths, the dispatch-on-shape behaviour for bare `screenshot`, the `WINDOW_NOT_FOUND` discovery payload, the path-only MCP schema invariant, and the bridge backwards-compat route).
- [x] `cargo clippy --workspace -- -D warnings` passes (no diagnostics, no warnings escalated).
- [ ] Tested manually with a Tauri app: validator + dispatch + MCP-tier paths have unit coverage; the real-capture round-trip needs an interactive Screen Recording grant and a windowed host, so I've deferred it to a maintainer's local manual run. Happy to add an `#[ignore]`d integration test that drives the full pipeline behind an env-gated host if useful.

## Checklist

- [x] No `.unwrap()` outside of tests. Workspace lint `clippy::unwrap_used = "deny"` enforces this; clippy is clean.
- [ ] All new files are under 150 lines. **Honest note**: `screenshot/ipc.rs` is 561 LOC and `screenshot/window_id.rs` is 310 LOC, both over the 150-line cap declared in `CONTRIBUTING.md`. The seams for a 4-module split of `ipc.rs` are clear (`ipc/validators.rs`, `ipc/dispatch.rs`, `ipc/tcc.rs`, `ipc/atomic_write.rs`); `window_id.rs` is harder to split usefully because the three functions all share the same `CGWindowListCopyWindowInfo` enumeration shape and splitting would duplicate the unsafe `wrap_under_*` block. Happy to land the split as either a follow-up PR or a pre-merge amend depending on which you prefer. Noting the existing files in this crate already over cap (`handler.rs` 1475, `lib.rs` 465, `mcp.rs` 1999); flagging the inconsistency rather than silently shipping over.
- [x] Error handling uses `thiserror` — `ScreenshotError` and the new `ipc::error::Code` are both `#[derive(thiserror::Error)]`. No `anyhow` in the plugin crate paths.
- [x] Commit messages follow conventional commits: two commits on top of `main` — `feat(plugin): add macOS native screenshot backend` (c8cec44) and `feat(plugin): wire pilot.screenshot IPC method` (52a9091). Both with multi-paragraph bodies.

---
## Summary by cubic
Adds a macOS-native screenshot backend and a new `screenshot_native` JSON-RPC/MCP method to capture a window directly to a PNG on disk and return its path with metadata. The bridge `screenshot` stays separate and unchanged.

- **New Features**
  - New `screenshot_native` method in `tauri-plugin-pilot` and MCP tool `pilot.screenshot_native`; capture by `window_id` to an absolute `output_path` (`format: "png"` only).
  - Atomic write via temp + rename; response includes `output_path`, `window_id`, `width`, `height`, `scale_factor`, `byte_size`, `backend` ("screencapture" | "cgwindow"), and `tcc_denied`.
  - macOS: auto-selects backend with CoreGraphics preflight; falls back to `CGWindowListCreateImage` when Screen Recording is denied and sets `tcc_denied: true`.
  - `WINDOW_NOT_FOUND` includes a capped `available_windows` list to aid discovery.
  - Non-macOS: method is registered, validators run, then `PERMISSION_DENIED`. The bridge `screenshot` remains html-to-image (base64) and never routes to native.

- **Bug Fixes**
  - Hardened validators: reject relative paths, symlinks, directories, and missing parent dirs with clear `INVALID_OUTPUT_PATH` errors.
  - Gated macOS-only screenshot modules to fix cross-platform clippy errors; no runtime changes.

<sup>Written for commit 4aa5d638de2fce15d28995159afcc3144697a7e8. Summary will update on new commits. <a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/100?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * Added macOS-native screenshot tool (pilot.screenshot_native) that captures a window to a caller-provided output path and returns the path plus metadata (dimensions, scale, size, backend, tcc_denied). Uses screencapture when permitted, falls back to CoreGraphics when not; on Linux/Windows the tool returns PERMISSION_DENIED.
* **Documentation**
  * CHANGELOG updated with native screenshot behavior and permission details.
* **Tests**
  * Added tests covering tool registration, input validation, platform behavior, and Unix IPC flow.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mpiton/tauri-pilot/pull/100?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - macOS native screenshot backend (`screencapture` shell-out plus
   `CGWindowListCreateImage` fallback) behind the existing `screenshot`
   module surface; WKWebView path scaffolded for follow-up.
+- `pilot.screenshot` JSON-RPC method (also advertised as `pilot.screenshot`
+  MCP tool) that captures a window by `window_id` to a caller-specified
+  `output_path`. Uses macOS `screencapture` when Screen Recording permission
+  is granted, falls back to `CGWindowListCreateImage` with `tcc_denied: true`
+  in metadata when permission is revoked between probe and call. Path-only
+  response shape — no inline bytes. The bare `screenshot` method continues
+  to route to the bridge html-to-image path for backwards compatibility, and
+  also accepts the new native shape when callers pass `output_path`.
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- macOS native screenshot backend (`screencapture` shell-out plus
+  `CGWindowListCreateImage` fallback) behind the existing `screenshot`
+  module surface; WKWebView path scaffolded for follow-up.
+
 ### Security
 
 - `SKILL.md`: second pass on skills.sh Snyk findings W007 and W011 — the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,14 +12,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - macOS native screenshot backend (`screencapture` shell-out plus
   `CGWindowListCreateImage` fallback) behind the existing `screenshot`
   module surface; WKWebView path scaffolded for follow-up.
-- `pilot.screenshot` JSON-RPC method (also advertised as `pilot.screenshot`
+- `screenshot_native` JSON-RPC method (advertised as the `pilot.screenshot_native`
   MCP tool) that captures a window by `window_id` to a caller-specified
   `output_path`. Uses macOS `screencapture` when Screen Recording permission
   is granted, falls back to `CGWindowListCreateImage` with `tcc_denied: true`
   in metadata when permission is revoked between probe and call. Path-only
-  response shape — no inline bytes. The bare `screenshot` method continues
-  to route to the bridge html-to-image path for backwards compatibility, and
-  also accepts the new native shape when callers pass `output_path`.
+  response shape — no inline bytes. Bare `screenshot` keeps its existing
+  bridge html-to-image behaviour and is wholly separate from the new method,
+  so the two surfaces cannot be confused. On Linux and Windows,
+  `pilot.screenshot_native` is registered but every call responds with
+  `PERMISSION_DENIED` and the message `"screenshot is only available on
+  macOS in this release"` — the request shape validators run first, so a
+  contract-violating call sees the validation error on every host.
 
 ### Security
 
--- a/crates/tauri-pilot-cli/src/mcp.rs
+++ b/crates/tauri-pilot-cli/src/mcp.rs
@@ -248,6 +248,17 @@ impl PilotMcpServer {
                 )
                 .await
             }
+            "pilot.screenshot" => {
+                let mut payload = json!({
+                    "window_id": required_u64(&args, "window_id")?,
+                    "output_path": required_string(&args, "output_path")?,
+                });
+                if let Some(format) = optional_string(&args, "format")? {
+                    payload["format"] = json!(format);
+                }
+                self.call_app_tool("pilot.screenshot", Some(payload), window)
+                    .await
+            }
             "navigate" => {
                 self.call_app_tool(
                     "navigate",
@@ -829,6 +840,14 @@ fn tool_specs() -> Vec<ToolSpec> {
             idempotent: false,
         },
         ToolSpec {
+            name: "pilot.screenshot",
+            description: "Capture a native window by `window_id` and write a PNG to `output_path`. Returns path + metadata; never inlines bytes.",
+            schema: pilot_screenshot_schema,
+            read_only: false,
+            destructive: false,
+            idempotent: false,
+        },
+        ToolSpec {
             name: "scroll",
             description: "Scroll the page or an element ref.",
             schema: scroll_schema,
@@ -1386,6 +1405,28 @@ fn ipc_schema() -> Arc<JsonObject> {
     )
 }
 
+fn pilot_screenshot_schema() -> Arc<JsonObject> {
+    object_schema(
+        props([
+            (
+                "window_id",
+                integer_prop("Native window id (e.g. `CGWindowID` on macOS) to capture."),
+            ),
+            (
+                "output_path",
+                string_prop(
+                    "Absolute path where the PNG is written. The parent directory must already exist; the file is written atomically via a temp + rename.",
+                ),
+            ),
+            (
+                "format",
+                enum_prop("Image format. v1 accepts only \"png\".", &["png"]),
+            ),
+        ]),
+        &["window_id", "output_path"],
+    )
+}
+
 fn selector_schema() -> Arc<JsonObject> {
     object_schema(
         props([("selector", string_prop("Optional CSS selector."))]),
@@ -1627,6 +1668,7 @@ mod tests {
             "logs",
             "navigate",
             "network",
+            "pilot.screenshot",
             "ping",
             "press",
             "record_start",
@@ -1726,6 +1768,120 @@ mod tests {
                 "scroll direction enum must include {expected}"
             );
         }
+    }
+
+    #[test]
+    fn pilot_screenshot_tool_advertises_path_only_contract() {
+        let tool = cached_tools()
+            .iter()
+            .find(|t| t.name == "pilot.screenshot")
+            .expect("pilot.screenshot tool registered");
+        let props = tool
+            .input_schema
+            .get("properties")
+            .and_then(Value::as_object)
+            .expect("schema has properties");
+        for required in ["window_id", "output_path", "format"] {
+            assert!(
+                props.contains_key(required),
+                "pilot.screenshot must advertise `{required}` in its schema"
+            );
+        }
+        let required = tool
+            .input_schema
+            .get("required")
+            .and_then(Value::as_array)
+            .expect("required list present");
+        let required: Vec<&str> = required.iter().filter_map(Value::as_str).collect();
+        assert!(
+            required.contains(&"window_id"),
+            "pilot.screenshot must require `window_id`"
+        );
+        assert!(
+            required.contains(&"output_path"),
+            "pilot.screenshot must require `output_path`"
+        );
+        // The path-only contract forbids inline byte / base64 fields on either
+        // surface — guard against a future revision silently growing one.
+        for forbidden in ["bytes", "base64", "data"] {
+            assert!(
+                !props.contains_key(forbidden),
+                "pilot.screenshot must not advertise `{forbidden}` (path-only contract)"
+            );
+        }
+    }
+
+    #[test]
+    fn pilot_screenshot_tool_routes_native_method() {
+        // The native contract requires `pilot.screenshot` (namespaced) on
+        // JSON-RPC so the existing bridge `screenshot` (html-to-image,
+        // base64) keeps working for current CLI/scenario callers. This test
+        // pins the surface so a future refactor cannot silently fold the two
+        // tools together and resurrect the bytes-inline payload shape.
+        let bare = cached_tools()
+            .iter()
+            .find(|t| t.name == "screenshot")
+            .expect("bridge screenshot tool still registered");
+        let prefixed = cached_tools()
+            .iter()
+            .find(|t| t.name == "pilot.screenshot")
+            .expect("pilot.screenshot tool registered");
+        assert_ne!(
+            bare.input_schema, prefixed.input_schema,
+            "bridge `screenshot` and native `pilot.screenshot` advertise different schemas"
+        );
+    }
+
+    #[tokio::test]
+    #[cfg(unix)]
+    async fn pilot_screenshot_forwards_native_params_to_jsonrpc() {
+        let socket = std::env::temp_dir().join(format!(
+            "tauri-pilot-mcp-pilot-screenshot-{}.sock",
+            std::process::id()
+        ));
+        let _ = std::fs::remove_file(&socket);
+        let listener = UnixListener::bind(&socket).expect("bind mock socket");
+        let server = tokio::spawn(async move {
+            let (stream, _) = listener.accept().await.expect("accept");
+            let (reader, mut writer) = stream.into_split();
+            let mut reader = BufReader::new(reader);
+            let mut line = String::new();
+            reader.read_line(&mut line).await.expect("read request");
+            let request: Request = serde_json::from_str(line.trim()).expect("parse request");
+            assert_eq!(request.method, "pilot.screenshot");
+            let params = request.params.expect("native screenshot params present");
+            assert_eq!(params["window_id"], json!(42_u64));
+            assert_eq!(params["output_path"], json!("/tmp/out.png"));
+            let response = Response::success(
+                request.id,
+                json!({
+                    "output_path": "/tmp/out.png",
+                    "window_id": 42_u32,
+                    "width": 100_u32,
+                    "height": 50_u32,
+                    "scale_factor": 2.0_f32,
+                    "byte_size": 1234_u64,
+                    "backend": "screencapture",
+                    "tcc_denied": false,
+                }),
+            );
+            let mut bytes = serde_json::to_vec(&response).expect("serialize response");
+            bytes.push(b'\n');
+            writer.write_all(&bytes).await.expect("write response");
+        });
+
+        let pilot = PilotMcpServer::new(Some(socket.clone()), None);
+        let mut args = Map::new();
+        args.insert("window_id".to_owned(), json!(42_u32));
+        args.insert("output_path".to_owned(), json!("/tmp/out.png"));
+        let result = pilot
+            .call_tool_by_name("pilot.screenshot", args)
+            .await
+            .expect("tool call succeeds");
+        assert_eq!(result.is_error, Some(false));
+
+        server.await.expect("mock server task");
+        let _ = std::fs::remove_file(&socket);
     }
 
     #[test]
--- a/crates/tauri-pilot-cli/src/mcp.rs
+++ b/crates/tauri-pilot-cli/src/mcp.rs
@@ -248,15 +248,15 @@ impl PilotMcpServer {
                 )
                 .await
             }
-            "pilot.screenshot" => {
+            "screenshot_native" => {
                 let mut payload = json!({
-                    "window_id": required_u64(&args, "window_id")?,
+                    "window_id": required_u32(&args, "window_id")?,
                     "output_path": required_string(&args, "output_path")?,
                 });
                 if let Some(format) = optional_string(&args, "format")? {
                     payload["format"] = json!(format);
                 }
-                self.call_app_tool("pilot.screenshot", Some(payload), window)
+                self.call_app_tool("screenshot_native", Some(payload), window)
                     .await
             }
             "navigate" => {
@@ -840,7 +840,7 @@ fn tool_specs() -> Vec<ToolSpec> {
             idempotent: false,
         },
         ToolSpec {
-            name: "pilot.screenshot",
+            name: "screenshot_native",
             description: "Capture a native window by `window_id` and write a PNG to `output_path`. Returns path + metadata; never inlines bytes.",
             schema: pilot_screenshot_schema,
             read_only: false,
@@ -1110,6 +1110,11 @@ fn required_u64(args: &JsonObject, name: &str) -> Result<u64, McpError> {
     args.get(name)
         .and_then(Value::as_u64)
         .ok_or_else(|| invalid_params(format!("'{name}' is required and must be an integer")))
+}
+
+fn required_u32(args: &JsonObject, name: &str) -> Result<u32, McpError> {
+    let value = required_u64(args, name)?;
+    u32::try_from(value).map_err(|_| invalid_params(format!("'{name}' is out of range for u32")))
 }
 
 fn optional_u64(args: &JsonObject, name: &str) -> Result<Option<u64>, McpError> {
@@ -1410,7 +1415,12 @@ fn pilot_screenshot_schema() -> Arc<JsonObject> {
         props([
             (
                 "window_id",
-                integer_prop("Native window id (e.g. `CGWindowID` on macOS) to capture."),
+                json!({
+                    "type": "integer",
+                    "minimum": 0,
+                    "maximum": u32::MAX,
+                    "description": "Native window id (e.g. `CGWindowID` on macOS) to capture. Must fit in u32.",
+                }),
             ),
             (
                 "output_path",
@@ -1668,7 +1678,6 @@ mod tests {
             "logs",
             "navigate",
             "network",
-            "pilot.screenshot",
             "ping",
             "press",
             "record_start",
@@ -1676,6 +1685,7 @@ mod tests {
             "record_stop",
             "replay",
             "screenshot",
+            "screenshot_native",
             "scroll",
             "select",
             "snapshot",
@@ -1774,8 +1784,8 @@ mod tests {
     fn pilot_screenshot_tool_advertises_path_only_contract() {
         let tool = cached_tools()
             .iter()
-            .find(|t| t.name == "pilot.screenshot")
-            .expect("pilot.screenshot tool registered");
+            .find(|t| t.name == "pilot.screenshot_native")
+            .expect("pilot.screenshot_native tool registered");
         let props = tool
             .input_schema
             .get("properties")
@@ -1784,7 +1794,7 @@ mod tests {
         for required in ["window_id", "output_path", "format"] {
             assert!(
                 props.contains_key(required),
-                "pilot.screenshot must advertise `{required}` in its schema"
+                "pilot.screenshot_native must advertise `{required}` in its schema"
             );
         }
         let required = tool
@@ -1795,40 +1805,41 @@ mod tests {
         let required: Vec<&str> = required.iter().filter_map(Value::as_str).collect();
         assert!(
             required.contains(&"window_id"),
-            "pilot.screenshot must require `window_id`"
+            "pilot.screenshot_native must require `window_id`"
         );
         assert!(
             required.contains(&"output_path"),
-            "pilot.screenshot must require `output_path`"
+            "pilot.screenshot_native must require `output_path`"
         );
         // The path-only contract forbids inline byte / base64 fields on either
         // surface — guard against a future revision silently growing one.
         for forbidden in ["bytes", "base64", "data"] {
             assert!(
                 !props.contains_key(forbidden),
-                "pilot.screenshot must not advertise `{forbidden}` (path-only contract)"
+                "pilot.screenshot_native must not advertise `{forbidden}` (path-only contract)"
             );
         }
     }
 
     #[test]
     fn pilot_screenshot_tool_routes_native_method() {
-        // The native contract requires `pilot.screenshot` (namespaced) on
-        // JSON-RPC so the existing bridge `screenshot` (html-to-image,
-        // base64) keeps working for current CLI/scenario callers. This test
-        // pins the surface so a future refactor cannot silently fold the two
-        // tools together and resurrect the bytes-inline payload shape.
-        let bare = cached_tools()
-            .iter()
-            .find(|t| t.name == "screenshot")
-            .expect("bridge screenshot tool still registered");
-        let prefixed = cached_tools()
+        // The native contract lives under a distinct name (`screenshot_native`
+        // advertised as `pilot.screenshot_native`) so the existing bridge
+        // `screenshot` (html-to-image, base64) keeps working for current
+        // CLI/scenario callers. This test pins the surface so a future
+        // refactor cannot silently fold the two tools together and resurrect
+        // the bytes-inline payload shape.
+        let bridge = cached_tools()
             .iter()
             .find(|t| t.name == "pilot.screenshot")
-            .expect("pilot.screenshot tool registered");
+            .expect("bridge pilot.screenshot tool still registered");
+        let native = cached_tools()
+            .iter()
+            .find(|t| t.name == "pilot.screenshot_native")
+            .expect("pilot.screenshot_native tool registered");
         assert_ne!(
-            bare.input_schema, prefixed.input_schema,
-            "bridge `screenshot` and native `pilot.screenshot` advertise different schemas"
+            bridge.input_schema, native.input_schema,
+            "bridge `pilot.screenshot` and native `pilot.screenshot_native` advertise different schemas"
         );
     }
 
@@ -1836,7 +1847,7 @@ mod tests {
     #[cfg(unix)]
     async fn pilot_screenshot_forwards_native_params_to_jsonrpc() {
         let socket = std::env::temp_dir().join(format!(
-            "tauri-pilot-mcp-pilot-screenshot-{}.sock",
+            "tauri-pilot-mcp-screenshot-native-{}.sock",
             std::process::id()
         ));
         let _ = std::fs::remove_file(&socket);
@@ -1848,10 +1859,11 @@ mod tests {
             let mut line = String::new();
             reader.read_line(&mut line).await.expect("read request");
             let request: Request = serde_json::from_str(line.trim()).expect("parse request");
-            assert_eq!(request.method, "pilot.screenshot");
+            assert_eq!(request.method, "screenshot_native");
             let params = request.params.expect("native screenshot params present");
             assert_eq!(params["window_id"], json!(42_u64));
             assert_eq!(params["output_path"], json!("/tmp/out.png"));
+            assert_eq!(params["format"], json!("png"));
             let response = Response::success(
                 request.id,
                 json!({
@@ -1874,8 +1886,9 @@ mod tests {
         let mut args = Map::new();
         args.insert("window_id".to_owned(), json!(42_u32));
         args.insert("output_path".to_owned(), json!("/tmp/out.png"));
+        args.insert("format".to_owned(), json!("png"));
         let result = pilot
-            .call_tool_by_name("pilot.screenshot", args)
+            .call_tool_by_name("screenshot_native", args)
             .await
             .expect("tool call succeeds");
         assert_eq!(result.is_error, Some(false));
--- a/crates/tauri-plugin-pilot/src/handler.rs
+++ b/crates/tauri-plugin-pilot/src/handler.rs
@@ -4,6 +4,7 @@ use crate::eval::EvalEngine;
 use crate::key;
 use crate::protocol::RpcError;
 use crate::recorder::{RecordEntry, Recorder};
+use crate::screenshot;
 use crate::server::{EvalFn, FocusFn, ListWindowsFn};
 
 use std::time::Duration;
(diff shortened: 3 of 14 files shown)


Naoray:
Heads up — the Ubuntu + Windows lanes were red on `52a9091` because my local clippy run only covered macOS. Pushed [`b9571f4`](https://github.com/EmpireTwo/tauri-pilot/commit/b9571f4) `fix(plugin): satisfy cross-platform clippy on non-macOS screenshot stubs` to address all 22 cross-platform clippy errors. Re-run is currently pending on run [26249434621](https://github.com/mpiton/tauri-pilot/actions/runs/26249434621).

**Error catalog (22 total, identical on Ubuntu + Windows):**

- 6 × `unused_imports` (the macOS-only re-exports leaking into non-macOS builds).
- 15 × `dead_code` (stub fns + cfg-gated items that the non-macOS IPC short-circuit already makes unreachable).
- 1 × `clippy::doc-markdown` (bare `CGWindow` identifier in a non-macOS stub doc comment).

**Fix approach:**

The root cause was that the `screenshot/` submodules each declared a `#[cfg(not(target_os = "macos"))]` stub fn, but the `ipc::handle_screenshot` entry point already short-circuits on non-macOS with `PERMISSION_DENIED` before reaching any backend. The stubs were dead source.

Cleaned that up:

1. **Gated macOS-only submodules in `screenshot/mod.rs`.** `cgwindow`, `error`, `native`, `probe`, `screencapture`, `window_id` are now all `#[cfg(target_os = "macos")]`. `ipc` and `handle_screenshot` remain unconditional so the JSON-RPC method is registered on every platform (and returns `PERMISSION_DENIED` outside macOS, per the original contract).
2. **Deleted the now-unreachable non-macOS stubs** in `cgwindow.rs`, `screencapture.rs`, `window_id.rs`, `native.rs`, plus the `probe_backend` non-macOS branch.
3. **For items in `ipc.rs` that compile on every platform** but only get exercised on macOS (`ScreenshotRequest` fields, `codes::WINDOW_NOT_FOUND`, `codes::CAPTURE_FAILED`): kept them, gated with `#[cfg_attr(not(target_os = "macos"), allow(dead_code, reason = "..."))]` so the contract surface stays consistent across platforms.
4. **Backticked `CGWindow` / `CoreGraphics`** in the surviving doc comments. No `#[allow(clippy::doc_markdown)]` suppression.

**Verification on the updated branch (macOS local):**

- `cargo fmt --all -- --check` clean.
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo test --workspace` → 276 passed.

Couldn't reproduce the Ubuntu/Windows clippy locally (only `aarch64-apple-darwin` and `x86_64-unknown-linux-musl` targets installed; musl stdlib not provisioned), so the upstream CI run is the actual verification. Tracking the run, will follow up if anything else surfaces.

**Not addressed in this commit (still on my list for a separate round, since you asked for the maintainer-side review specifically):**

- cubic P1 (tmp path uniqueness on `ipc.rs:326`).
- coderabbit MAJOR (`cgwindow.rs:54` row arithmetic overflow guard).
- coderabbit MINOR (`mcp.rs:253` / `1391` `required_u32` enforcement).
- coderabbit nitpick (`mcp.rs:1772` format-field assertion in the JSON-RPC forwarding test).

Happy to bundle those into a follow-up round once you've had a chance to look at the surface itself. Let me know which order you'd prefer.

CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- macOS native screenshot backend (`screencapture` shell-out plus
+  `CGWindowListCreateImage` fallback) behind the existing `screenshot`
+  module surface; WKWebView path scaffolded for follow-up.
+- `pilot.screenshot` JSON-RPC method (also advertised as `pilot.screenshot`
+  MCP tool) that captures a window by `window_id` to a caller-specified
+  `output_path`. Uses macOS `screencapture` when Screen Recording permission
+  is granted, falls back to `CGWindowListCreateImage` with `tcc_denied: true`
+  in metadata when permission is revoked between probe and call. Path-only
+  response shape — no inline bytes. The bare `screenshot` method continues
```
mpiton: Worth adding one line about non-macOS behavior. On Linux and Windows `pilot.screenshot` is registered but responds `PERMISSION_DENIED` with `screenshot is only available on macOS in this release` (per `ipc.rs:212-217`). Without it, operators wiring up MCP clients on those hosts will discover it empirically.

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -247,6 +247,17 @@ impl PilotMcpServer {
                 )
                 .await
             }
+            "pilot.screenshot" => {
+                let mut payload = json!({
+                    "window_id": required_u64(&args, "window_id")?,
```
mpiton: `required_u64` accepts anything up to `u64::MAX` and forwards it to the plugin, which then rejects with `INVALID_PARAMS` for the u32 overflow. Extra round-trip for nothing. Add a `required_u32` helper (wrap `required_u64` + `u32::try_from`) and use it here so the MCP boundary enforces the same contract the plugin does.

Also: once this branch is rebased onto main, the spec name `"pilot.screenshot"` collides with the existing bridge spec `"screenshot"` after `namespaced_tool_name` lifts both to `pilot.screenshot`. See main review for the rename/unify options.

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -1365,6 +1384,28 @@ fn ipc_schema() -> Arc<JsonObject> {
     )
 }
 
+fn pilot_screenshot_schema() -> Arc<JsonObject> {
+    object_schema(
+        props([
+            (
+                "window_id",
+                integer_prop("Native window id (e.g. `CGWindowID` on macOS) to capture."),
```
mpiton: Schema advertises `integer` with no bounds; the plugin contract is u32. Tighten to `{ "type": "integer", "minimum": 0, "maximum": 4294967295 }` so MCP clients see the right range without a JSON-RPC round-trip.

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -1665,6 +1707,120 @@ mod tests {
...
+            );
+            let mut bytes = serde_json::to_vec(&response).expect("serialize response");
+            bytes.push(b'\n');
+            writer.write_all(&bytes).await.expect("write response");
+        });
+
+        let pilot = PilotMcpServer::new(Some(socket.clone()), None);
+        let mut args = Map::new();
+        args.insert("window_id".to_owned(), json!(42_u32));
+        args.insert("output_path".to_owned(), json!("/tmp/out.png"));
```
mpiton, line 1888: Test doesn't exercise the optional `format` field, so the dispatch logic at line 255 has no coverage and a typo like `payload["fmt"] = ...` would ship undetected. Insert `args.insert("format".to_owned(), json!("png"));` here and assert `params["format"] == json!("png")` in the mock server handler.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -152,9 +153,21 @@ pub(crate) async fn dispatch(
...
+        // The bare `screenshot` JSON-RPC method has been the bridge-side
+        // html-to-image path since v0.1 (returns a base64 PNG data URL). The
+        // new native contract ships under `pilot.screenshot` to avoid breaking
+        // those callers. As a convenience, a `screenshot` request that carries
+        // the native-shape `output_path` param is dispatched to the native
+        // handler too — old callers (no `output_path`) keep getting the
+        // bridge result.
         "screenshot" => {
-            handle_eval_method(method, params, engine, eval_fn, win, SCREENSHOT_TIMEOUT).await
+            if params.and_then(|p| p.get("output_path")).is_some() {
```
mpiton: Dispatch-on-shape mixes two contracts under one method name and is undiscoverable from the MCP surface — the bare `screenshot` MCP spec at `mcp.rs:242-249` only forwards `selector`, so this branch is only reachable from raw JSON-RPC and scenario callers. A bridge-era caller that experimentally adds `output_path` gets `INVALID_OUTPUT_PATH` instead of the historical base64 PNG, silently changing the contract.

Recommend dropping this branch and keeping `pilot.screenshot` (line 170) as the only native entry. If forward-compat under the bare name is a hard requirement, return a structured deprecation warning when `output_path` is sniffed but ignored.

crates/tauri-plugin-pilot/src/screenshot/cgwindow.rs
```diff
@@ -0,0 +1,82 @@
...
+                })?;
+        if row_end > src.len() {
+            return Err(ScreenshotError::CaptureFailed {
+                message: "CGImage row out of bounds".to_owned(),
+            });
+        }
+        let row = &src[row_start..row_end];
+        for px in row.chunks_exact(4) {
+            // BGRA -> RGBA
+            rgba.extend_from_slice(&[px[2], px[1], px[0], px[3]]);
```
mpiton: `CGWindowListCreateImage` returns BGRA with `kCGImageAlphaPremultipliedFirst`. Writing the buffer as `ColorType::Rgba8` tells PNG consumers the alpha is straight, so semi-transparent pixels render darker than the source for anyone doing standard compositing. Either add a demultiply pass (`r = r * 255 / a` with `a == 0` guard) before `save_buffer`, or document the premultiplied output in the function doc so downstream pixel-diff pipelines compensate.

crates/tauri-plugin-pilot/src/screenshot/cgwindow.rs
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    let w = u32::try_from(width).map_err(|_| ScreenshotError::CaptureFailed {
```
mpiton: `try_from(width)` and `try_from(height)` validate after `Vec::with_capacity(total)` and the full BGRA->RGBA loop have already run. Move both checks above line 39 so dimensions that don't fit `u32` fail fast without the allocation.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+use std::path::Path;
+
+use serde_json::Value;
+#[cfg(target_os = "macos")]
+use serde_json::json;
+
+use crate::protocol::RpcError;
+
+/// Numeric JSON-RPC error code for invalid params per the JSON-RPC 2.0 spec.
+const RPC_INVALID_PARAMS: i32 = -32602;
```
mpiton: `crate::protocol` already owns the JSON-RPC error vocabulary. Importing from there avoids the duplicate constants here and keeps numeric codes in one place if the protocol module ever moves to typed errors.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+const RPC_INTERNAL_ERROR: i32 = -32603;
+
+/// Domain error codes carried in the JSON-RPC `error.data.error` field.
+///
+/// Kept centralized so tests, the implementation, and the docs cannot drift.
+pub(crate) mod codes {
+    pub(crate) const INVALID_OUTPUT_PATH: &str = "INVALID_OUTPUT_PATH";
+    pub(crate) const UNSUPPORTED_FORMAT: &str = "UNSUPPORTED_FORMAT";
+    #[cfg_attr(
+        not(target_os = "macos"),
```
mpiton, line 28: Other domain codes live in the `codes` module, but `"INVALID_PARAMS"` is a literal at lines 85, 107, 115, 123. Add `pub(crate) const INVALID_PARAMS: &str = "INVALID_PARAMS";` and swap the call sites.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+                "output_path \"{}\" has no parent directory",
+                output_path.display()
+            ),
+            Value::Null,
+        )
+    })?;
+    // Refuse to create parent directories: an absolute path beneath a
+    // non-existent directory could be an attacker-controlled prefix on a
+    // shared host. Caller must pre-create the destination directory.
+    if !parent.as_os_str().is_empty() && !parent.is_dir() {
```
mpiton, line 203: nit: no check that `output_path` itself is a symlink. On a single-UID dev box that's fine, but if the socket is ever shared a caller could point at a symlink to anywhere-writable. `if output_path.is_symlink() { return Err(rpc_error(... INVALID_OUTPUT_PATH ...)) }` after the absolute-path check is a one-liner.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+
+    let metadata = match read_png_metadata(&tmp_path) {
+        Ok(m) => m,
+        Err(err) => {
+            cleanup_tmp(&tmp_path);
+            return Err(err);
+        }
+    };
+
+    let logical_bounds = get_window_bounds(window_id).unwrap_or(WindowBounds {
```
mpiton: `list_layer_zero_windows()` at line 233 and `get_window_bounds(window_id)` here both call `CGWindowListCopyWindowInfo`. Two separate kernel snapshots for data already in hand. Returning the bounds from the first enumeration drops one call and removes the TOCTOU window between them.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+        ));
+    }
+
+    // The other `ScreenshotBackend` variants never reach this point because
+    // the dispatcher in `capture_with_fallback` already maps them to error
+    // responses, so we only need to encode the two backends that produce a
+    // success result.
+    let backend_label = match backend {
+        ScreenshotBackend::ScreencaptureProbe => "screencapture",
+        _ => "cgwindow",
```
mpiton: `_ => "cgwindow"` silently absorbs `WkWebViewSnapshot` and `PlatformUnsupported` today and will silently mislabel any new variant tomorrow. Replace with explicit arms; for variants `capture_with_fallback` maps to errors, `unreachable!("capture_with_fallback handles these as errors")` is fine. Cleaner: put the label on the enum with `fn to_label(&self) -> &'static str` so dispatch and label can't drift.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+/// Compute the temp-write target alongside the final output path. Embedding
+/// the pid avoids two concurrent captures racing on the same final path from
+/// stomping each other's tmp file before the rename.
+#[cfg(target_os = "macos")]
+fn tmp_path_for(final_path: &Path) -> std::path::PathBuf {
+    let mut name = final_path
+        .file_name()
+        .map(std::ffi::OsString::from)
+        .unwrap_or_default();
+    name.push(format!(".tmp.{}", std::process::id()));
```
mpiton: Race condition: same `output_path` + same `process::id()` across two concurrent `pilot.screenshot` calls resolves to the same tmp path. Each socket connection is `tokio::spawn`'d and these calls go through `spawn_blocking`, so two callers will interleave writes and one `cleanup_tmp` removes the other's file before its rename can run. The atomic-write contract breaks.

Fix with a process-local counter:

```rust
use std::sync::atomic::{AtomicU64, Ordering};
static COUNTER: AtomicU64 = AtomicU64::new(0);
let seq = COUNTER.fetch_add(1, Ordering::Relaxed);
name.push(format!(".tmp.{}.{seq}", std::process::id()));
```

Worth a regression test that fires two concurrent captures at the same path.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+    final_path.with_file_name(name)
+}
+
+/// Remove the tmp file if it exists. Errors during cleanup are intentionally
+/// swallowed because the caller is already returning an error and we do not
+/// want to mask the original cause.
+#[cfg(target_os = "macos")]
+fn cleanup_tmp(path: &Path) {
+    let _ = std::fs::remove_file(path);
```
mpiton: `let _ = std::fs::remove_file(path)` drops cleanup errors silently and leaves stale PNGs on disk with no trace:

```rust
if let Err(e) = std::fs::remove_file(path) {
    tracing::warn!(path = %path.display(), error = %e, "failed to remove tmp screenshot");
}
```

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -0,0 +1,578 @@
...
+        rpc_error(
+            RPC_INTERNAL_ERROR,
+            codes::CAPTURE_FAILED,
+            format!("failed to stat captured PNG {}: {err}", path.display()),
+            Value::Null,
+        )
+    })?;
+    let byte_size = metadata.len();
+
+    let img = image::open(path).map_err(|err| {
```
mpiton: `image::open` fully decodes the PNG into a `DynamicImage` just to read width and height. At 4K Retina that's ~33MB allocated per call for two `u32` values. Switch to `image::ImageReader::open(path)?.into_dimensions()` which reads only the PNG IHDR. Error paths stay identical.

crates/tauri-plugin-pilot/src/screenshot/probe.rs
```diff
@@ -0,0 +1,44 @@
...
+    if screencapture_probe_allows_window_capture() {
+        ScreenshotBackend::ScreencaptureProbe
+    } else {
+        ScreenshotBackend::CgWindowList
+    }
+}
+
+fn screencapture_probe_allows_window_capture() -> bool {
+    std::process::Command::new("screencapture")
+        .args(["-l", "0", "/dev/null"])
```
mpiton: `kCGNullWindowID` (0) isn't documented as a valid target for `screencapture -l`; the exit code varies across macOS releases. The capture pipeline survives because `capture_with_fallback` retries on CGWindow, but `backend` in the success response can be wrong on certain versions (TCC granted, probe says false, caller sees `backend: "cgwindow"`). `CGPreflightScreenCaptureAccess` from CoreGraphics is the documented permission probe.

crates/tauri-plugin-pilot/src/screenshot/window_id.rs
```diff
@@ -0,0 +1,274 @@
...
+    use core_foundation::dictionary::CFDictionary;
+    use core_foundation::number::CFNumber;
+    use core_foundation::string::CFString;
+    use core_graphics::window::{
+        CGWindowListCopyWindowInfo, kCGNullWindowID, kCGWindowLayer,
+        kCGWindowListExcludeDesktopElements, kCGWindowListOptionOnScreenOnly, kCGWindowName,
+        kCGWindowNumber, kCGWindowOwnerName,
+    };
+
+    let raw = unsafe {
```
mpiton, line 70: Every `unsafe` block in this file (here, plus `get_window_bounds` and `get_window_id`) needs a `// SAFETY:` comment justifying soundness. Suggested phrasing for this one:

```rust
// SAFETY: CGWindowListCopyWindowInfo is thread-safe and requires a
// WindowServer connection, which Tauri establishes at startup. The null
// return is checked on the next line.
let raw = unsafe { ... };
```

Same treatment for `wrap_under_create_rule` (raw non-null and Create-rule owned by this call) and the `wrap_under_get_rule` calls on the `kCGWindow*` constants (static `CFStringRef` with effectively infinite retain count; CFRetain + Drop's CFRelease stay balanced).

crates/tauri-plugin-pilot/src/screenshot/window_id.rs
```diff
@@ -0,0 +1,274 @@
...
+    let windows: CFArray<CFDictionary<CFString, CFType>> =
+        unsafe { TCFType::wrap_under_create_rule(raw) };
+
+    let owner_key = unsafe { CFString::wrap_under_get_rule(kCGWindowOwnerName) };
+    let title_key = unsafe { CFString::wrap_under_get_rule(kCGWindowName) };
+    let layer_key = unsafe { CFString::wrap_under_get_rule(kCGWindowLayer) };
+    let number_key = unsafe { CFString::wrap_under_get_rule(kCGWindowNumber) };
+
+    let mut out: Vec<DiscoveredWindow> = Vec::new();
+    for idx in 0..windows.len() {
```
mpiton: nit: `for idx in 0..windows.len()` then `windows.get(idx)` is C-style. `CFArray` implements `Iterator`, so `for dict in windows.iter()` reads cleaner and the cap check becomes `.take(AVAILABLE_WINDOWS_CAP)`.

crates/tauri-plugin-pilot/src/screenshot/window_id.rs
```diff
@@ -0,0 +1,274 @@
...
+///
+/// Walks the on-screen window list, skipping desktop elements, and returns the
+/// first layer-0 window whose owner equals `owner` and whose title contains
+/// `title` (when provided).
+///
+/// # Errors
+///
+/// Returns [`ScreenshotError::CaptureFailed`] when the `CoreGraphics` window
+/// list call returns null or no layer-0 window matches the filter.
+#[allow(
```
mpiton: `#[allow(dead_code)]` on a function the PR doesn't call yet. The workspace runs `cargo clippy -- -D warnings`, so the suppression hides real signal from CI. Two options: delete the fn and reintroduce in the milestone PR, or gate behind a cargo feature so the compiler elides it without `allow`. Same treatment recommended for the `WkWebViewSnapshot` and `PlatformUnsupported` arms in `probe.rs:9, 16` and the stubs in `native.rs:14`, `error.rs:9-11`, `ipc.rs:31, 36`.

mpiton (changes requested):
Thanks for the thorough write-up and the careful self-review in the PR description. This is a meaty change with a clear contract surface (path-only, atomic write, TCC-as-metadata, discovery payload in `WINDOW_NOT_FOUND`) and good unit coverage on the validators.

Marking as request-changes for a few items that need attention before merge. Main ones:

1. **Rebase needed**. This branched from `66aa823`, before PR #99 landed the `pilot.*` MCP namespacing on main (`ffb2a48`). On main, the bridge spec `name: "screenshot"` already advertises as `pilot.screenshot` via `namespaced_tool_name`, so adding a second spec literally named `pilot.screenshot` produces a duplicate registration, and the new dispatch arm at `mcp.rs:250` becomes dead code (the prefix is stripped by `normalize_tool_name` before matching). Two options: rename the new tool (`screenshot_native` lifts to `pilot.screenshot_native`), or unify both entries behind bare `screenshot` with a single dispatch arm that branches on `output_path`. Worth doing first so the next review pass sees the post-rebase surface.

2. **Atomic-write race** (`ipc.rs:343`) and **`window_id` boundary type** (`mcp.rs:252`) — both flagged by bots and both valid. Details inline.

3. **Unsafe FFI blocks need `// SAFETY:` comments** in `window_id.rs`. Standard for any `unsafe` block; once a reviewer enables `clippy::undocumented_unsafe_blocks` it would fail CI.

Lower-priority items inline (premultiplied alpha, dispatch-on-shape footgun, fragile screencapture probe, a handful of nits).

On the CodeRabbit `row_start = y * bytes_per_row` flag in `cgwindow.rs:51`: pushing back on that one. On a 64-bit target the multiplication can't overflow because `y * bytes_per_row` is bounded by the CFData buffer length, which by definition fits in `usize`. The two `checked_mul` calls above already guard the only unconstrained arithmetic (`Vec::with_capacity`). Happy to add a defensive `checked_mul` anyway if you'd rather have belt-and-braces.

On the CHANGELOG entry: add one sentence about non-macOS behavior (responds `PERMISSION_DENIED` with `screenshot is only available on macOS in this release` per `ipc.rs:212-217`) so cross-platform operators don't have to discover it empirically.

On the >150-line files you flagged in the description: I'm fine with `ipc.rs` and `window_id.rs` staying as-is for this PR. The cohesion is real and a forced split would just move the seams. If you do split `ipc.rs` later, the four-way breakdown you sketched looks right.

Really appreciate the discipline on the test coverage and the structured `WINDOW_NOT_FOUND` payload. Once the items above land this is good to merge.

Naoray:
Thanks for the thorough review. All items addressed — branch rebased onto current `main` (76566a1) and amended via [`db442e8`](https://github.com/EmpireTwo/tauri-pilot/commit/db442e8) `refactor(plugin): address PR #100 review feedback`. Force-pushed (rebase rewrote the three prior commits onto post-#99 main, no functional change to those).

## Rebase + namespace collision

You were right about the post-#99 collision — `build_tools()` now wraps every spec via `namespaced_tool_name`, so a literal `name: "pilot.screenshot"` would have produced a duplicate registration and the new dispatch arm would have been dead.

Went with option (a) from your review — renamed rather than dispatch-on-shape, since the latter is the footgun you flagged at handler.rs:164. Net result:

- MCP tool: `screenshot_native` ToolSpec → advertised as `pilot.screenshot_native` via `namespaced_tool_name`.
- JSON-RPC method: `screenshot_native` on the plugin side. 1:1 mapping with the MCP tool name.
- Bare `screenshot` is bridge-only again — no `output_path` sniffing. `test_dispatch_screenshot_routes_to_bridge_regardless_of_params` at `handler.rs:1193-1215` pins that contract.

## Inline items addressed

**`crates/tauri-pilot-cli/src/mcp.rs`**
- `required_u32` helper at `mcp.rs:1115-1118` wraps `required_u64` + `u32::try_from`; used in the new dispatch arm so over-u32 values are rejected at the MCP boundary instead of the plugin.
- `window_id` schema at `mcp.rs:1417-1424` tightened to explicit `{"type":"integer","minimum":0,"maximum":4294967295,"description":"..."}`.
- `pilot_screenshot_forwards_native_params_to_jsonrpc` now inserts `format: "png"` (line 1889) and the mock asserts `params["format"] == json!("png")` (line 1866).

**`crates/tauri-plugin-pilot/src/handler.rs`**
- Dispatch-on-shape arm gone. Bare `screenshot` at `handler.rs:161-163` routes to bridge eval unconditionally. New `"screenshot_native"` arm at `handler.rs:164`.

**`crates/tauri-plugin-pilot/src/screenshot/cgwindow.rs`**
- BGRA premultiplied → straight RGBA demultiply pass at `cgwindow.rs:85-107` with `alpha == 0 → (0,0,0)` guard. PNG output is now standard straight-alpha so downstream pixel-diff pipelines composite correctly.
- `# Output color` doc section at `cgwindow.rs:7-13` documenting the demultiply contract.
- `u32::try_from(width|height)` moved to `cgwindow.rs:49-54`, ahead of the `Vec::with_capacity` allocation at `cgwindow.rs:66`. Over-u32 dimensions now fail fast without allocation.
- `row_start = y * bytes_per_row` left unchecked per your pushback; inline comment at `cgwindow.rs:67-71` documents the CFData-length bound so the next reviewer doesn't re-trigger the same flag.

**`crates/tauri-plugin-pilot/src/screenshot/ipc.rs`**
- `RPC_INVALID_PARAMS` and `RPC_INTERNAL_ERROR` moved to `crate::protocol` (now at `protocol.rs:11,16`) and imported here. No more duplicated numeric error vocabulary.
- `codes::INVALID_PARAMS` const added at `ipc.rs:24`; the four `"INVALID_PARAMS"` literals at `ipc.rs:83,105,113,121` now use the const.
- Symlink rejection at `ipc.rs:167-177`, after the absolute-path check at `ipc.rs:151-161`.
- Single-pass window enumeration: replaced `list_layer_zero_windows()` + `get_window_bounds()` with `enumerate_layer_zero_windows(target) -> EnumeratedLayerZeroWindows` (`window_id.rs:48-50`, struct at `window_id.rs:25-28`). One `CGWindowListCopyWindowInfo` snapshot per request, TOCTOU window between the two old calls is gone.
- Exhaustive `match backend` at `ipc.rs:325-331` — `WkWebViewSnapshot | PlatformUnsupported => unreachable!("capture_with_fallback returns Err for these variants")`. Future enum additions will fail to compile rather than silently mislabel.
- Tmp uniqueness via `static TMP_COUNTER: AtomicU64` at `ipc.rs:369-384`. Tmp path is now `<output>.tmp.<pid>.<counter>` (line 382), so two concurrent `pilot.screenshot_native` calls to the same `output_path` get distinct staging files — the rename remains the atomic step but the per-request tmp file no longer races.
- `cleanup_tmp` at `ipc.rs:389-400` now logs via `tracing::warn!` (with `NotFound` suppressed). `tracing` was already in the workspace deps.
- `image::open` replaced with `image::ImageReader::open(..)?.into_dimensions()` at `ipc.rs:475-493`. Reads only the PNG IHDR — saves the ~33 MiB allocation on 4K Retina captures.

**`crates/tauri-plugin-pilot/src/screenshot/probe.rs`**
- `screencapture -l 0 /dev/null` shell-out gone. `core_graphics::access::ScreenCaptureAccess.preflight()` at `probe.rs:46-48` is the new probe — the documented CoreGraphics permission check, no exit-code variance across macOS releases. `backend` in the success response is now accurate even when TCC is granted on macOS versions where the old shell probe was unreliable.

**`crates/tauri-plugin-pilot/src/screenshot/window_id.rs`**
- Every `unsafe { ... }` block now carries a `// SAFETY:` justification (lines 63-69, 81-84, 88-90).
- C-style index loop replaced with `for dict in windows.iter()` at `window_id.rs:100`.
- `fn get_window_id` deleted (was `#[allow(dead_code)]`); will reintroduce alongside the consumer that needs it. Recoverable from history.

**`CHANGELOG.md`**
- Non-macOS behavior documented at `CHANGELOG.md:22-26`: "On Linux and Windows, `pilot.screenshot_native` is registered but every call responds with `PERMISSION_DENIED`..."

## Skipped with rationale

- **`row_start = y * bytes_per_row` `checked_mul` (cgwindow.rs:67)** — per your pushback. Bounded by CFData buffer length, can't overflow on 64-bit. Replaced with the inline-comment bound documentation above.
- **`ipc.rs` / `window_id.rs` >150 LOC split** — per your "fine as-is for this PR" line. Will revisit if cohesion ever stops being the right call.

## Verification on the updated branch

- `cargo fmt --all -- --check` clean.
- `cargo clippy --workspace --all-targets -- -D warnings` clean (zero warnings, zero errors).
- `cargo test --workspace` → 279 passed (was 276 pre-amend; +3 from format-forwarding assertion + new bridge-only regression test + the single-pass enumeration unit test).

CI re-running on the new HEAD now. Will follow up if anything else surfaces.

mpiton:
Walked the diff on db442e8 against each of the 19 inline items; all addressed cleanly, and the rebase + `screenshot_native` rename closed the namespacing collision the right way. CI is green, the demultiply pass looks correct, and the `EnumeratedLayerZeroWindows` refactor reads well.

CodeRabbit caught two real items on the new code that I'd address before merging:

1. The `WINDOW_NOT_FOUND` gate at `ipc.rs:277` is still `discovered.iter().any(...)`, but `available` is hard-capped at 20 while `target_bounds` keeps populating past the cap (`window_id.rs:148-159`). So a valid `window_id` sitting at the 21st+ layer-0 entry will incorrectly return `WINDOW_NOT_FOUND`. Swap to `if target_bounds.is_none()`;  that's what the refactor was set up to enable.
2. The directory-target nit is a quick win; rejecting `output_path.is_dir()` early gives callers `INVALID_OUTPUT_PATH` instead of a late `CAPTURE_FAILED` at rename.

Once those two land I'll dismiss the request-changes. The skip rationale on `row_start` checked_mul and the file split is fine.

mpiton (approved):

mpiton:
Thanks !

PR #108, opened 2026-05-27
fix: restore eval completion and native screenshots

Naoray:
## Summary

- restore eval result delivery on Tauri 2.11 / wry 0.55 by using WebView eval callbacks instead of JS-to-Tauri IPC callbacks
- keep async eval and bridge calls awaited, and make bridge bootstrap idempotent
- add macOS screenshot_native support with path-only PNG output and CLI/MCP routing
- avoid windows.list URL panics by omitting unavailable webview URLs from the listing path

## Verification

- cargo fmt --all -- --check
- cargo clippy --workspace -- -D warnings
- cargo test --workspace
- EmpireTwo fork CI passed on the same driver head across macOS, Ubuntu, Windows, and CodeQL.

## API Surface

The intended stable surface for downstream harness use is:

- JSON-RPC transport methods: eval, html, snapshot
- windows.list
- screenshot_native with params: window_id, output_path, format optional png
- CLI route: tauri-pilot screenshot_native --window-id <id> --output <path> [--format png]

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to capture native window screenshots by platform window ID (configurable output path and format).

* **Bug Fixes**
  * Bridge initialization is now idempotent to avoid duplicate instrumentation.
  * Restored reliable eval result delivery on newer runtimes and prevented window-listing panics when a webview URL can't be reported.
  * Ensured screenshot temp files preserve PNG extension.

* **Refactor**
  * Switched eval callback flow to a callback-style payload with improved macOS handling.

* **Documentation**
  * Updated architecture and permission docs and changelog to reflect these changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mpiton/tauri-pilot/pull/108?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/tauri-plugin-pilot/src/lib.rs
+++ b/crates/tauri-plugin-pilot/src/lib.rs
@@ -141,7 +141,7 @@ fn make_eval_fn<R: tauri::Runtime>(app: &tauri::AppHandle<R>, engine: EvalEngine
         let engine = engine.clone();
         #[cfg(target_os = "macos")]
         {
-            return eval_with_webkit_callback(&target, script, engine);
+            eval_with_webkit_callback(&target, script, engine)
         }
         #[cfg(not(target_os = "macos"))]
         target
--- a/crates/tauri-plugin-pilot/src/lib.rs
+++ b/crates/tauri-plugin-pilot/src/lib.rs
@@ -28,7 +28,7 @@ use std::sync::Arc;
 use tauri::Manager;
 
 #[cfg(all(any(unix, windows), debug_assertions))]
-const BRIDGE_JS: &str = concat!(
+pub(crate) const BRIDGE_JS: &str = concat!(
     include_str!("../js/vendor/html-to-image.iife.js"),
     "\n",
     include_str!("../js/bridge.js"),
@@ -52,6 +52,11 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
     {
         tauri::plugin::Builder::new("pilot")
             .js_init_script(BRIDGE_JS.to_owned())
+            .on_webview_ready(|webview| {
+                if let Err(err) = webview.eval(BRIDGE_JS) {
+                    tracing::warn!(error = %err, "failed to inject tauri-pilot bridge on webview ready");
+                }
+            })
             .setup(|app, _api| {
                 let engine = EvalEngine::new();
                 app.manage(engine.clone());
@@ -59,7 +64,7 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 let identifier = sanitize_identifier(&app.config().identifier);
                 let socket_path = server::socket_path(&identifier);
 
-                let eval_fn = make_eval_fn(app);
+                let eval_fn = make_eval_fn(app, engine.clone());
                 let list_fn = make_list_fn(app);
                 let focus_fn = make_focus_fn(app);
 
@@ -82,7 +87,10 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
 
                 Ok(())
             })
-            .invoke_handler(tauri::generate_handler![handler::__callback])
+            .invoke_handler(tauri::generate_handler![
+                handler::callback,
+                handler::__callback
+            ])
             .build()
     }
 }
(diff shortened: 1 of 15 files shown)


Naoray:
Addressed the CodeRabbit permission feedback in b3c05f9: the manual eval-callback permission is now `allow-eval-callbacks`, while the autogenerated command permission remains the only `allow-callback`. Verified locally with duplicate identifier/schema checks, `cargo fmt --all -- --check`, `cargo clippy --workspace -- -D warnings`, and `cargo test --workspace`.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -52,14 +52,19 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
     {
         tauri::plugin::Builder::new("pilot")
             .js_init_script(BRIDGE_JS.to_owned())
+            .on_webview_ready(|webview| {
```
mpiton, line 55: nit: the bridge now gets injected three ways: `js_init_script`, `on_webview_ready`, and the `ensure_bridge_script` wrapper on each eval. The `__PILOT__` guard makes it safe, but `ensure_bridge_script` re-runs the whole html-to-image bundle whenever `__PILOT__` is missing, even for `ipc` which never touches `__PILOT__`. If `on_webview_ready` is reliable you could probably drop the per-call wrap. Just noting the redundancy.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -113,27 +121,112 @@ fn sanitize_identifier(raw: &str) -> String {
...
 }
 
+#[cfg(all(target_os = "macos", debug_assertions))]
+fn eval_with_webkit_callback<R: tauri::Runtime>(
+    target: &tauri::WebviewWindow<R>,
+    script: String,
+    engine: EvalEngine,
+) -> Result<(), String> {
+    target
+        .with_webview(move |platform| unsafe {
```
mpiton, line 167: nit: every other unsafe block in the crate carries a `// SAFETY:` comment (see `server/unix.rs`, `screenshot/window_id.rs`). Can you add one here? Roughly: `platform.inner()` hands back the WKWebView pointer wry owns, it's valid for the `with_webview` closure, and the RcBlock is heap-allocated so WebKit's Block_copy keeps it alive past this scope. Took me a while to convince myself it wasn't a use-after-free, a comment would save the next reader that trip.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -113,27 +121,112 @@ fn sanitize_identifier(raw: &str) -> String {
...
+            use objc2::{AnyThread, runtime::AnyObject};
+            use objc2_foundation::{
+                NSError, NSJSONSerialization, NSJSONWritingOptions, NSString, NSUTF8StringEncoding,
+            };
+            use objc2_web_kit::WKWebView;
+
+            let webview: &WKWebView = &*platform.inner().cast();
+            let handler =
+                block2::RcBlock::new(move |value: *mut AnyObject, error: *mut NSError| {
+                    if !error.is_null() {
```
mpiton, line 177: When the NSError is non-null we log and return without resolving the pending id, so the caller waits out the full 10s timeout instead of getting the error back. Same when `value` is null or serialization fails, `payload` stays empty and `handle_eval_callback_payload` bails before resolving. Could you resolve the id with an error on those branches? A syntax error in an eval script would hit this, and a fast "JS error" beats a 10s hang. Minor since this is debug-only tooling, but cheap to fix.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -174,7 +267,7 @@ fn make_list_fn<R: tauri::Runtime>(app: &tauri::AppHandle<R>) -> ListWindowsFn {
             .map(|(label, wv)| {
                 serde_json::json!({
                     "label": label,
-                    "url": wv.url().map(|u| u.to_string()).unwrap_or_default(),
+                    "url": "",
```
mpiton: q: what's the actual panic here? `wv.url()` returns a Result and the old code already had `unwrap_or_default()`, so I'm curious whether it panics inside `url()` itself on 0.55. Hardcoding "" drops the URL everywhere, including Linux/Windows where it probably still works, and it leaves a blank column in `tauri-pilot windows` (`output.rs` `format_windows` renders the field). Could we keep the value where `url()` succeeds and only fall back to empty on failure? A `catch_unwind` around the call would do it if it's a genuine panic. Not blocking, I'd just rather not lose the field on platforms that don't need the workaround.

mpiton (changes requested):
Thanks for digging into this. The eval breakage on 2.11/wry 0.55 has been biting us and your callback approach is the right fix. I traced the new path on Linux and the macOS native block, both look correct. I also checked the unsafe WKWebView cast against tauri-runtime-wry's source and it's the same pattern wry uses internally (RcBlock gets heap-allocated, WebKit copies the completion block), so no use-after-free there.

A few things before I merge, mostly housekeeping. I left inline notes on the unsafe SAFETY comment, the windows.list url change, the macOS error path, and the triple bridge injection.

Two more that aren't inline because the files aren't in the diff:

**CHANGELOG.md** — we keep it updated on every PR. Could you add entries under `[Unreleased]`? Something like:

```
### Fixed
- Restore eval result delivery on Tauri 2.11 / wry 0.55. Results now use native
  WebView eval callbacks instead of the JS-to-Tauri `__callback` IPC, which stopped
  dispatching from `webview.eval`-injected scripts. (#108)
- `windows.list` no longer panics on wry 0.55. (#108)

### Added
- `screenshot_native` CLI subcommand to capture a native window PNG by window id. (#108)
- Idempotent bridge bootstrap, re-injection is now a no-op. (#108)
```

Tweak the wording however you like.

**Docs** — this PR changes how eval results come back, so the old `__callback` IPC description is now stale. The architecture guide (`docs/src/content/docs/guides/architecture.md`, around L94-95) still describes the IPC round-trip, and three doc comments in `eval.rs` need a refresh: the struct doc (L23), `resolve` (L72 still says "Called from the IPC __callback handler"), and the first line of `wrap_script` (L90).

Nice work overall, this is a solid PR.

crates/tauri-plugin-pilot/src/lib.rs
Naoray: Addressed in d7835d6. windows.list now keeps the URL when wv.url() succeeds and falls back to an empty string only if the call errors or panics under wry 0.55.

crates/tauri-plugin-pilot/src/lib.rs
Naoray: Addressed in d7835d6 by dropping the per-call bridge wrapper. The bridge is still injected via js_init_script/on_webview_ready, and eval/ipc calls now execute without re-running the bundle per request.

crates/tauri-plugin-pilot/src/lib.rs
Naoray: Addressed in d7835d6 with a SAFETY comment covering the wry-owned WKWebView pointer lifetime and the heap-allocated RcBlock/WebKit copy behavior.

crates/tauri-plugin-pilot/src/lib.rs
Naoray: Addressed in d7835d6. The macOS native callback now resolves the pending eval id with an error when WebKit returns NSError, a null value, or a non-serializable result, so callers fail fast instead of waiting for the timeout.

mpiton (approved):

mpiton:
Thanks !

Issue #16, opened 2026-04-02
feat: MCP server — expose tauri-pilot as a Model Context Protocol tool


mpiton:
## Summary
Expose tauri-pilot as an MCP (Model Context Protocol) server so Claude Code and other AI agents can use it natively as a tool without shelling out to the CLI.

## Benefits
- Zero-latency tool calls (no process spawn overhead)
- Structured input/output (no string parsing)
- Auto-discovery by MCP-compatible agents
- Could replace the CLI entirely for AI workflows

## Implementation
- `tauri-pilot mcp` starts an MCP server (stdio transport)
- Each CLI command becomes an MCP tool: `snapshot`, `click`, `fill`, `logs`, etc.
- Configure in `.claude/settings.json` or `mcp.json`

```json
{
  "mcpServers": {
    "tauri-pilot": {
      "command": "tauri-pilot",
      "args": ["mcp"]
    }
  }
}
```

## Priority
🔮 Low (but high impact) — makes tauri-pilot a first-class AI agent tool.

Issue #15, opened 2026-04-02
feat: record/replay — capture interactions as replayable scripts


mpiton:
## Summary
Record user/agent interactions and generate replayable test scripts. The agent can create E2E tests by interacting with the app, then export the recording.

## Proposed CLI

```bash
tauri-pilot record start                   # start recording
tauri-pilot click @e3                      # interactions are captured
tauri-pilot fill @e2 "test"
tauri-pilot record stop --output test.json # save recording
tauri-pilot replay test.json               # replay the recording
tauri-pilot replay test.json --export sh   # export as shell script
```

## Output format (test.json)
```json
[
  {"action": "click", "ref": "e3", "timestamp": 0},
  {"action": "fill", "ref": "e2", "value": "test", "timestamp": 1200},
  {"action": "assert", "ref": "e1", "text": "Results", "timestamp": 2500}
]
```

## Priority
🔮 Low — powerful differentiator for test generation workflows.

Issue #14, opened 2026-04-02
feat: multi-window support


mpiton:
## Summary
Support Tauri apps with multiple windows. Currently tauri-pilot assumes a single WebView.

## Proposed CLI

```bash
tauri-pilot windows                       # list all windows
tauri-pilot snapshot -i --window settings # target a specific window
tauri-pilot click @e3 --window main       # interact with a specific window
```

## Implementation
- Plugin tracks all windows with labels
- JSON-RPC commands accept optional `window` parameter
- `windows.list` returns available windows with labels and URLs

## Priority
🔮 Low — needed for complex multi-window apps.

Issue #12, opened 2026-04-02
feat: localStorage/sessionStorage access


mpiton:
## Summary
Read and write browser storage from the CLI. Useful for AI agents inspecting persisted state, auth tokens, or modifying app configuration during testing.

## Proposed CLI

```bash
tauri-pilot storage get "auth_token"                  # read a key
tauri-pilot storage set "theme" "dark"                 # write a key
tauri-pilot storage list                               # dump all keys
tauri-pilot storage clear                              # clear all
tauri-pilot storage --session                          # use sessionStorage instead
tauri-pilot storage --json                             # structured output
```

## Priority
🔮 Low — nice-to-have for state inspection.

PR #140, opened 2026-07-29
chore(deps): upgrade docs to Astro 7 and sharp 0.35

mpiton:
Clears all 16 open Dependabot alerts. Every one of them was in `docs/` — the plugin and CLI crates are untouched.

## What moved

| Package | From | To | Alerts |
|---|---|---|---|
| astro | 6.3.2 | 7.1.5 | #20 #19 #18 #17 #16 #15 #11 #10 |
| sharp | 0.34.5 | 0.35.3 | #23 #22 |
| @astrojs/starlight | 0.38.5 | 0.41.5 | (peer: astro ^7.0.2) |
| js-yaml | 4.1.x | 4.3.0 | #24 #14 |
| svgo | 4.0.x | 4.0.2 | #21 |
| vite | 7.3.x | 8.1.5 | #12 #9 |
| esbuild | 0.27.7 | 0.28.1 | #7 |

Astro carried four XSS advisories (spread attribute names, `transition:*` directives on hydrated islands, View Transition animation properties) plus a host-header SSRF in the prerendered error page fetch. sharp inherited four libvips CVEs. The bottom four rows are transitive — they came along when the lockfile was regenerated, which was needed anyway since npm's incremental resolver wouldn't move Starlight off the astro ^6 peer.

## Testing

- `npm audit` — 0 vulnerabilities (was 16)
- `npm ci` reproduces the tree cleanly, matching what `deploy-docs.yml` runs
- `npm run build` — 8 pages, image optimization intact, same page set as the pre-upgrade baseline. Base path `/tauri-pilot/` and the Pagefind search index still generate correctly.

The pre-existing `Entry docs → 404 was not found.` line at the end of the build is unchanged from before the upgrade.

## One thing to note

Astro 7 requires node `>=22.12.0`. `deploy-docs.yml` pins `node-version: 22`, which resolves to the latest 22.x and satisfies it — no CI change needed, but worth knowing the floor moved.

## Summary by CodeRabbit

* **Security**
  * Updated documentation-site dependencies to address reported security vulnerabilities.
  * Resolved Dependabot security alerts, including issues affecting prerendered error pages and image processing.

* **Documentation**
  * Added changelog details covering the security updates and their scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- Upgrade the docs site to Astro 7.1.5 (from 6.3.2), Starlight 0.41.5 and sharp
+  0.35.3, clearing all 16 open Dependabot alerts. Astro carried four XSS issues
+  and a host-header SSRF in the prerendered error page fetch; sharp inherited
+  four libvips CVEs. Refreshing the lockfile also moved the transitive
+  js-yaml (4.3.0), svgo (4.0.2), vite (8.1.5) and esbuild (0.28.1) onto patched
+  versions. Docs-only — the plugin and CLI crates are untouched.
+
 ## [0.7.2] - 2026-06-10
 
 ### Added
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -8,38 +8,226 @@
       "name": "docs",
       "version": "0.0.1",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


mpiton:
Side note, not changed here: the `overrides: { devalue: "^5.8.1" }` in `docs/package.json` is now redundant. Astro 7.1.5 depends on `devalue@^5.8.1` directly and resolves to 5.8.2, so the override no longer pins anything the tree wouldn't already pick. Left it in place to keep this PR to the security bump — worth dropping in a follow-up.

Issue #31, opened 2026-04-05
Unix socket in /tmp with no permission restriction - any local user can execute arbitrary JS in webview


consigcody94:
Found via code audit. crates/tauri-plugin-pilot/src/server.rs:60-101. Socket at /tmp/tauri-pilot-*.sock with default permissions. No chmod, no auth. Local users can send eval/ipc/navigate commands.

consigcody94:
@mpiton That is a thorough fix. The three-layer approach (umask + chmod + peer_cred UID check) covers all the edge cases. Moving to XDG_RUNTIME_DIR is the right call since it is already 0700 per spec. The CLI-side resolve_socket filtering by UID is a nice addition that prevents the TOCTOU attack where a rogue socket is planted before the daemon starts.

mpiton:
Thanks for the audit this was a real gap. TOCTOU was exactly the scenario I was worried about when adding the CLI-side UID filter. Appreciate you taking the time to look through the code.

PR #145, opened 2026-09-08
feat(plugin): support Android via ADB

Darkatse:
## Summary

- Add Android device and emulator support via ADB from macOS or Linux, as a first step toward mobile support.

## Motivation

I would like to bring tauri-pilot to mobile Tauri apps. For projects such as [TauriTavern](https://github.com/Darkatse/TauriTavern) that run on both desktop and mobile, inspecting and testing the mobile WebView is an important part of development.

## Changes

The implementation uses an abstract Unix socket on Android, forwarded by ADB to a local Unix socket on the host. This lets the existing CLI, JSON-RPC handlers and JS bridge be reused. Most of the change is in socket address handling and peer authentication.

Windows hosts need a further step: the Windows CLI currently connects through named pipes, while this setup gives the host a Unix socket. One option would be to forward a loopback TCP port through ADB to the same Android socket and add TCP connection support to the CLI. Would you be open to that approach? If so, I can submit a separate PR for it.

I also plan to work on iOS support, starting with the simulator.

## Test Plan

- [x] Tested manually with a Tauri app (if applicable)

To check the approach, I used a small Tauri v2 app with a counter and a text input, built with `default-features = false`. From macOS, I connected to an Android emulator and a physical Android 16 device through ADB, then checked `ping`, snapshots, clicks, text input and assertions. I also checked the existing macOS socket connection.

`cargo check -p tauri-plugin-pilot --target aarch64-linux-android --no-default-features` also passes.

## Checklist

---
## Summary by cubic
Adds Android device and emulator support via ADB, letting tauri-pilot inspect and test mobile WebViews without changing the existing CLI, JSON-RPC handlers, or JS bridge. Android uses a randomized per-instance abstract socket name (identifier capped at 16 characters plus a 16-hex random suffix) discoverable from the app's startup logs, and connections are authorized via kernel peer credentials (app, root, or ADB shell).

**Notes**
- Windows hosts need a TCP forward; a separate PR will add that.
- iOS support is planned next, starting with the simulator.

<sup>Written for commit c758c6b9bb8c264cbde710c1ec1a7dd9ab86b10c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/145?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added Android device and emulator support through ADB socket forwarding.
  * Added secure connection authorization for Android app, root, and ADB shell access.

* **Documentation**
  * Added Android setup guidance, including configuration, socket forwarding, device selection, troubleshooting, and cleanup.
  * Documented Android limitations, including unsupported Windows hosts and keyboard injection; use text input commands instead.
  * Updated platform requirements and security information.

### DIFF
--- a/crates/tauri-plugin-pilot/src/lib.rs
+++ b/crates/tauri-plugin-pilot/src/lib.rs
@@ -39,6 +39,7 @@ pub(crate) const BRIDGE_JS: &str = concat!(
 /// On non-Unix, non-Windows platforms or in release builds, returns a no-op plugin.
 /// In debug builds on Unix, injects the JS bridge, stores an `EvalEngine`,
 /// and starts a Unix socket server at `$XDG_RUNTIME_DIR/tauri-pilot-{identifier}.sock` (falls back to `/tmp` if unavailable).
+/// Android uses the abstract Unix socket `tauri-pilot-{identifier}.sock`, reachable via ADB forwarding.
 /// In debug builds on Windows, starts a Named Pipe server at
 /// `\\.\pipe\tauri-pilot-{identifier}` and registers the instance under `%LOCALAPPDATA%\tauri-pilot\instances\`.
 #[must_use]
@@ -62,7 +63,6 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 app.manage(engine.clone());
 
                 let identifier = sanitize_identifier(&app.config().identifier);
-                let socket_path = server::socket_path(&identifier);
 
                 let eval_fn = make_eval_fn(app);
                 let list_fn = make_list_fn(app);
@@ -76,8 +76,9 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 // listener to tokio once it is already on the runtime.
                 #[cfg(unix)]
                 {
-                    let (listener, guard) = server::bind(&socket_path).map_err(|e| {
-                        tracing::error!(path = %socket_path.display(), "failed to bind socket: {e}");
+                    let address = server::socket_address(&identifier)?;
+                    let (listener, guard) = server::bind(&address).map_err(|e| {
+                        tracing::error!(?address, "failed to bind socket: {e}");
                         e
                     })?;
                     tauri::async_runtime::spawn(server::run(
@@ -98,7 +99,7 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 // the context of a Tokio 1.x runtime" (#115).
                 #[cfg(windows)]
                 tauri::async_runtime::spawn(server::run(
-                    socket_path,
+                    server::socket_path(&identifier),
                     engine,
                     Some(eval_fn),
                     Some(list_fn),
--- a/crates/tauri-plugin-pilot/src/lib.rs
+++ b/crates/tauri-plugin-pilot/src/lib.rs
@@ -37,9 +37,10 @@ pub(crate) const BRIDGE_JS: &str = concat!(
 /// Initialize the tauri-pilot plugin.
 ///
 /// On non-Unix, non-Windows platforms or in release builds, returns a no-op plugin.
-/// In debug builds on Unix, injects the JS bridge, stores an `EvalEngine`,
+/// In debug builds on Unix other than Android, injects the JS bridge, stores an `EvalEngine`,
 /// and starts a Unix socket server at `$XDG_RUNTIME_DIR/tauri-pilot-{identifier}.sock` (falls back to `/tmp` if unavailable).
-/// Android uses the abstract Unix socket `tauri-pilot-{identifier}.sock`, reachable via ADB forwarding.
+/// Android uses `tauri-pilot-{identifier}-{random}.sock` in the abstract namespace,
+/// reachable via ADB forwarding. The per-instance address is logged at info level.
 /// In debug builds on Windows, starts a Named Pipe server at
 /// `\\.\pipe\tauri-pilot-{identifier}` and registers the instance under `%LOCALAPPDATA%\tauri-pilot\instances\`.
 #[must_use]
@@ -76,7 +77,9 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 // listener to tokio once it is already on the runtime.
                 #[cfg(unix)]
                 {
-                    let address = server::socket_address(&identifier)?;
+                    let address = server::socket_address(&identifier).inspect_err(|e| {
+                        tracing::error!(identifier, "failed to build tauri-pilot socket address: {e}");
+                    })?;
                     let (listener, guard) = server::bind(&address).map_err(|e| {
                         tracing::error!(?address, "failed to bind socket: {e}");
                         e
--- a/crates/tauri-plugin-pilot/src/server/unix.rs
+++ b/crates/tauri-plugin-pilot/src/server/unix.rs
@@ -6,6 +6,7 @@ use crate::recorder::Recorder;
 
 use std::os::unix::fs::PermissionsExt;
 use std::os::unix::io::AsRawFd;
+use std::os::unix::net::SocketAddr;
 use std::sync::Arc;
 use tokio::net::UnixListener;
 
@@ -45,6 +46,7 @@ fn inode_from_raw_fd(fd: std::os::unix::io::RawFd) -> u64 {
 }
 
 /// Returns true if `path` is a directory owned by the current user with no group/world permissions.
+#[cfg(not(target_os = "android"))]
 fn is_private_dir(path: &std::path::Path) -> bool {
     use std::os::unix::fs::MetadataExt;
     match std::fs::metadata(path) {
@@ -59,6 +61,7 @@ fn is_private_dir(path: &std::path::Path) -> bool {
 
 /// Core implementation — accepts the XDG value directly so tests can call it without mutating
 /// the process environment.
+#[cfg(not(target_os = "android"))]
 fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
     if let Some(val) = xdg.filter(|v| !v.is_empty()) {
         let path = std::path::PathBuf::from(&val);
@@ -76,13 +79,21 @@ fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
 /// Returns the directory for the socket file.
 /// Prefers `$XDG_RUNTIME_DIR` when it is a private directory (owned by current user, no
 /// group/world access). Falls back to `/tmp` with a warning if the directory is not private.
+#[cfg(not(target_os = "android"))]
 fn socket_dir() -> std::path::PathBuf {
     socket_dir_from(std::env::var_os("XDG_RUNTIME_DIR"))
 }
 
-/// Build the full socket path for the given app identifier.
-pub fn socket_path(identifier: &str) -> std::path::PathBuf {
-    socket_dir().join(format!("tauri-pilot-{identifier}.sock"))
+/// Android uses an abstract address; other Unix platforms use a filesystem path.
+pub fn socket_address(identifier: &str) -> std::io::Result<SocketAddr> {
+    let name = format!("tauri-pilot-{identifier}.sock");
+    #[cfg(target_os = "android")]
+    {
+        use std::os::android::net::SocketAddrExt;
+        SocketAddr::from_abstract_name(name)
+    }
+    #[cfg(not(target_os = "android"))]
+    SocketAddr::from_pathname(socket_dir().join(name))
 }
 
 /// Bind the socket using the **std** (sync) listener so this can be called
@@ -90,10 +101,20 @@ pub fn socket_path(identifier: &str) -> std::path::PathBuf {
 ///
 /// Tries bind first; only removes stale files on `AddrInUse` after verifying
 /// no live server is listening.
-/// Returns a std listener and a [`SocketGuard`] that cleans up on drop.
+/// Abstract sockets are released by the kernel; only filesystem sockets need a guard.
 pub fn bind(
-    socket_path: &std::path::Path,
-) -> Result<(std::os::unix::net::UnixListener, SocketGuard), Error> {
+    address: &SocketAddr,
+) -> Result<(std::os::unix::net::UnixListener, Option<SocketGuard>), Error> {
+    let Some(socket_path) = address.as_pathname() else {
+        let listener = std::os::unix::net::UnixListener::bind_addr(address)?;
+        listener.set_nonblocking(true)?;
+        tracing::info!(
+            version = env!("CARGO_PKG_VERSION"),
+            ?address,
+            "tauri-pilot socket listening"
+        );
+        return Ok((listener, None));
+    };
     // SAFETY: umask is always safe to call; we restore the old mask immediately.
     let old_mask = unsafe { libc::umask(0o177) };
     let first_bind = std::os::unix::net::UnixListener::bind(socket_path);
@@ -142,10 +163,10 @@ pub fn bind(
 
     Ok((
         listener,
-        SocketGuard {
+        Some(SocketGuard {
             path: socket_path.to_path_buf(),
             inode,
-        },
+        }),
     ))
 }
 
@@ -153,7 +174,7 @@ pub fn bind(
 /// The `_guard` is held for its `Drop` cleanup.
 pub async fn run(
     listener: std::os::unix::net::UnixListener,
-    _guard: SocketGuard,
+    _guard: Option<SocketGuard>,
     engine: EvalEngine,
     eval_fn: Option<EvalFn>,
     list_fn: Option<ListWindowsFn>,
@@ -191,16 +212,21 @@ async fn accept_loop(
             }
         };
 
-        // Verify the connecting process belongs to the same user.
+        // Authenticate kernel-provided credentials before reading requests.
         match stream.peer_cred() {
             Ok(cred) => {
                 // SAFETY: getuid() is always safe to call; it has no preconditions.
                 let my_uid = unsafe { libc::getuid() };
-                if cred.uid() != my_uid {
+                #[cfg(target_os = "android")]
+                let authorized = android_peer_allowed(cred.uid(), cred.gid(), my_uid);
+                #[cfg(not(target_os = "android"))]
+                let authorized = cred.uid() == my_uid;
+                if !authorized {
                     tracing::warn!(
                         peer_uid = cred.uid(),
-                        expected_uid = my_uid,
-                        "rejected connection from different user"
+                        peer_gid = cred.gid(),
+                        app_uid = my_uid,
+                        "rejected unauthorized Unix peer"
                     );
                     continue;
                 }
@@ -228,7 +254,14 @@ async fn accept_loop(
     }
 }
 
-#[cfg(test)]
+// Android reserves UID 0 for root and 2000 for the ADB shell. This matches
+// Chromium Android DevTools: same UID/GID, and only the app, shell or root.
+#[cfg(target_os = "android")]
+fn android_peer_allowed(uid: u32, gid: u32, app_uid: u32) -> bool {
+    uid == gid && (uid == app_uid || uid == 0 || uid == 2000)
+}
+
+#[cfg(all(test, not(target_os = "android")))]
 mod tests {
     use super::*;
     use crate::protocol::Response;
@@ -249,7 +282,8 @@ mod tests {
     }
 
     async fn start_test_server(path: &Path) -> tokio::task::JoinHandle<()> {
-        let (listener, guard) = bind(path).expect("bind test socket");
+        let address = SocketAddr::from_pathname(path).expect("test socket address");
+        let (listener, guard) = bind(&address).expect("bind test socket");
         let engine = EvalEngine::new();
         let handle = tokio::spawn(async move {
             run(listener, guard, engine, None, None, None, Recorder::new()).await;
@@ -379,7 +413,8 @@ mod tests {
     async fn test_bind_socket_has_mode_0o600() {
         use std::os::unix::fs::PermissionsExt;
         let socket = unique_socket_path();
-        let (listener, guard) = bind(&socket).expect("bind test socket");
+        let address = SocketAddr::from_pathname(&socket).expect("test socket address");
+        let (listener, guard) = bind(&address).expect("bind test socket");
         let meta = std::fs::metadata(&socket).expect("socket metadata");
         let mode = meta.permissions().mode() & 0o777;
         assert_eq!(
--- a/crates/tauri-plugin-pilot/src/server/unix.rs
+++ b/crates/tauri-plugin-pilot/src/server/unix.rs
@@ -84,37 +84,69 @@ fn socket_dir() -> std::path::PathBuf {
     socket_dir_from(std::env::var_os("XDG_RUNTIME_DIR"))
 }
 
-/// Android uses an abstract address; other Unix platforms use a filesystem path.
+/// Android uses a random per-instance abstract name; other Unix platforms use a path.
+///
+/// # Errors
+/// Returns an error if OS randomness cannot be read or the address is invalid or too long.
 pub fn socket_address(identifier: &str) -> std::io::Result<SocketAddr> {
-    let name = format!("tauri-pilot-{identifier}.sock");
     #[cfg(target_os = "android")]
     {
+        use std::io::Read;
         use std::os::android::net::SocketAddrExt;
+
+        let mut random = [0_u8; 16];
+        std::fs::File::open("/dev/urandom")?.read_exact(&mut random)?;
+        let name = format!(
+            "tauri-pilot-{identifier}-{:032x}.sock",
+            u128::from_ne_bytes(random)
+        );
         SocketAddr::from_abstract_name(name)
     }
     #[cfg(not(target_os = "android"))]
-    SocketAddr::from_pathname(socket_dir().join(name))
+    SocketAddr::from_pathname(socket_dir().join(format!("tauri-pilot-{identifier}.sock")))
 }
 
 /// Bind the socket using the **std** (sync) listener so this can be called
 /// outside a tokio runtime (e.g. from Tauri plugin `setup`).
 ///
-/// Tries bind first; only removes stale files on `AddrInUse` after verifying
-/// no live server is listening.
-/// Abstract sockets are released by the kernel; only filesystem sockets need a guard.
+/// Returns the listener and a cleanup guard for pathname sockets. Abstract sockets
+/// have no guard because the kernel releases their names when the listener closes.
+///
+/// # Errors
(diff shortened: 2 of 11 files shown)


crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -76,24 +79,42 @@ fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
...
 fn socket_dir() -> std::path::PathBuf {
     socket_dir_from(std::env::var_os("XDG_RUNTIME_DIR"))
 }
 
-/// Build the full socket path for the given app identifier.
-pub fn socket_path(identifier: &str) -> std::path::PathBuf {
-    socket_dir().join(format!("tauri-pilot-{identifier}.sock"))
+/// Android uses an abstract address; other Unix platforms use a filesystem path.
+pub fn socket_address(identifier: &str) -> std::io::Result<SocketAddr> {
+    let name = format!("tauri-pilot-{identifier}.sock");
```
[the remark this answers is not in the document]
Darkatse: Thanks for pointing this out. You're right that the current peer checks do not authenticate the server, so they do not prevent another app from claiming the socket name before startup.
This integration is intended for debug builds on development devices, using ADB. It follows a similar model to [Chromium's Android debugging sockets](https://chromium.googlesource.com/chromium/src/+/main/chrome/browser/android/devtools_server.cc), which also use predictable abstract names. I think that is a reasonable starting point for this support, while recognizing the limitation you've described.
Per-instance capabilities and server authentication would also need a trusted way to discover the endpoint. I'd prefer to discuss that as a separate hardening change if protecting against hostile apps on the development device is part of the intended threat model. Would that scope be reasonable for this PR?

mpiton:
Hello, thank you for your contribution. I’m going to give you some feedback on a few points that I think need correcting.

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -76,24 +79,42 @@ fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
...
+    let Some(socket_path) = address.as_pathname() else {
+        let listener = std::os::unix::net::UnixListener::bind_addr(address)?;
+        listener.set_nonblocking(true)?;
+        tracing::info!(
+            version = env!("CARGO_PKG_VERSION"),
+            ?address,
+            "tauri-pilot socket listening"
+        );
+        return Ok((listener, None));
+    };
```
mpiton: > [!WARNING]
> **🟠 High · 🔒 Security** — Abstract socket name is first-come-first-served, so any app can squat it

The abstract namespace has no permission bits and is global to the device's network namespace, so `tauri-pilot-<identifier>.sock` is a well-known name any installed app can bind first. If it wins the race, `bind_addr` fails with AddrInUse, the `?` in lib.rs:79 propagates out of the Tauri setup hook and the app does not start. If it also answers the ADB-forwarded connection it receives the developer's `fill`/`eval`/`storage.set` traffic and returns forged snapshots, which is a prompt-injection channel in `mcp` mode. The filesystem path keeps four layers here (private XDG dir, umask 0o177, chmod 0o600, peer_cred); the abstract address keeps only peer_cred, and peer_cred does not run until after the attacker has already taken the name.

**Fix:** Add a random per-instance suffix to the abstract name so the name cannot be predicted or pre-bound. `lib.rs:81` already logs `?address` at info level, so the developer reads it back with `adb logcat` and passes it to `adb forward`. Keep that log at info. Please also confirm on a device whether an untrusted_app can bind an arbitrary abstract name under current SELinux policy; if it cannot, this drops to medium.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the abstract bind branch is added by this diff
- **References:** CWE-283

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.8</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -228,7 +254,14 @@ async fn accept_loop(
     }
 }
 
-#[cfg(test)]
+// Android reserves UID 0 for root and 2000 for the ADB shell. This matches
+// Chromium Android DevTools: same UID/GID, and only the app, shell or root.
+#[cfg(target_os = "android")]
+fn android_peer_allowed(uid: u32, gid: u32, app_uid: u32) -> bool {
+    uid == gid && (uid == app_uid || uid == 0 || uid == 2000)
+}
```
mpiton: > [!WARNING]
> **🟠 High · 🧪 Tests** — android_peer_allowed is compiled, linted and tested by no build configuration

`android_peer_allowed` is gated on `cfg(target_os = "android")` while the test module below is gated on `cfg(all(test, not(target_os = "android")))`. The two are mutually exclusive by construction: on the CI matrix (ubuntu/macos/windows) the function is not even type-checked, and under `cargo test --target aarch64-linux-android` the test module does not exist. There is no configuration, not even on a real device, in which a test can call it. Since `set_permissions(0o600)` is skipped on the abstract path, this predicate is the only thing standing between another installed app and `webview.eval()`, and a typo such as `uid == 200` would compile everywhere CI runs.

**Fix:** Compile it for tests as well and cover it. `#[cfg(any(target_os = "android", test))]` keeps clippy quiet without `#[allow(dead_code)]` precisely because the test calls it. Add a table-driven test over (app,app)->true, (0,0)->true, (2000,2000)->true, (app,0)->false, (2000,0)->false, (other,other)->false. Verified locally: clippy --all-targets -D warnings clean, fmt clean, suite green, Android target still compiles; removing the `uid == gid` clause makes the test fail on the (app,0) row.

```suggestion
#[cfg(any(target_os = "android", test))]
fn android_peer_allowed(uid: u32, gid: u32, app_uid: u32) -> bool {
    uid == gid && (uid == app_uid || uid == 0 || uid == 2000)
}
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** both the function and the narrowed test cfg are added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.98</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -76,8 +76,9 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                 // listener to tokio once it is already on the runtime.
                 #[cfg(unix)]
                 {
-                    let (listener, guard) = server::bind(&socket_path).map_err(|e| {
-                        tracing::error!(path = %socket_path.display(), "failed to bind socket: {e}");
+                    let address = server::socket_address(&identifier)?;
```
> **🟡 Medium · 🐛 Correctness** — socket_address error is propagated bare, with no mention of tauri-pilot or the identifier

`socket_address` is newly fallible and the `?` here drops all context, while the very next line logs `?address` on the bind failure. `sanitize_identifier` does not cap length, so an identifier of 91 bytes or more fails `from_abstract_name` and the app owner sees only `abstract socket name must be shorter than SUN_LEN` in the Tauri setup error, with no mention of tauri-pilot, of the identifier, or of Android.

**Fix:** Log before propagating: `.map_err(|e| { tracing::error!(identifier, "failed to build socket address: {e}"); e })?`, matching the treatment `bind` already gets on the next line.

```suggestion
                    let address = server::socket_address(&identifier).map_err(|e| {
                        tracing::error!(identifier, "failed to build socket address: {e}");
                        e
                    })?;
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the fallible call is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -76,24 +79,42 @@ fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
...
 /// Tries bind first; only removes stale files on `AddrInUse` after verifying
 /// no live server is listening.
-/// Returns a std listener and a [`SocketGuard`] that cleans up on drop.
+/// Abstract sockets are released by the kernel; only filesystem sockets need a guard.
 pub fn bind(
-    socket_path: &std::path::Path,
-) -> Result<(std::os::unix::net::UnixListener, SocketGuard), Error> {
+    address: &SocketAddr,
+) -> Result<(std::os::unix::net::UnixListener, Option<SocketGuard>), Error> {
+    let Some(socket_path) = address.as_pathname() else {
```
> **🟡 Medium · 🧱 Maintainability** — bind() picks its platform branch by sniffing as_pathname() instead of stating it

The implicit contract is "no pathname means Android", which is nowhere written down. Every platform compiles roughly fifty lines it can never reach, and the abstract branch silently skips the umask, the stale-socket probe, the chmod 0o600 and the inode guard without saying that it does. `as_pathname()` also returns None for an unnamed address (`SocketAddr::from_pathname("")` is Ok and unnamed), so an unnamed address would fall into the abstract branch and get a kernel autobind; `bind` is pub, so that path is reachable from outside `socket_address`.

**Fix:** Make the switch explicit, either two cfg-gated `bind` functions or a delegation to `bind_abstract` / `bind_pathname`. At minimum, comment the abstract branch listing the hardening that does not apply to it, and reject `address.is_unnamed()` explicitly.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the branch and the sniffing are added by this diff

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -228,7 +254,14 @@ async fn accept_loop(
...
 
-#[cfg(test)]
+// Android reserves UID 0 for root and 2000 for the ADB shell. This matches
+// Chromium Android DevTools: same UID/GID, and only the app, shell or root.
+#[cfg(target_os = "android")]
+fn android_peer_allowed(uid: u32, gid: u32, app_uid: u32) -> bool {
+    uid == gid && (uid == app_uid || uid == 0 || uid == 2000)
+}
+
+#[cfg(all(test, not(target_os = "android")))]
```
mpiton, line 303: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The abstract branch of bind() has no test although it runs on ordinary Linux

The abstract branch is entirely disjoint from the pathname branch: its own `bind_addr`, its own `set_nonblocking`, its own `None` return. It is the branch Android actually executes and nothing exercises it. Abstract sockets are not Android-specific, they exist on plain Linux through `std::os::linux::net::SocketAddrExt`, so this is testable on ubuntu-latest today with no device.

**Fix:** Add a test that binds an abstract address on Linux, asserts the guard is `None` and asserts the listener is non-blocking via `fcntl(F_GETFL)`. Assert the flag rather than asserting `accept()` returns WouldBlock: the WouldBlock form is correct but hangs instead of failing when `set_nonblocking` is removed.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the branch is added by this diff and the test module is simultaneously narrowed away from it

</details>

docs/src/content/docs/guides/plugin-setup.md
```diff
@@ -109,3 +109,24 @@ cargo update -p tauri-plugin-pilot
 # then rebuild
 cargo tauri dev
 ```
+
+## Android via ADB
+
+On macOS or Linux, follow the setup above with `default-features = false` (Android has no native `press`) and register the plugin in the mobile `run()` entry point in `src-tauri/src/lib.rs`. Start the app with `cargo tauri android dev`.
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — The host-side Linux/macOS-only requirement is buried in a subordinate clause

"On macOS or Linux" reads as a statement about the phone, not about the developer's machine, and the README bullet added by this PR says nothing at all. A Windows host cannot use this feature: `crates/tauri-pilot-cli/src/client/mod.rs:22-31` routes `#[cfg(windows)]` to named pipes only, so `--socket` with a forwarded Unix socket has no code path. A Windows user follows the whole guide before finding out.

**Fix:** State it as its own sentence or a `:::caution` under the section heading: the device may be anything, but the host running the CLI must be Linux or macOS because the Windows client only speaks named pipes. Add a matching bullet under README `## Known limitations`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the section is added by this diff

</details>

docs/src/content/docs/guides/plugin-setup.md
```diff
@@ -109,3 +109,24 @@ cargo update -p tauri-plugin-pilot
...
+
+Connect a device or emulator through ADB, then replace `com.example.app` below with the identifier from `tauri.conf.json`:
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — Android section has no prerequisites, no adb -s guidance and no list of what does not work

The reader is told to run `cargo tauri android dev` and `adb` with no mention of `cargo tauri android init`, the SDK, ANDROID_HOME/NDK_HOME, adb on PATH or USB debugging. With more than one device attached every adb command here fails without `-s <serial>`. And nothing says which commands are unusable: `press` answers `-32601 "press disabled (compile tauri-plugin-pilot with the press feature)"` (handler.rs:215-221), which sends the user to enable a feature that will not help on Android; `screenshot_native` is PERMISSION_DENIED off macOS by design; `windows.list` works but will only ever return one window (lib.rs:199-220). `screenshot`, `record`/`replay`, `run`, `watch`, `diff` and MCP are untested on Android.

**Fix:** Add a prerequisites list, an `adb -s <serial>` note, and a short "not available on Android" list. Add one troubleshooting line for a forward that succeeds against nothing: `adb shell cat /proc/net/unix | grep tauri-pilot`, plus the reminder that the plugin only exists in debug builds.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the section is added by this diff

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -228,7 +254,14 @@ async fn accept_loop(
     }
 }
 
-#[cfg(test)]
+// Android reserves UID 0 for root and 2000 for the ADB shell. This matches
+// Chromium Android DevTools: same UID/GID, and only the app, shell or root.
+#[cfg(target_os = "android")]
+fn android_peer_allowed(uid: u32, gid: u32, app_uid: u32) -> bool {
+    uid == gid && (uid == app_uid || uid == 0 || uid == 2000)
```
mpiton, line 300: > [!NOTE]
> **🔵 Low · 🐛 Correctness** — uid == gid is undocumented, adds nothing, and rejects sharedUserId peers

All three allowed uids (the app, 0, 2000) already satisfy uid == gid on a normal device, so the clause rejects nothing an attacker would send. What it does reject is a legitimate peer whose primary gid differs from its uid, which is what `sharedUserId` produces, and the rejection is silent apart from a warn line. The comment above explains the uid set but never this conjunct.

**Fix:** Either document what the clause is defending against, with the case that motivates it, or drop it and keep the uid set alone.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the predicate is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.85</sub>

docs/src/content/docs/guides/plugin-setup.md
```diff
@@ -109,3 +109,24 @@ cargo update -p tauri-plugin-pilot
...
+
+```sh
+pilot_dir=$(mktemp -d /tmp/tauri-pilot.XXXXXX)
+adb forward "localfilesystem:$pilot_dir/pilot.sock" \
+  localabstract:tauri-pilot-com.example.app.sock
+tauri-pilot --socket "$pilot_dir/pilot.sock" snapshot
+```
```
mpiton, line 154: > [!NOTE]
> **🔵 Low · 🎯 Product** — Android needs --socket on every command where desktop needs no flag at all

Desktop is zero-flag: the CLI auto-discovers the socket. The Android flow asks for `mktemp -d`, a hand-substituted identifier, `--socket` on every single command and three teardown commands. For a tool whose stated audience is an agent, that is a real step down, and it costs nothing to avoid.

**Fix:** Two zero-code improvements: export `TAURI_PILOT_SOCKET` once instead of repeating `--socket` (the clap env var already exists, cli.rs:12), and forward into `$XDG_RUNTIME_DIR/tauri-pilot-<identifier>.sock` so CLI auto-discovery finds it, since the uid filter passes and 0700 preserves the privacy `mktemp -d` was buying. Keep `mktemp -d` plus `--socket` for macOS hosts, which have no XDG_RUNTIME_DIR. Do not forward to a bare /tmp name: adb creates the host socket with its own umask, not the plugin's 0o177, which is the regression issue #31 closed.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the workflow is introduced by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.85</sub>

docs/src/content/docs/getting-started.md
```diff
@@ -6,6 +6,7 @@ description: Install tauri-pilot and start testing your Tauri v2 app interactive
 ## Requirements
 
 - Linux (WebKitGTK) or macOS (WebKit) — Windows planned
+- Android via [ADB](/tauri-pilot/guides/plugin-setup/#android-via-adb)
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — The line above the new Android bullet still says Windows is planned

Line 8 reads "Linux (WebKitGTK) or macOS (WebKit) — Windows planned", but Windows shipped in 0.5.0 (#44) and the README lists it as supported. The new bullet lands directly under a sentence that contradicts the rest of the docs.

**Fix:** Fix line 8 to list Windows as supported while editing this block.

```suggestion
- Linux (WebKitGTK), macOS (WebKit), or Windows (WebView2)
- Android via [ADB](/tauri-pilot/guides/plugin-setup/#android-via-adb)
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR edits the Requirements block that carries the stale line

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 1.0</sub>

docs/src/content/docs/guides/plugin-setup.md
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — default-features = false is presented as required when it is runtime advice

The Android build does not need it. Verified with a real NDK link (28.2.13676358) using default features: the test binary and, more to the point, a cdylib depending on the plugin both link, and readelf shows NEEDED = libdl/libm/libc with zero xkb references. enigo pulls xkbcommon into the Android dependency graph, but its Android path references nothing from it, so rustc drops the crate at link time and the `#[link(name = "xkbcommon")]` directive goes with it. Meanwhile a bare `default-features = false` also disables `press` on the same app's desktop builds, which is a regression the reader is not warned about.

**Fix:** Reword as recommended rather than required, with the real reason: `press` is inoperative on Android and enigo only adds APK weight. Show the target-scoped form so desktop builds keep `press`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the claim is added by this diff

</details>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -39,6 +39,7 @@ pub(crate) const BRIDGE_JS: &str = concat!(
 /// On non-Unix, non-Windows platforms or in release builds, returns a no-op plugin.
 /// In debug builds on Unix, injects the JS bridge, stores an `EvalEngine`,
 /// and starts a Unix socket server at `$XDG_RUNTIME_DIR/tauri-pilot-{identifier}.sock` (falls back to `/tmp` if unavailable).
+/// Android uses the abstract Unix socket `tauri-pilot-{identifier}.sock`, reachable via ADB forwarding.
```
> **🔵 Low · 📄 Docs** — init() doc says Unix uses XDG_RUNTIME_DIR and then that Android is abstract

Line 41 states that debug builds on Unix start a server at `$XDG_RUNTIME_DIR/tauri-pilot-{identifier}.sock`, and the line added here states that Android uses an abstract socket. Android is Unix, so the block contradicts itself and a reader cannot tell which sentence governs.

**Fix:** Make line 41 say "on Unix other than Android", or restructure the block as one sentence per platform.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the contradiction is created by the line this diff adds

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -76,24 +79,42 @@ fn socket_dir_from(xdg: Option<std::ffi::OsString>) -> std::path::PathBuf {
...
+    SocketAddr::from_pathname(socket_dir().join(name))
 }
 
 /// Bind the socket using the **std** (sync) listener so this can be called
 /// outside a tokio runtime (e.g. from Tauri plugin `setup`).
 ///
 /// Tries bind first; only removes stale files on `AddrInUse` after verifying
 /// no live server is listening.
-/// Returns a std listener and a [`SocketGuard`] that cleans up on drop.
+/// Abstract sockets are released by the kernel; only filesystem sockets need a guard.
```
> **🔵 Low · 📄 Docs** — bind() doc was replaced instead of amended, so nothing documents the Option or the errors

"Returns a std listener and a SocketGuard that cleans up on drop" was swapped for a sentence about the kernel, so the return value is no longer described at all and nothing says when the Option is None. The surviving line above, "only removes stale files on AddrInUse after verifying no live server is listening", is now true of one branch out of two. `socket_address` is likewise a new public fallible function with no `# Errors` section.

**Fix:** Keep both sentences: describe the return, say the guard is None for abstract addresses because the kernel releases them, and scope the stale-file sentence to the pathname branch. Add an `# Errors` section to `socket_address`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the doc block is rewritten by this diff

</details>

mpiton (changes requested):
## 🧨 mpiton-review — PR #145

> [!CAUTION]
> 🚫 **Changes requested** — 3 blocking issues (0 critical, 3 high) out of 17 findings.

**Actionable comments posted: 13**

|:---|---:|:---|
| 🟠 High | 3 | must fix before merge |
| 🟡 Medium | 7 | fix now or file a follow-up |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/server/unix.rs` | 0 | 2 | 2 | 2 |
| `docs/src/content/docs/guides/plugin-setup.md` | 0 | 0 | 2 | 2 |
| `crates/tauri-plugin-pilot/src/lib.rs` | 0 | 0 | 1 | 1 |
| `docs/src/content/docs/getting-started.md` | 0 | 0 | 0 | 1 |
| `.github/workflows/ci.yml` | 0 | 1 | 0 | 0 |
| `SECURITY.md` | 0 | 0 | 1 | 0 |
| `docs/src/content/docs/reference/cli.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟠 **Abstract socket name is first-come-first-served, so any app can squat it** — `crates/tauri-plugin-pilot/src/server/unix.rs:117`
2. 🟠 **android_peer_allowed is compiled, linted and tested by no build configuration** — `crates/tauri-plugin-pilot/src/server/unix.rs:262`
3. 🟡 **socket_address error is propagated bare, with no mention of tauri-pilot or the identifier** — `crates/tauri-plugin-pilot/src/lib.rs:79`
4. 🟡 **bind() picks its platform branch by sniffing as_pathname() instead of stating it** — `crates/tauri-plugin-pilot/src/server/unix.rs:108`
5. 🟡 **The abstract branch of bind() has no test although it runs on ordinary Linux** — `crates/tauri-plugin-pilot/src/server/unix.rs:264`

<details>
<summary>🧭 Findings outside the diff or over the inline cap (4)</summary>

- 🟠 **Nothing in CI compiles the Android target, and release.yml publishes through the same gap** — `.github/workflows/ci.yml:33` — The matrix is [ubuntu-latest, macos-latest, windows-latest] and no step passes `--target`, so every line this PR adds under `cfg(target_os = "android")` can rot without a single job going red. This PR is green because CI does not look at the code it adds. `release.yml:71-78` replays the same fmt/clippy/test on the same base before publishing to crates.io, so the publication gate inherits the blind
  - *Fix:* Add a Linux-only step after the cargo cache: `rustup target add aarch64-linux-android` then `cargo check -p tauri-plugin-pilot --target aarch64-linux-android --no-default-features` and the matching `cargo clippy ... -- -D warnings`. No NDK is needed because cargo check does not link. Measured locall
  - *Link to this PR:* this PR adds a platform branch that no existing job compiles; the guard has to land with it
- 🟡 **CHANGELOG.md is missing from the PR** — `CHANGELOG.md:8` — release.yml:61-69 fails the tag when CHANGELOG.md has no section for the version being released, so this has to land before the next release rather than after it. The `[Unreleased]` section on main is currently empty, so nothing masks the omission and the release would ship an undocumented platform.
  - *Fix:* Add an entry under `## [Unreleased]` / `### Added`: Android support via ADB, abstract socket `tauri-pilot-{identifier}.sock` reached through `adb forward localfilesystem:... localabstract:...`, peer authorization limited to uid==gid and uid in {app, 0, 2000}, referencing #145. No version bump in the
  - *Link to this PR:* the changelog entry describing this change has to ship with it; release.yml gates the tag on it
- 🟡 **SECURITY.md still claims the socket is a /tmp file with user-level permissions** — `SECURITY.md:5` — The published policy says "The Unix socket is local-only (`/tmp/`) with user-level permissions". On Android there is no filesystem entry and no permission bits at all; the only control is the peer_cred check. Anyone consulting the security policy to reason about the Android transport gets a description of a protection that does not exist there.
  - *Fix:* Extend the Scope section with the Android model: abstract namespace, no filesystem permissions, authorization limited to SO_PEERCRED (app uid, root, adb shell), and the fact that it depends on the device's SELinux policy.
  - *Link to this PR:* this PR introduces the Android transport that the existing policy statement no longer describes
- 🔵 **cli.md claims press works on all supported platforms, which Android makes false** — `docs/src/content/docs/reference/cli.md:386` — "Events are isTrusted=true and reach DOM listeners and Tauri accelerators on all supported platforms". Once Android is a supported platform that sentence is wrong: `press` is unavailable there and returns -32601.
  - *Fix:* Scope the sentence to desktop platforms, or add the Android exception next to it.
  - *Link to this PR:* this PR adds Android to the supported platform list that the sentence quantifies over

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `cca9e7650a` |
| Base | `main` |
| Files changed | 6 (+80/-21) |
| Languages | rust |
| Reviewers | documentation, maintainability, product-logic, quality, security, tests |

</details>

Darkatse:
Thanks for the review. I've addressed the feedback and verified the changes on an Android device and emulator; CI is also passing.

Darkatse:
Updated in c758c6b: Android socket names now use at most 16 identifier characters and a 16-digit OS-random hexadecimal suffix. Added a long-identifier regression test and clarified cleanup for the XDG forwarding alternative. Workspace tests, Clippy, Android check/Clippy, and the docs build pass locally.

mpiton:
@coderabbit resolve

mpiton (approved):

PR #147, opened 2026-09-10
fix(bridge): curate the style properties screenshot copies onto the clone

### LINKED ISSUE #146: screenshot pins the webview for minutes on macOS 26 WebKit: per-property style clone is quadratic
## Summary

`screenshot` (the bridge-side html-to-image path) pins the WebContent process at ~100 % CPU for roughly ten minutes on a modest page (648 DOM nodes, 1200×800 viewport) under macOS 26 WebKit. The RPC returns `Eval error: eval timed out after 30s`, but the clone keeps running, so every bridge-backed method afterwards (`eval`, `logs`, `ipc`, `state`, …) fails with `eval timed out after 10s` until it finishes. `ping`, `windows` and `screenshot_native` keep answering. The page recovers by itself; nothing reloads.

## Environment

- tauri-plugin-pilot 0.7.2 / tauri-pilot-cli 0.7.2 (0.7.3 does not touch this path)
- html-to-image 1.11.13 (vendored `js/vendor/html-to-image.iife.js`)
- macOS 26 (Darwin 25.5.0), WebKit 21624.2.5.11.8, Tauri 2
- Page: React 19 + Tailwind v4, 648 elements, 103 stylesheets, 2346 rules, 1 `@font-face`, no images, no iframes, no portals open

## Where the time goes

`sample` on the WebContent process while wedged (2039 samples, main thread), collapsed:

```
jsCSSStyleDeclarationPrototypeFunction_setProperty
  PropertySetCSSStyleProperties::setProperty
    InlineCSSStyleProperties::didMutate
      StyleProperties::asTextAtom
        StyleProperties::asTextInternal
          serializeShorthandValue → ShorthandSerializer::commonSerializationChecks   (1195 of 2039 samples at top of stack)
```

So on this WebKit build, every `CSSStyleDeclaration.setProperty` on an element's inline style eagerly re-serializes the whole style attribute, and that serialization walks every shorthand for every longhand. html-to-image's `cloneCSSStyle` falls into its per-property branch on WebKit (`getComputedStyle().cssText` is empty there) and calls `setProperty` once per property name from `getComputedStyle(document.documentElement)` — on this page that is 2180 names, 1761 of them custom properties from Tailwind v4 and design-token variables on `:root`. Measured: one element × 2180 `setProperty` calls = 1.2 s; the full `toSvg` clone of the 648-node page = 370 s and a 40 MB SVG string. The `toPng` that follows never resolved at all (no result, no error) after the second clone. That makes the copy quadratic in the number of properties on every single element, and the whole clone runs synchronously on the JS thread — the 30 s Rust-side timeout cannot interrupt it.

## Reproduction

1. Any Tauri 2 app on macOS 26 with a Tailwind v4 (or otherwise variable-heavy) page of a few hundred nodes.
2. `tauri-pilot screenshot /tmp/x.png`
3. Observe `eval timed out after 30s`, then `tauri-pilot eval "1+1"` → `eval timed out after 10s` for ~10 min while `ping` succeeds and the WebContent process sits at 100 % CPU.

## Suggested fixes

- Pass `includeStyleProperties` to html-to-image with a curated list, or diff each element's computed style against the tag's default style before copying, so the per-element `setProperty` count drops from ~1000+ to a few dozen. Either takes the clone from minutes to under a second on this page.
- Prefer native capture on macOS: `screenshot_native` already works (cgwindow backend, ~1 s, even while the bridge is wedged). The scaffolded `WKWebView takeSnapshot` backend would avoid TCC entirely.
- Two small CLI/plugin gaps that make the native path hard to find:
  - `screenshot_native` with an unknown id returns `available_windows` in `error.data`, but the CLI prints only the message, so there is no way to discover a window id from the CLI. Printing that list (or adding it to `windows --json`) would fix it.
  - With `tcc_denied: true` the result reports `scale_factor: 2.87` for a 1× display: it looks like image width ÷ window logical width, which is wrong when the image is the whole screen.

mpiton:
Closes #146.

## What was wrong

`screenshot()` calls `htmlToImage.toPng()` without `includeStyleProperties`. In the vendored bundle, `cloneCSSStyle` does:

```js
i.cssText ? (n.cssText = i.cssText, …) : R(r).forEach(o => n.setProperty(o, …))
```

`getComputedStyle(el).cssText` is empty on WebKit *and* Blink (checked in headless Chromium: `cssText.length === 0`), so the per-property branch is always the one taken. `R()` returns every name from `getComputedStyle(document.documentElement)`, custom properties included — ~2200 names on a Tailwind v4 page, ~1760 of them `--*` — and copies all of them onto every cloned element.

macOS 26 WebKit re-serializes the entire style attribute on each `setProperty` (`InlineCSSStyleProperties::didMutate` -> `StyleProperties::asTextAtom` -> `ShorthandSerializer::commonSerializationChecks`, 1195 of 2039 samples in the profile on the issue), which makes the copy quadratic in the property count. The clone loop is synchronous, so the 30 s eval timeout cannot interrupt it and every later bridge call (`eval`, `logs`, `ipc`, `state`) times out until it finishes — the ~10 minute wedge described in the issue.

## The change

Pass `includeStyleProperties` with a curated list of the properties a page actually paints: box and layout, flex/grid, background/border, paint effects, text (including `-webkit-text-fill-color`, `-webkit-line-clamp`, `-webkit-background-clip`), `content` for `::before`/`::after`, replaced content, SVG and the `d` property the library special-cases.

Custom properties are dropped on purpose — computed values arrive with their `var()` already resolved, so nothing on the clone reads them. The tradeoff is real and worth stating: **a painted property missing from this list is missing from the capture**, so the list is the thing to extend when a screenshot comes back wrong.

## Verification

Repro harness: headless Chromium page loading the *real* vendor + bridge.js concatenation (same as the plugin's `concat!`), 648 nodes and 1761 custom properties on `:root`, with `CSSStyleDeclaration.prototype.setProperty` counted.

| | setProperty/element | total | ms (Chromium) |
|---|---|---|---|
| before | 2239 | 2 919 656 | 13 568 |
| after | 158 | 206 032 | 486 |

Isolated: `perElement = 478 standard longhands + number of custom properties`. Chromium is not quadratic per call, so the 13.5 s here is what shows up as ~370 s on the reported macOS build.

Fidelity: a page with gradients, gradient-clipped text, box shadows, flex, grid, list markers, `::after` content, inline SVG and line-clamp renders **pixel-identical** before and after (same PNG md5).

`node --test crates/tauri-plugin-pilot/js/*.test.mjs` 31/31 (the new test is red without the fix), `cargo test --workspace` 315/315.

## Not in this PR

- Estimated macOS cost after the fix is ~4 s on that page, not under a second. Going lower means one `cssText =` assignment instead of N `setProperty` calls, which the library exposes no option for — it would take patching the vendored bundle.
- The two CLI gaps reported alongside this one (`screenshot_native` not printing `available_windows` from `error.data`, and `scale_factor: 2.87` under `tcc_denied`) are separate bugs, untouched here. Filed as #149.
- `cargo clippy --all-targets` already fails on `main` in the CLI crate's MCP module, line 679 (`manual_is_variant_and`, new lint). Pre-existing, fixed separately in #148.

---
## Summary by cubic
Fixes #146 by replacing html-to-image's full computed-style copy during `screenshot()` with a curated 187-property allowlist. This avoids the quadratic WebKit work that can wedge style-variable-heavy pages for minutes, while CSS painted through properties outside the list is omitted from the image.

**Bug Fixes**
- Preserves masked fields, hidden subtrees, counters including `counter-set`, SVG content, masks, and form controls.
- Applies the allowlist to selector captures and verifies that the vendored html-to-image bundle still reads it.
- `screenshot_native` remains the pixel-exact fallback on macOS; it returns `PERMISSION_DENIED` on Linux and Windows.

<sup>Written for commit f8678029115d81844c5051cabf228aafdffe49f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/147?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved screenshot performance, particularly on pages with many custom CSS properties.
  - Screenshot rendering now uses a curated set of visual styles, including counter styles, for more efficient and consistent captures.

- **Documentation**
  - Documented that unsupported CSS properties may be omitted and could cause PNG output to differ from the app.
  - Clarified that native screenshots are macOS-only, bypass the CSS-property allowlist, and return `PERMISSION_DENIED` on Linux and Windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `dangerous_mcp_tools_enabled`; it reads `.is_ok_and(...)` now, same
   behaviour.
 
+- `screenshot` no longer pins the webview for minutes on style-variable-heavy
+  pages. html-to-image copies computed styles onto the clone, and whenever
+  `getComputedStyle(el).cssText` is empty — WebKit and Blink both — it falls
+  back to one `setProperty` per name in `getComputedStyle(documentElement)`.
+  On a Tailwind v4 page that is ~2200 names, ~1760 of them custom properties,
+  applied to every cloned element; macOS WebKit re-serializes the whole style
+  attribute on each call, making the clone quadratic (~370 s and 100 % CPU on a
+  648-node page, with every later bridge call timing out until it finished).
+  The bridge now passes a curated `includeStyleProperties` list of the
+  properties a page actually paints, so the copy drops from ~2240 to 158
+  `setProperty` calls per element (28x faster end-to-end on a Chromium repro of
+  that page, pixel-identical output). Custom properties are dropped on purpose:
+  computed values already arrive with their `var()` resolved. [#146]
+
 ## [0.7.3] - 2026-08-30
 
 ### Changed
@@ -619,3 +633,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#142]: https://github.com/mpiton/tauri-pilot/pull/142
 [#143]: https://github.com/mpiton/tauri-pilot/pull/143
 [#145]: https://github.com/mpiton/tauri-pilot/pull/145
+[#146]: https://github.com/mpiton/tauri-pilot/issues/146
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,8 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   repro of that page at the original 156-name cut, pixel-identical output).
   Custom properties are dropped on purpose: computed values already arrive with
   their `var()` resolved. The list is an allowlist, though, so a standard
-  painted property that is not on it is missing from the capture as well —
-  `screenshot_native` remains the pixel-exact escape hatch. [#146]
+  painted property that is not on it is missing from the capture as well;
+  on macOS `screenshot_native` remains the pixel-exact escape hatch. [#146]
 
 ## [0.7.3] - 2026-08-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   attribute on each call, making the clone quadratic (~370 s and 100 % CPU on a
   648-node page, with every later bridge call timing out until it finished).
   The bridge now passes a curated `includeStyleProperties` list of the
-  properties a page actually paints, so the copy drops from ~2240 to 158
-  `setProperty` calls per element (28x faster end-to-end on a Chromium repro of
-  that page, pixel-identical output). Custom properties are dropped on purpose:
-  computed values already arrive with their `var()` resolved. [#146]
+  properties a page actually paints, so the copy drops from ~2200 to 187
+  `setProperty` calls per element (measured 28x faster end-to-end on a Chromium
+  repro of that page at the original 156-name cut, pixel-identical output).
+  Custom properties are dropped on purpose: computed values already arrive with
+  their `var()` resolved. The list is an allowlist, though, so a standard
+  painted property that is not on it is missing from the capture as well —
+  `screenshot_native` remains the pixel-exact escape hatch. [#146]
 
 ## [0.7.3] - 2026-08-30
 
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -1205,7 +1205,7 @@
     "max-height", "box-sizing", "aspect-ratio", "margin-top", "margin-right",
     "margin-bottom", "margin-left", "padding-top", "padding-right",
     "padding-bottom", "padding-left", "overflow-x", "overflow-y", "visibility",
-    "opacity", "vertical-align",
+    "opacity", "vertical-align", "content-visibility", "clip",
     // Flex + grid
     "flex-direction", "flex-wrap", "flex-grow", "flex-shrink", "flex-basis",
     "justify-content", "justify-items", "justify-self", "align-content",
@@ -1213,7 +1213,7 @@
     "grid-template-columns", "grid-template-rows", "grid-template-areas",
     "grid-auto-flow", "grid-auto-columns", "grid-auto-rows",
     "grid-column-start", "grid-column-end", "grid-row-start", "grid-row-end",
-    "column-count",
+    "column-count", "column-width",
     // Background + border
     "background-color", "background-image", "background-position",
     "background-size", "background-repeat", "background-clip",
@@ -1226,10 +1226,14 @@
     "border-bottom-right-radius", "border-bottom-left-radius",
     "border-collapse", "border-spacing", "table-layout", "outline-color",
     "outline-style", "outline-width", "outline-offset", "box-shadow",
+    "border-image-source", "border-image-slice", "border-image-width",
+    "border-image-outset", "border-image-repeat",
     // Paint effects
-    "filter", "backdrop-filter", "mix-blend-mode", "clip-path", "mask-image",
-    "transform", "transform-origin", "transform-style", "translate", "rotate",
-    "scale", "perspective",
+    "filter", "backdrop-filter", "mix-blend-mode", "background-blend-mode",
+    "isolation", "clip-path", "mask-image", "mask-size", "mask-position",
+    "mask-repeat", "mask-mode", "mask-composite", "transform",
+    "transform-origin", "transform-style", "translate", "rotate", "scale",
+    "perspective", "perspective-origin", "backface-visibility",
     // Text
     "color", "font-family", "font-size", "font-weight", "font-style",
     "font-variant", "font-stretch", "font-feature-settings",
@@ -1237,16 +1241,21 @@
     "text-align", "text-indent", "text-transform", "text-shadow",
     "text-overflow", "text-decoration-line", "text-decoration-color",
     "text-decoration-style", "text-decoration-thickness",
-    "text-underline-offset", "-webkit-text-fill-color", "-webkit-text-stroke",
+    "text-underline-offset", "-webkit-text-fill-color",
+    "-webkit-text-stroke-width", "-webkit-text-stroke-color",
+    "-webkit-text-security", "-webkit-font-smoothing",
     "-webkit-line-clamp", "-webkit-box-orient", "white-space", "word-break",
     "overflow-wrap", "hyphens", "direction", "unicode-bidi", "writing-mode",
     "text-orientation", "tab-size", "list-style-type", "list-style-position",
-    "list-style-image", "content",
+    "list-style-image", "content", "counter-reset", "counter-increment",
     // Replaced content + SVG
     "object-fit", "object-position", "image-rendering", "fill", "fill-opacity",
     "fill-rule", "stroke", "stroke-width", "stroke-opacity", "stroke-linecap",
     "stroke-linejoin", "stroke-dasharray", "stroke-dashoffset", "clip-rule",
-    "stop-color", "stop-opacity", "d",
+    "stop-color", "stop-opacity", "d", "text-anchor", "dominant-baseline",
+    "paint-order", "marker-start", "marker-mid", "marker-end",
+    // Form controls
+    "appearance", "-webkit-appearance", "accent-color",
   ];
 
   async function screenshot(options) {
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -1248,6 +1248,7 @@
     "overflow-wrap", "hyphens", "direction", "unicode-bidi", "writing-mode",
     "text-orientation", "tab-size", "list-style-type", "list-style-position",
     "list-style-image", "content", "counter-reset", "counter-increment",
+    "counter-set",
     // Replaced content + SVG
     "object-fit", "object-position", "image-rendering", "fill", "fill-opacity",
     "fill-rule", "stroke", "stroke-width", "stroke-opacity", "stroke-linecap",
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -1188,6 +1188,67 @@
     });
   }
 
+  // html-to-image copies computed styles onto the clone. When
+  // `getComputedStyle(el).cssText` is empty — WebKit and Blink both — it falls
+  // back to one `setProperty` per name in `getComputedStyle(documentElement)`,
+  // which on a Tailwind v4 page is ~2200 names (~1760 of them custom
+  // properties). WebKit re-serializes the whole style attribute on every
+  // `setProperty`, so that copy is quadratic: minutes of 100% CPU on a few
+  // hundred nodes, and the webview stays wedged past the RPC timeout (#146).
+  // Custom properties are dead weight here — computed values arrive with their
+  // `var()` already resolved — so we hand html-to-image the painted properties
+  // only. Anything not in this list is lost from the capture.
+  var STYLE_PROPERTIES = [
+    // Box + layout
+    "display", "position", "top", "right", "bottom", "left", "float", "clear",
+    "z-index", "width", "height", "min-width", "min-height", "max-width",
+    "max-height", "box-sizing", "aspect-ratio", "margin-top", "margin-right",
+    "margin-bottom", "margin-left", "padding-top", "padding-right",
+    "padding-bottom", "padding-left", "overflow-x", "overflow-y", "visibility",
+    "opacity", "vertical-align",
+    // Flex + grid
+    "flex-direction", "flex-wrap", "flex-grow", "flex-shrink", "flex-basis",
+    "justify-content", "justify-items", "justify-self", "align-content",
+    "align-items", "align-self", "order", "row-gap", "column-gap",
+    "grid-template-columns", "grid-template-rows", "grid-template-areas",
+    "grid-auto-flow", "grid-auto-columns", "grid-auto-rows",
+    "grid-column-start", "grid-column-end", "grid-row-start", "grid-row-end",
+    "column-count",
+    // Background + border
+    "background-color", "background-image", "background-position",
+    "background-size", "background-repeat", "background-clip",
+    "background-origin", "background-attachment", "-webkit-background-clip",
+    "border-top-width", "border-right-width", "border-bottom-width",
+    "border-left-width", "border-top-style", "border-right-style",
+    "border-bottom-style", "border-left-style", "border-top-color",
+    "border-right-color", "border-bottom-color", "border-left-color",
+    "border-top-left-radius", "border-top-right-radius",
+    "border-bottom-right-radius", "border-bottom-left-radius",
+    "border-collapse", "border-spacing", "table-layout", "outline-color",
+    "outline-style", "outline-width", "outline-offset", "box-shadow",
+    // Paint effects
+    "filter", "backdrop-filter", "mix-blend-mode", "clip-path", "mask-image",
+    "transform", "transform-origin", "transform-style", "translate", "rotate",
+    "scale", "perspective",
+    // Text
+    "color", "font-family", "font-size", "font-weight", "font-style",
+    "font-variant", "font-stretch", "font-feature-settings",
+    "font-variation-settings", "line-height", "letter-spacing", "word-spacing",
+    "text-align", "text-indent", "text-transform", "text-shadow",
+    "text-overflow", "text-decoration-line", "text-decoration-color",
+    "text-decoration-style", "text-decoration-thickness",
+    "text-underline-offset", "-webkit-text-fill-color", "-webkit-text-stroke",
+    "-webkit-line-clamp", "-webkit-box-orient", "white-space", "word-break",
+    "overflow-wrap", "hyphens", "direction", "unicode-bidi", "writing-mode",
+    "text-orientation", "tab-size", "list-style-type", "list-style-position",
+    "list-style-image", "content",
+    // Replaced content + SVG
+    "object-fit", "object-position", "image-rendering", "fill", "fill-opacity",
+    "fill-rule", "stroke", "stroke-width", "stroke-opacity", "stroke-linecap",
+    "stroke-linejoin", "stroke-dasharray", "stroke-dashoffset", "clip-rule",
+    "stop-color", "stop-opacity", "d",
+  ];
+
   async function screenshot(options) {
     var selector = options && options.selector;
     var el = selector ? document.querySelector(selector) : document.documentElement;
@@ -1195,7 +1256,7 @@
     if (typeof htmlToImage === "undefined" || !htmlToImage.toPng) {
       throw new Error("html-to-image library not loaded. Bundle it into bridge.js for screenshot support.");
     }
-    var renderOptions = { pixelRatio: 1 };
+    var renderOptions = { pixelRatio: 1, includeStyleProperties: STYLE_PROPERTIES };
     if (!selector) {
       // html-to-image sizes the capture from clientWidth/clientHeight, which
       // for documentElement is the viewport — the render always starts at the
--- a/crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
+++ b/crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
@@ -117,3 +117,37 @@ test("screenshot with unmatched selector throws element not found", async () => 
     /Element not found: #missing/,
   );
 });
+
+// #146: html-to-image's per-property clone path (taken whenever
+// `getComputedStyle(el).cssText` is empty — WebKit and Blink both) copies every
+// name returned by `getComputedStyle(document.documentElement)` onto every
+// cloned element. On a Tailwind v4 page that list is ~2200 names, ~1760 of them
+// custom properties, and WebKit re-serializes the whole style attribute on each
+// `setProperty` — quadratic, minutes of wedged webview. The bridge must hand
+// html-to-image a curated list instead.
+test("screenshot restricts the style properties html-to-image copies", async () => {
+  const documentElement = { scrollWidth: 800, scrollHeight: 600 };
+  const { pilot, calls } = loadBridge({ documentElement, body: null });
+
+  await pilot.screenshot({});
+
+  const props = calls[0].options.includeStyleProperties;
+  assert.ok(Array.isArray(props), "includeStyleProperties must be passed");
+  // A few hundred names is already quadratic pain on WebKit; keep it small.
+  assert.ok(props.length < 200, `expected a curated list, got ${props.length}`);
+  assert.equal(props.filter((p) => p.startsWith("--")).length, 0);
+  // Anything the page actually paints has to survive the clone.
+  for (const required of [
+    "color",
+    "background-color",
+    "font-family",
+    "border-top-width",
+    "transform",
+    "transform-origin",
+    "content",
+    "fill",
+  ]) {
+    assert.ok(props.includes(required), `missing ${required}`);
+  }
+  assert.equal(new Set(props).size, props.length, "duplicate property names");
+});
--- a/crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
+++ b/crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
@@ -107,6 +107,9 @@ test("screenshot with selector does not override element dimensions", async () =
   assert.equal("width" in calls[0].options, false);
   assert.equal("height" in calls[0].options, false);
   assert.equal(calls[0].options.pixelRatio, 1);
+  // #146 hits element captures too — the CLI and the MCP tool both take this
+  // path for `--selector`, so the curated list must be passed here as well.
+  assert.ok(Array.isArray(calls[0].options.includeStyleProperties));
 });
 
 test("screenshot with unmatched selector throws element not found", async () => {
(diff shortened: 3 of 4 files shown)


crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -1188,14 +1188,75 @@
...
+    "transform", "transform-origin", "transform-style", "translate", "rotate",
+    "scale", "perspective",
+    // Text
+    "color", "font-family", "font-size", "font-weight", "font-style",
+    "font-variant", "font-stretch", "font-feature-settings",
+    "font-variation-settings", "line-height", "letter-spacing", "word-spacing",
+    "text-align", "text-indent", "text-transform", "text-shadow",
+    "text-overflow", "text-decoration-line", "text-decoration-color",
+    "text-decoration-style", "text-decoration-thickness",
+    "text-underline-offset", "-webkit-text-fill-color", "-webkit-text-stroke",
```
mpiton: > [!WARNING]
> **🟠 High · 🔒 Security** — `-webkit-text-security` is dropped, so author-masked fields capture in cleartext

`-webkit-text-security` is part of the standard computed-style enumeration (476 names in Chromium, verified locally on this branch — `getComputedStyle(el)[i]` lists it and `getPropertyValue` returns `disc`), so before this PR it was copied onto the clone and an author-masked input captured as dots. It is absent from `STYLE_PROPERTIES`, while html-to-image still copies the raw value through `cloneInputValue` (`e.setAttribute("value", t.value)`) — the mask goes, the secret stays. `<input type="password">` is unaffected because the UA stylesheet still masks it inside the foreignObject, so this hits exactly the hand-masked cases: PIN fields, API-key inputs, reveal-secret toggles. Every `tauri-pilot screenshot` writes those values into the PNG, including through the MCP server, which hands the image to a model.

**Fix:** Add `"-webkit-text-security"` to the Text block, and to the `required` array in the new test so it cannot be dropped again. 157 names still clears the `props.length < 200` assertion.

```suggestion
    "text-underline-offset", "-webkit-text-fill-color", "-webkit-text-stroke",
    "-webkit-text-security",
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the property was copied before this diff and is omitted from the curated list the diff introduces
- **References:** CWE-549, CWE-200

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.9</sub>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -1188,14 +1188,75 @@
...
+  // `var()` already resolved — so we hand html-to-image the painted properties
+  // only. Anything not in this list is lost from the capture.
+  var STYLE_PROPERTIES = [
+    // Box + layout
+    "display", "position", "top", "right", "bottom", "left", "float", "clear",
+    "z-index", "width", "height", "min-width", "min-height", "max-width",
+    "max-height", "box-sizing", "aspect-ratio", "margin-top", "margin-right",
+    "margin-bottom", "margin-left", "padding-top", "padding-right",
+    "padding-bottom", "padding-left", "overflow-x", "overflow-y", "visibility",
+    "opacity", "vertical-align",
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🔒 Security** — `content-visibility: hidden` and `clip` subtrees are now painted into the capture

Same mechanism, different consequence: `content-visibility` and `clip` are both in the computed-style enumeration (verified in Chromium on this branch) and both absent from `STYLE_PROPERTIES`. A `content-visibility: hidden` container renders byte-identical to its `visible` control under the curated list and differently under the full list, so a subtree the user cannot see is now painted into the screenshot. `clip: rect(0,0,0,0)` behaves the same way, though it matters less in practice because the usual `.sr-only` recipe also carries `overflow: hidden` and a 1px box, both still copied; `content-visibility: hidden` has no such backstop.

**Fix:** Add `"content-visibility"` and `"clip"` to the Box + layout block.

```suggestion
    "opacity", "vertical-align", "content-visibility", "clip",
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** both properties were copied before this diff and are omitted from the curated list it introduces
- **References:** CWE-200

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.85</sub>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -1188,14 +1188,75 @@
...
+    "text-underline-offset", "-webkit-text-fill-color", "-webkit-text-stroke",
+    "-webkit-line-clamp", "-webkit-box-orient", "white-space", "word-break",
+    "overflow-wrap", "hyphens", "direction", "unicode-bidi", "writing-mode",
+    "text-orientation", "tab-size", "list-style-type", "list-style-position",
+    "list-style-image", "content",
+    // Replaced content + SVG
+    "object-fit", "object-position", "image-rendering", "fill", "fill-opacity",
+    "fill-rule", "stroke", "stroke-width", "stroke-opacity", "stroke-linecap",
+    "stroke-linejoin", "stroke-dasharray", "stroke-dashoffset", "clip-rule",
+    "stop-color", "stop-opacity", "d",
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — Thirteen more painted properties are dropped by the curated list

Four reviewers walking the list independently turned up thirteen painted properties that the computed-style enumeration used to carry and this list does not (each checked against `getComputedStyle` in Chromium on this branch, and against how the vendored bundle consumes the list):

- **mask geometry** — `mask-image` is in, `mask-size`/`mask-position`/`mask-repeat`/`mask-mode`/`mask-composite` are not, so the icon-as-mask pattern (Tailwind v4 `mask-*`, Iconify/UnoCSS `presetIcons`: `mask-image: url(...)` + `mask-size: contain` + `mask-repeat: no-repeat` + `background-color: currentColor`) captures as a tiled, mis-sized smear. `embedProp` still inlines the URL, so it fails as a wrong image rather than a missing one.
- **form controls** — no `appearance` / `-webkit-appearance` / `accent-color`, so a checkbox, radio or `<select>` flattened with `appearance: none` (Tailwind `appearance-none`, `@tailwindcss/forms`) clones back at `appearance: auto` and repaints as a native platform widget.
- **CSS counters** — `content` is copied but its computed value keeps `counter()` unresolved (verified: `counter(s) ". "`), and `counter-reset` / `counter-increment` are not in the list, so numbered steps and custom ordered lists capture as `0.` on every item. Native `<ol>` markers are fine.
- **`column-width`** — `column-count` is kept without it, so a `columns: 20rem` layout collapses to one column.
- **SVG text and markers** — `fill`/`stroke`/`clip-rule` are in, `text-anchor`/`dominant-baseline`/`paint-order`/`marker-*` are not, so a chart whose labels get their anchoring from CSS renders off-anchor and arrowheads disappear.
- **`border-image-*`, `isolation` + `background-blend-mode`, `perspective-origin` + `backface-visibility`, `-webkit-font-smoothing`** — gradient borders vanish, blended elements lose the stacking context that made them blend, 3D transforms rebuild around the wrong origin, `antialiased` text renders heavier.

Thirteen misses in one pass, on the properties reviewers happened to think of, is the argument against the shape rather than against any one entry: the failure mode is a silently wrong PNG with no error anywhere, and the comment on line 1200 already concedes it.

**Fix:** Add the names above to their groups (the suggestion closes the SVG and form-control end; `mask-*` belongs next to `mask-image` on line 1230, the counters next to `content` on line 1244, `column-width` next to `column-count` on line 1216, the rest in the border and paint-effects groups). ~172 names, still under the test's `< 200` bound.

Worth answering in the PR, though: the measurement in the body says the blow-up is the ~1760 custom properties, not the ~480 standard ones. Filtering `--*` off the live enumeration is one line that never goes stale and drops no fidelity at all — `Array.prototype.filter.call(getComputedStyle(document.documentElement), function (p) { return p.indexOf("--") !== 0; })`, ~480 names instead of 156. If that was measured on the macOS repro and does not fit the budget, say so in the comment so the next reader does not re-litigate it; if it was not measured, it is the cheaper fix.

(One reviewer also claimed `-webkit-text-stroke` serializes to `""` on WebKitGTK, which would make that entry a no-op there. Not reproducible here — Blink returns `2px rgb(255, 0, 0)` and no WebKit engine was available to check — so treat it as a question, not a defect: the `-webkit-text-stroke-width` / `-webkit-text-stroke-color` longhands are what both engines enumerate and would sidestep it.)

```suggestion
    "stop-color", "stop-opacity", "text-anchor", "dominant-baseline",
    "paint-order", "marker-start", "marker-mid", "marker-end", "d",
    // Form controls
    "appearance", "-webkit-appearance", "accent-color",
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the list added by this diff is what decides which properties reach the clone; every property named here was carried by the enumerated computed style it replaces

</details>

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.85</sub>

crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
```diff
@@ -117,3 +117,37 @@ test("screenshot with unmatched selector throws element not found", async () =>
...
+// `setProperty` — quadratic, minutes of wedged webview. The bridge must hand
+// html-to-image a curated list instead.
+test("screenshot restricts the style properties html-to-image copies", async () => {
+  const documentElement = { scrollWidth: 800, scrollHeight: 600 };
+  const { pilot, calls } = loadBridge({ documentElement, body: null });
+
+  await pilot.screenshot({});
+
+  const props = calls[0].options.includeStyleProperties;
+  assert.ok(Array.isArray(props), "includeStyleProperties must be passed");
```
mpiton, line 138: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Nothing pins that the vendored bundle still honours `includeStyleProperties`

`loadBridge` stubs `globalThis.htmlToImage`, so this test proves the bridge *passes* the option, never that the library *reads* it. The whole fix hangs on one expression in `js/vendor/html-to-image.iife.js`: `function R(t={}){return g||(t.includeStyleProperties?(g=t.includeStyleProperties,g):(g=f(window.getComputedStyle(document.documentElement)),g))}`. html-to-image ignores options it does not know, and `scripts/build-html-to-image.sh` regenerates that file from the `html-to-image@1.11.13` pin under a DO NOT EDIT banner. A version bump that renames or drops the option puts the ~2200-name enumeration back on every cloned element, #146 returns in full, and the suite stays green.

**Fix:** Assert against the real artifact, not the mock. This test file already imports `readFileSync`, `join` and `here` — read `js/vendor/html-to-image.iife.js` and assert it still mentions `includeStyleProperties`. (The Rust-side `bridge_js_contains_html_to_image_and_pilot` in `crates/tauri-plugin-pilot/src/lib.rs:226` already reads the concatenated bundle and is an equally good home for the same one-line check.)

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR is what makes the capture depend on that vendored option, and it ships no check that the dependency holds

</details>

crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
```diff
@@ -117,3 +117,37 @@ test("screenshot with unmatched selector throws element not found", async () =>
...
+  // A few hundred names is already quadratic pain on WebKit; keep it small.
+  assert.ok(props.length < 200, `expected a curated list, got ${props.length}`);
```
mpiton, line 140: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Selector path unasserted, and the size guard only catches growth

Two mutations both keep the suite green. Moving the `includeStyleProperties` assignment inside the `if (!selector)` block still passes 5/5, because the new test only calls `pilot.screenshot({})` — the `--selector` path the CLI and the MCP tool take for element captures would silently go quadratic again. And deleting the whole 25-name Flex + grid block leaves 131 properties, under the `< 200` ceiling, with none of the eight required names touching that block — every flex/grid layout would collapse in the render and nothing would fail.

**Fix:** Assert `Array.isArray(calls[0].options.includeStyleProperties)` next to the existing `pixelRatio` assertion in the selector test (line ~107), and widen the required set to one name per block of `STYLE_PROPERTIES` so losing a whole category fails — a better guard than adding a numeric floor next to the ceiling.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** both the assignment on bridge.js:1259 and this test are added by the diff

</details>

CHANGELOG.md
```diff
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `screenshot` no longer pins the webview for minutes on style-variable-heavy
+  pages. html-to-image copies computed styles onto the clone, and whenever
+  `getComputedStyle(el).cssText` is empty — WebKit and Blink both — it falls
+  back to one `setProperty` per name in `getComputedStyle(documentElement)`.
+  On a Tailwind v4 page that is ~2200 names, ~1760 of them custom properties,
+  applied to every cloned element; macOS WebKit re-serializes the whole style
+  attribute on each call, making the clone quadratic (~370 s and 100 % CPU on a
+  648-node page, with every later bridge call timing out until it finished).
+  The bridge now passes a curated `includeStyleProperties` list of the
+  properties a page actually paints, so the copy drops from ~2240 to 158
```
> **🔵 Low · 📄 Docs** — Changelog count is wrong and neither it nor the CLI reference states the new fidelity limit

`STYLE_PROPERTIES` holds 156 names, not 158 (counted on this branch, no duplicates), and the same sentence says `~2240` where the line five above says `~2200`. Those two figures are the headline result of the fix. The entry also reads as a pure win — the only loss it admits is custom properties — while the real user-visible cost is that any standard painted property outside the list is dropped too. And `docs/src/content/docs/reference/cli.md:864-869` already carries this command's capture caveats (whole document height, scroll position ignored, `screenshot_native` for a pixel-exact capture); this one is missing, so a user looking at a PNG that does not match the app has nothing to check.

**Fix:** Fix the count and the before-figure, add one sentence to the changelog entry saying a painted property outside the list is missing from the capture, and add the same caveat to the `screenshot` block in `docs/src/content/docs/reference/cli.md` with `screenshot_native` as the escape hatch.

```suggestion
  properties a page actually paints, so the copy drops from ~2200 to 156
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the count and the tradeoff both come from the list this diff introduces; CLAUDE.md requires the changelog entry to be accurate

</details>

mpiton:
## 🧨 mpiton-review — PR #147

> [!CAUTION]
> 🚫 **Changes requested** — 1 blocking issue (0 critical, 1 high) out of 6 findings.
> Posted as a comment: GitHub refuses `REQUEST_CHANGES` on your own PR.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/js/bridge.js` | 0 | 1 | 2 | 0 |
| `crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs` | 0 | 0 | 1 | 1 |

### 🎯 Start here

1. 🟠 **`-webkit-text-security` is dropped, so author-masked fields capture in cleartext** — `crates/tauri-plugin-pilot/js/bridge.js:1240`
2. 🟡 **`content-visibility: hidden` and `clip` subtrees are now painted into the capture** — `crates/tauri-plugin-pilot/js/bridge.js:1208`
3. 🟡 **Thirteen more painted properties are dropped by the curated list** — `crates/tauri-plugin-pilot/js/bridge.js:1249`
4. 🟡 **Nothing pins that the vendored bundle still honours `includeStyleProperties`** — `crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs:135`
5. 🔵 **Selector path unasserted, and the size guard only catches growth** — `crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs:137`

<details>

| | |
|:---|:---|
| Commit reviewed | `605667c88e` |
| Base | `main` |
| Files changed | 3 (+113/-1) |
| Languages | javascript |
| Reviewers | documentation, lang-javascript, maintainability, performance, product-logic, quality, security, tests |

</details>

<sub>Generated by <code>mpiton-review</code>. Reply to a comment to push back — findings are ranked by severity x confidence x reviewer agreement.</sub>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -1188,14 +1188,75 @@
...
+  // `getComputedStyle(el).cssText` is empty — WebKit and Blink both — it falls
+  // back to one `setProperty` per name in `getComputedStyle(documentElement)`,
+  // which on a Tailwind v4 page is ~2200 names (~1760 of them custom
+  // properties). WebKit re-serializes the whole style attribute on every
+  // `setProperty`, so that copy is quadratic: minutes of 100% CPU on a few
+  // hundred nodes, and the webview stays wedged past the RPC timeout (#146).
+  // Custom properties are dead weight here — computed values arrive with their
+  // `var()` already resolved — so we hand html-to-image the painted properties
+  // only. Anything not in this list is lost from the capture.
+  var STYLE_PROPERTIES = [
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Pages using stylesheet-defined `background-blend-mode`, `border-image-*`, or non-default mask sizing/positioning now render differently in screenshots because this allowlist drops those painted properties.

Valid, fixed in `3f91f2a`. All three groups are on the list now: `background-blend-mode` plus `isolation` (a blended element loses the stacking context that made it blend without it), the five `border-image-*` longhands, and `mask-size` / `mask-position` / `mask-repeat` / `mask-mode` / `mask-composite` next to `mask-image`. 156 names to 187.

No visual regression case though, because the bridge tests run under `node --test` with `htmlToImage` stubbed and there is no renderer to diff against. What guards the list instead is `bridge.screenshot.test.mjs`, which now requires one name per block so deleting a whole category fails, plus a caveat in `docs/src/content/docs/reference/cli.md` saying the allowlist is lossy by construction.

crates/tauri-plugin-pilot/js/bridge.js
mpiton: > `-webkit-text-security` is dropped, so author-masked fields capture in cleartext

Fixed in `3f91f2a`. It is in the Text block and in the required set of the test, so it cannot be dropped again silently. The required set grew past that one name: it now covers `display`, `flex-direction`, `background-color`, `border-image-source`, `transform`, `mask-size`, `color`, `font-family`, `content`, `fill`, `text-anchor` and `appearance`, one per block. 187 names total, still under the `< 200` ceiling.

crates/tauri-plugin-pilot/js/bridge.js
mpiton: > `content-visibility: hidden` and `clip` subtrees are now painted into the capture

Fixed in `3f91f2a`. Both sit in the Box + layout block, and both are in the test's required set, so a future trim of that block fails there rather than quietly painting a hidden subtree into the PNG.

crates/tauri-plugin-pilot/js/bridge.js
mpiton: > Four reviewers walking the list independently turned up thirteen painted properties that the computed-style enumeration used to carry and this list does not

All in, `3f91f2a`: the five `mask-*` geometry names next to `mask-image`, `appearance` / `-webkit-appearance` / `accent-color`, `counter-reset` / `counter-increment` next to `content`, `column-width` next to `column-count`, the SVG text and marker set (`text-anchor`, `dominant-baseline`, `paint-order`, `marker-start|mid|end`), the five `border-image-*` longhands, `isolation` + `background-blend-mode`, `perspective-origin` + `backface-visibility`, `-webkit-font-smoothing`. 187 names.

On `-webkit-text-stroke`: not worth arguing about, so I swapped it for `-webkit-text-stroke-width` and `-webkit-text-stroke-color`. Both engines enumerate the longhands, so how the shorthand serializes stops mattering.

On filtering `--*` off the live enumeration instead: not measured, and the arithmetic says it does not fit. The whole fix rests on the copy being quadratic in names-per-element, and the macOS repro is 370 s at ~2200 names on 648 nodes. At ~480 names that is 370 x (480/2200)^2 = ~18 s for a single capture, against the 30 s `SCREENSHOT_TIMEOUT` in the plugin handler, with the webview wedged and every other bridge call blocked for the duration. 187 names lands near 2.7 s by the same arithmetic. Derived rather than measured, since I do not have the macOS box here, but a 6x gap is not something a measurement flips.

So the list stays curated and keeping it current is its cost. That is what the widened test is for: the required set is now one name per block, so losing a whole category fails instead of only growth being caught.

crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
mpiton: > `loadBridge` stubs `globalThis.htmlToImage`, so this test proves the bridge *passes* the option, never that the library *reads* it.

Fixed in `3f91f2a`. New test at the end of the file reads `js/vendor/html-to-image.iife.js` and asserts it still mentions `includeStyleProperties`, with a failure message pointing at `scripts/build-html-to-image.sh` and the upstream pin. Kept it in this file rather than the Rust-side bundle test so the failure lands next to the option it protects.

crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
mpiton: > Moving the `includeStyleProperties` assignment inside the `if (!selector)` block still passes 5/5

Both mutations fail now, `3f91f2a`. The selector test asserts `Array.isArray(calls[0].options.includeStyleProperties)` next to the `pixelRatio` assertion, so the `--selector` path the CLI and the MCP tool take is pinned. And the required set is one name per block instead of eight scattered names: `display`, `flex-direction`, `background-color`, `border-image-source`, `transform`, `mask-size`, `color`, `font-family`, `content`, `fill`, `text-anchor`, `appearance`, plus `content-visibility` / `clip` / `-webkit-text-security` for the three that hide something. Deleting the Flex + grid block fails on `flex-direction` now. Went with block coverage over a numeric floor, same reasoning you gave.

CHANGELOG.md
mpiton: > `STYLE_PROPERTIES` holds 156 names, not 158 [...] and the same sentence says `~2240` where the line five above says `~2200`.

Fixed in `3f91f2a`. Reads `~2200 to 187` now, 187 because the list grew with the properties restored in the other threads. The 28x stays but now says what it measured, the original 156-name cut on the Chromium repro, since I have not re-measured at 187.

Added the sentence you asked for: the list is an allowlist, so a standard painted property outside it is missing from the capture too, with `screenshot_native` as the escape hatch. Same caveat now sits in the `screenshot` block of `docs/src/content/docs/reference/cli.md`, right after the whole-document-height and scroll-position ones.

crates/tauri-plugin-pilot/js/bridge.js
mpiton: Correction to my own number above: the estimate after the fix is ~4 s on that page, not the ~2.7 s I derived. The `370 x (n/2200)^2` model I used is too generous at the small end, since it assumes the whole 370 s is the quadratic term and there is fixed per-node cost underneath it.

Does not change the conclusion — ~4 s against ~18 s for the `--*`-filtered enumeration is the same call — but the figure in the PR body is the right one.

Worth saying where the remaining ~4 s goes: it is N `setProperty` calls per node. Getting under that means one `cssText =` assignment instead, and html-to-image exposes no option for it, so it would mean patching the vendored bundle. Out of scope here.

The two CLI gaps reported alongside #146 are now #149.

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -1188,14 +1188,84 @@
...
+    "text-align", "text-indent", "text-transform", "text-shadow",
+    "text-overflow", "text-decoration-line", "text-decoration-color",
+    "text-decoration-style", "text-decoration-thickness",
+    "text-underline-offset", "-webkit-text-fill-color",
+    "-webkit-text-stroke-width", "-webkit-text-stroke-color",
+    "-webkit-text-security", "-webkit-font-smoothing",
+    "-webkit-line-clamp", "-webkit-box-orient", "white-space", "word-break",
+    "overflow-wrap", "hyphens", "direction", "unicode-bidi", "writing-mode",
+    "text-orientation", "tab-size", "list-style-type", "list-style-position",
+    "list-style-image", "content", "counter-reset", "counter-increment",
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When a page uses `counter-set` with generated `content`, the clone loses the counter value because the allowlist copies only `counter-reset` and `counter-increment`.

Right, added in `451a110`. `counter-set` is the third of the trio and it is enumerated the same as the other two, so leaving it out meant a page that seeds its counters that way numbered every item identically in the capture. 188 names.

docs/src/content/docs/reference/cli.md
```diff
@@ -868,6 +868,13 @@ entire page from the top. Use `--selector` to capture a single element
 (works for elements below the fold too). For a pixel-exact capture of the
 native window, use `screenshot_native`.
 
+The bridge also copies only an allowlist of painted CSS properties onto the
+clone it renders (a full computed-style copy wedges WebKit for minutes on
+style-variable-heavy pages). A page that paints through a property outside
+that list renders without it, with no error — if the PNG does not match what
+you see in the app, that is the first thing to check, and `screenshot_native`
+is the way around it.
```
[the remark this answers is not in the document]

> On Linux and Windows, `screenshot_native` cannot serve as this fallback because the native handler always returns `PERMISSION_DENIED`.

Confirmed and fixed in `451a110`. The non-macOS arm bails with `"screenshot is only available on macOS in this release"` under `PERMISSION_DENIED`, so the sentence was sending Linux and Windows readers to a dead end.

Both places now say so, and I qualified the pre-existing mention two lines above as well rather than leave one recommendation caveated and its twin bare.

crates/tauri-plugin-pilot/js/bridge.screenshot.test.mjs
```diff
@@ -117,3 +120,65 @@ test("screenshot with unmatched selector throws element not found", async () =>
...
+// html-to-image silently ignores options it does not know, so a version bump
+// that renames or drops it would put the ~2200-name enumeration back on every
+// cloned element with the suite still green. Pin the option to the artifact.
+test("the vendored html-to-image still reads includeStyleProperties", () => {
+  const vendor = readFileSync(
+    join(here, "vendor", "html-to-image.iife.js"),
+    "utf8",
+  );
+  assert.ok(
+    vendor.includes("includeStyleProperties"),
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> This regression test can pass while the vendored bundle ignores `includeStyleProperties`, because it checks only token presence

Fair on the token check, tightened in `451a110`: the assertion is `includeStyleProperties?` now, so the name has to appear as the condition of the ternary that picks the option over the full enumeration. The minifier renames the parameter, never the property, so it survives a rebuild of the same upstream code and goes red if a bump drops the branch.

Not doing the behavioural version. Exercising the real clone path means driving `toPng` through `foreignObject`, `XMLSerializer`, `Image` and `canvas`, and this suite deliberately runs on bare `node --test` with no DOM — adding jsdom or a headless browser to pin one option is a worse trade than the string check. The live check for the clone path is Prism, per the project's bridge workflow.

PR #150, opened 2026-09-10
fix: surface RPC error data and drop the bogus scale_factor under tcc_denied

### LINKED ISSUE #149: screenshot_native on macOS: error data is swallowed by the CLI, and scale_factor is nonsense when TCC is denied
Two reporting bugs found while working #146. Neither affects the capture itself, both make a failed or degraded capture hard to diagnose.

## 1. `available_windows` never reaches the user

The plugin builds a helpful payload for `WINDOW_NOT_FOUND`: `render_available_windows()` puts the discovered windows into `error.data` (`crates/tauri-plugin-pilot/src/screenshot/ipc.rs:294`), capped and truncated so the caller can see what it should have passed for `--window-id`.

The CLI throws it away. Every RPC error goes through one line:

```rust
// crates/tauri-pilot-cli/src/client/mod.rs:66
if let Some(err) = response.error {
    bail!("RPC error ({}): {}", err.code, err.message);
}
```

`err.data` is dropped on the floor. So `tauri-pilot screenshot_native --window-id 999 out.png` prints the message and nothing else, and the user has no way to learn the real window ids short of reading the plugin source.

This is not specific to `screenshot_native` — it is the shared error path, so any command whose plugin side attaches structured `data` loses it the same way. Fixing it once in `client/mod.rs` covers all of them.

Suggested shape: when `err.data` is a non-null object, render it under the message rather than discarding it. `available_windows` is a list of `{id, title, ...}`, so a short table or one line per window would do.

## 2. `scale_factor` is derived from mismatched dimensions when `tcc_denied` is true

`compute_scale_factor` divides the captured PNG's pixel width by the target window's logical width (`ipc.rs:316`, `ipc.rs:516`). That only holds when the PNG is actually the window.

When screen-recording permission is missing, `capture_with_fallback` falls back to `CGWindowList` and returns `tcc_denied: true` (`ipc.rs:431`, `ipc.rs:448`). Without permission, `CGWindowListCreateImage` does not return the window contents, so the PNG's width is not the window's width — the division compares a screen-sized image against a window-sized bound and reports things like `scale_factor: 2.87` where the display is 2.0.

The value is documented as stable enough for strict-pixel-diff harnesses to tag artifacts by, so a bogus one is worse than a missing one. Options, roughly in order of preference:

- return `scale_factor: null` (or omit it) when `tcc_denied` is true, since the number cannot be derived from a capture that is not the window
- read the display's backing scale directly instead of inferring it from the image
- keep inferring, but sanity-check the result against the plausible Retina range and fall back to 1.0 outside it

Repro for both: macOS with screen recording denied for the terminal, then `tauri-pilot screenshot_native --window-id <id> out.png` for (1) a bad id and (2) a good one.

mpiton:
Closes #149. Two reporting bugs on the `screenshot_native` path — neither touches the capture itself, both made a failed or degraded capture hard to read.

## 1. `error.data` was dropped by the CLI

`Client::call` turned every JSON-RPC error into `RPC error (code): message` and discarded `err.data`. The plugin builds an `available_windows` list for `WINDOW_NOT_FOUND` precisely so the caller can learn the real ids, and none of it reached the terminal.

The fix lands in the shared error path, not on `screenshot_native`, so any command whose plugin side attaches structured data keeps it. `message` is mirrored into `data` by `rpc_error`, so that key is dropped rather than printed twice.

Before:

```
Error: RPC error (-32602): window_id 999 not found
```

After:

```
Error: RPC error (-32602): window_id 999 not found
{
  "available_windows": [
    { "layer": 0, "owner": "Prism", "title": "Prism — Home", "window_id": 10042 },
    ...
  ],
  "error": "WINDOW_NOT_FOUND"
}
```

## 2. `scale_factor` was derived from mismatched dimensions

`compute_scale_factor` divided the captured PNG's pixel width by the window's logical width. That holds only when the PNG is the window. With screen recording denied, `capture_with_fallback` drops to `CGWindowList`, which does not return the window contents — the division then compares a screen-sized image against a window-sized bound and reports values like `2.87` on a 2.0 display.

It returns `None` when `tcc_denied` is set, so the field serializes to `null`. That is the first option listed in the issue: the number cannot be derived from a capture that is not the window, and the value is documented as stable enough for pixel-diff harnesses to tag artifacts by, so absent beats plausible-but-wrong.

## Test seam

`compute_scale_factor` was `#[cfg(target_os = "macos")]` and took a `WindowBounds` from the CoreGraphics-only `cgwindow` module, so nothing about it could compile — let alone run — off macOS. It now takes an `f64` width and compiles under `cfg(any(target_os = "macos", test))`, which is what let the fix be driven from a failing test on Linux:

- `crates/tauri-plugin-pilot/src/screenshot/ipc.rs` — `test_compute_scale_factor_tcc_denied_reports_no_scale` was red with `left: Some(2.87), right: None`, the exact number from the issue
- `crates/tauri-pilot-cli/tests/rpc_error_data.rs` — mock socket returning the real `WINDOW_NOT_FOUND` payload, asserting the window ids land on stderr

## Not covered

The end-to-end macOS repro (screen recording actually denied for the terminal) still needs a human at the machine — it is a system permission toggle, not something a test can set. The tests cover the decision, not the capture.

## Checks

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check` all clean locally.

---
## Summary by cubic
Fixes two reporting bugs on the `screenshot_native` path, closing #149: RPC error `data` now reaches the terminal, and `scale_factor` is no longer reported when the capture isn't the window (e.g. when screen recording is denied).

**Bug Fixes**
- The CLI renders structured `error.data` (like `available_windows` for `WINDOW_NOT_FOUND`) instead of discarding it; `data.message` is dropped only when it duplicates the top-level message.
- `scale_factor` is `null` unless both pixel and logical dimensions agree on the ratio and the capture's provenance backs the number — a `CGWindowList` capture can only legitimately be 1:1, so a screen-sized image whose aspect matches the window's still gets no scale; `screencapture` also runs with `-o` so its own captures aren't shadow-padded.
- `available_windows` strips titles of windows owned by other processes, since that payload now surfaces on stderr and to MCP clients; `window_id` and `owner` remain for retargeting.

<sup>Written for commit e8bbc1fe608b1a1239b5db76eb121085c6fb7e24. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - RPC errors now include structured error details in CLI output without duplicating messages.
  - Screenshot scale factors are calculated more reliably using capture-source information.
  - macOS native screenshots no longer include window drop shadows, improving image bounds and scaling.
  - Available window metadata now exposes titles for host-app windows while protecting titles from other applications.
- **Tests**
  - Added coverage for RPC error details, screenshot scaling, window metadata, and CLI failure handling.

### DIFF
--- a/crates/tauri-pilot-cli/src/client/mod.rs
+++ b/crates/tauri-pilot-cli/src/client/mod.rs
@@ -68,10 +68,18 @@ impl Client {
             // only way a caller learns, say, the real window ids behind a
             // WINDOW_NOT_FOUND. Render it under the message instead of
             // dropping it (#149). `message` is mirrored into `data` by the
-            // plugin, so drop that key rather than printing it twice.
+            // plugin, so drop that key rather than printing it twice — but
+            // only when it really is the duplicate: the two crates ship
+            // separately, so a producer carrying a distinct `data.message`
+            // must not lose it in the very path that was fixed to stop
+            // losing detail.
             let detail = match err.data {
                 Some(serde_json::Value::Object(mut obj)) => {
-                    obj.remove("message");
+                    if obj.get("message").and_then(serde_json::Value::as_str)
+                        == Some(err.message.as_str())
+                    {
+                        obj.remove("message");
+                    }
                     if obj.is_empty() {
                         String::new()
                     } else {
--- a/crates/tauri-pilot-cli/src/client/mod.rs
+++ b/crates/tauri-pilot-cli/src/client/mod.rs
@@ -64,7 +64,29 @@ impl Client {
         }
 
         if let Some(err) = response.error {
-            bail!("RPC error ({}): {}", err.code, err.message);
+            // Structured detail the plugin attaches under `error.data` is the
+            // only way a caller learns, say, the real window ids behind a
+            // WINDOW_NOT_FOUND. Render it under the message instead of
+            // dropping it (#149). `message` is mirrored into `data` by the
+            // plugin, so drop that key rather than printing it twice.
+            let detail = match err.data {
+                Some(serde_json::Value::Object(mut obj)) => {
+                    obj.remove("message");
+                    if obj.is_empty() {
+                        String::new()
+                    } else {
+                        let data = serde_json::Value::Object(obj);
+                        format!(
+                            "\n{}",
+                            serde_json::to_string_pretty(&data)
+                                .unwrap_or_else(|_| data.to_string())
+                        )
+                    }
+                }
+                Some(data) if !data.is_null() => format!("\n{data}"),
+                _ => String::new(),
+            };
+            bail!("RPC error ({}): {}{detail}", err.code, err.message);
         }
 
         // A missing `result` field (or explicit `"result": null`) means the
--- a/crates/tauri-pilot-cli/tests/rpc_error_data.rs
+++ b/crates/tauri-pilot-cli/tests/rpc_error_data.rs
@@ -75,6 +75,11 @@ fn spawn_mock_window_not_found_server(socket: &PathBuf) -> thread::JoinHandle<()
 
 /// Run the binary against `socket` with a `screenshot_native` call that is
 /// guaranteed to reach the RPC error path, and return its stderr.
+///
+/// Panics before the caller can `join()` the one-shot mock server when the
+/// binary never reached that path — an arg-parse or environment failure exits
+/// non-zero without ever connecting, which would otherwise leave the server
+/// blocked on `accept()` and hang the test run.
 fn stderr_of_failed_screenshot(socket: &Path) -> String {
     let tmpdir = tempfile::tempdir().expect("tempdir");
     let output_path = tmpdir.path().join("shot.png");
@@ -97,7 +102,12 @@ fn stderr_of_failed_screenshot(socket: &Path) -> String {
         !output.status.success(),
         "a JSON-RPC error must exit non-zero"
     );
-    String::from_utf8_lossy(&output.stderr).into_owned()
+    let stderr = String::from_utf8_lossy(&output.stderr).into_owned();
+    if !stderr.contains("RPC error") {
+        let _ = std::fs::remove_file(socket);
+        panic!("binary never reached the RPC error path.\n--- stderr ---\n{stderr}\n--- end ---");
+    }
+    stderr
 }
 
 #[test]
--- a/crates/tauri-pilot-cli/tests/rpc_error_data.rs
+++ b/crates/tauri-pilot-cli/tests/rpc_error_data.rs
@@ -1,0 +1,114 @@
+//! Regression test for issue #149, bug 1: structured `error.data` returned by
+//! the plugin must reach the user instead of being dropped by the shared RPC
+//! error path in `client/mod.rs`.
+//!
+//! Same harness as `snapshot_save_json.rs`: a one-shot mock JSON-RPC unix
+//! socket server answers the single request, then the binary runs against that
+//! socket via `assert_cmd` and its stderr is inspected.
+
+#![cfg(unix)]
+
+use std::io::{BufRead, BufReader, Write};
+use std::os::unix::net::UnixListener;
+use std::path::PathBuf;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::thread;
+
+use assert_cmd::Command;
+
+static SOCK_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+fn unique_socket_path(tag: &str) -> PathBuf {
+    let n = SOCK_COUNTER.fetch_add(1, Ordering::Relaxed);
+    PathBuf::from(format!(
+        "/tmp/tauri-pilot-it-{}-{}-{}.sock",
+        tag,
+        std::process::id(),
+        n
+    ))
+}
+
+/// Answer one request with the exact `WINDOW_NOT_FOUND` payload the macOS
+/// plugin builds in `screenshot/ipc.rs`: `RPC_INVALID_PARAMS`, a message, and
+/// an `available_windows` list carried under `error.data`.
+fn spawn_mock_window_not_found_server(socket: &PathBuf) -> thread::JoinHandle<()> {
+    let _ = std::fs::remove_file(socket);
+    let listener = UnixListener::bind(socket).expect("bind mock socket");
+    thread::spawn(move || {
+        let (stream, _) = listener.accept().expect("accept");
+        let mut reader = BufReader::new(stream.try_clone().expect("clone stream"));
+        let mut writer = stream;
+        let mut line = String::new();
+        reader.read_line(&mut line).expect("read line");
+        let req: serde_json::Value = serde_json::from_str(line.trim()).expect("parse request");
+        let id = req.get("id").cloned().unwrap_or(serde_json::Value::Null);
+        let resp = serde_json::json!({
+            "jsonrpc": "2.0",
+            "id": id,
+            "error": {
+                "code": -32602,
+                "message": "window_id 999 not found",
+                "data": {
+                    "error": "WINDOW_NOT_FOUND",
+                    "message": "window_id 999 not found",
+                    "available_windows": [
+                        {"window_id": 10042, "owner": "Prism", "title": "Prism — Home", "layer": 0},
+                        {"window_id": 20087, "owner": "Finder", "title": "Downloads", "layer": 0}
+                    ]
+                }
+            }
+        });
+        let mut bytes = serde_json::to_vec(&resp).expect("serialize");
+        bytes.push(b'\n');
+        writer.write_all(&bytes).expect("write");
+        writer.flush().expect("flush");
+    })
+}
+
+#[test]
+fn test_rpc_error_with_data_prints_available_windows_on_stderr() {
+    let socket = unique_socket_path("errdata");
+    let handle = spawn_mock_window_not_found_server(&socket);
+
+    let tmpdir = tempfile::tempdir().expect("tempdir");
+    let output_path = tmpdir.path().join("shot.png");
+
+    let output = Command::cargo_bin("tauri-pilot")
+        .expect("cargo_bin")
+        .args([
+            "--socket",
+            socket.to_str().expect("socket path is UTF-8"),
+            "screenshot_native",
+            "--window-id",
+            "999",
+            "--output",
+            output_path.to_str().expect("output path is UTF-8"),
+        ])
+        .output()
+        .expect("run tauri-pilot");
+
+    let stderr = String::from_utf8_lossy(&output.stderr).into_owned();
+
+    // Fail fast before join() so a binary that exits before connecting (e.g.
+    // arg-parse error) cannot leave the mock server blocked on accept() and
+    // hang the test indefinitely.
+    if !stderr.contains("window_id 999 not found") {
+        let _ = std::fs::remove_file(&socket);
+        panic!("binary never reached the RPC error path.\n--- stderr ---\n{stderr}\n--- end ---");
+    }
+    handle.join().expect("mock server join");
+    let _ = std::fs::remove_file(&socket);
+
+    assert!(
+        !output.status.success(),
+        "a JSON-RPC error must exit non-zero"
+    );
+
+    for needle in ["10042", "Prism", "20087", "Finder"] {
+        assert!(
+            stderr.contains(needle),
+            "issue #149: `error.data.available_windows` must reach the user, \
+             missing `{needle}`.\n--- stderr ---\n{stderr}\n--- end ---"
+        );
+    }
+}
--- a/crates/tauri-pilot-cli/tests/rpc_error_data.rs
+++ b/crates/tauri-pilot-cli/tests/rpc_error_data.rs
@@ -10,7 +10,7 @@
 
 use std::io::{BufRead, BufReader, Write};
 use std::os::unix::net::UnixListener;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::thread;
 
@@ -28,10 +28,8 @@ fn unique_socket_path(tag: &str) -> PathBuf {
     ))
 }
 
-/// Answer one request with the exact `WINDOW_NOT_FOUND` payload the macOS
-/// plugin builds in `screenshot/ipc.rs`: `RPC_INVALID_PARAMS`, a message, and
-/// an `available_windows` list carried under `error.data`.
-fn spawn_mock_window_not_found_server(socket: &PathBuf) -> thread::JoinHandle<()> {
+/// Answer one request with `error` as the JSON-RPC error object.
+fn spawn_mock_error_server(socket: &PathBuf, error: serde_json::Value) -> thread::JoinHandle<()> {
     let _ = std::fs::remove_file(socket);
     let listener = UnixListener::bind(socket).expect("bind mock socket");
     thread::spawn(move || {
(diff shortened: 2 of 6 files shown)


crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -64,7 +64,29 @@ impl Client {
...
+                        format!(
+                            "\n{}",
+                            serde_json::to_string_pretty(&data)
+                                .unwrap_or_else(|_| data.to_string())
+                        )
+                    }
+                }
+                Some(data) if !data.is_null() => format!("\n{data}"),
+                _ => String::new(),
+            };
```
mpiton, line 96: > [!IMPORTANT]
> **🟡 Medium · 🎯 Product** — Rendered error.data ships other apps' window titles to stderr and to the MCP client

The `WINDOW_NOT_FOUND` `error.data` built at `screenshot/ipc.rs:293` carries up to 20 entries from `enumerate_layer_zero_windows`, which walks the whole on-screen list with `kCGNullWindowID` and records `owner` plus `title` for every application, not just the host Tauri app. Rendering that payload here means `tool_error(err)` in `mcp.rs:1138` stringifies it into `CallToolResult::structured_error`, so one `pilot.screenshot_native` call with a stale `window_id` ships a developer's document names, URLs and chat window titles to whatever model backs the MCP client. Before this diff the payload was dropped at the CLI boundary and never left the process.

**Fix:** The caller needs the ids, not the titles. Either blank or omit `title` for windows the host process does not own in `render_available_windows` (`screenshot/ipc.rs:356`), or render only `window_id` and `owner` in the CLI detail block and keep the full payload behind verbose output.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the payload existed before but was dropped at the CLI boundary; this diff is what makes it leave the process
- **References:** CWE-200

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: product-logic, security · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -313,7 +313,7 @@ fn run_macos(req: ScreenshotRequest) -> Result<Value, RpcError> {
         width: 0.0,
         height: 0.0,
     });
-    let scale_factor = compute_scale_factor(metadata.width, logical_bounds);
+    let scale_factor = compute_scale_factor(metadata.width, logical_bounds.width, tcc_denied);
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — tcc_denied is a permission flag, not a statement about the capture geometry

The flag is wrong in both directions for this decision. On the path the PR keeps, `capture_screencapture` runs `screencapture -x -l <id>` with no `-o` (screencapture.rs:20-27), so the PNG carries the window's drop shadow while `target_bounds` comes from `kCGWindowBounds`, which does not - the ratio lands above the real scale on a permitted capture too, which is the same complaint #149 opens with. In the other direction, `capture_with_fallback` returns `tcc_denied = true` after a *granted* probe whose per-window `screencapture` call failed (ipc.rs:426-431); there `capture_cgwindow` asks for `kCGWindowListOptionIncludingWindow | kCGWindowImageBoundsIgnoreFraming | kCGWindowImageNominalResolution` (cgwindow.rs:30-37), so the image really is the window at 1x and the scale was derivable - it is now `null`.

**Fix:** Gate the derivation on the geometry rather than on the flag: pass `metadata.height` and `logical_bounds.height` too and return `None` unless both dimensions agree on the same ratio. That covers the shadow case, the denied case and any future backend without a second source of truth. Measure the shadow first - `screencapture -x -l <id> a.png` vs `screencapture -x -o -l <id> b.png`, then `sips -g pixelWidth` on both; if `-o` is what makes the primary capture match `kCGWindowBounds`, that belongs in `capture_screencapture`. Related: the new doc comment and the PR body both say the `CGWindowList` fallback "returns a screen-sized image rather than the window", but cgwindow.rs asks for the window only, at nominal resolution. If #149's 2.87 is reproducible, the flags are not behaving as requested under denial and the comment should say that instead.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the PR closes #149 on the premise that the number is only bogus under tcc_denied; line 316 is where that premise is wired in, and the same mismatch exists on the path it leaves untouched

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: product-logic, quality · confidence 0.75</sub>

crates/tauri-pilot-cli/tests/rpc_error_data.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+    handle.join().expect("mock server join");
+    let _ = std::fs::remove_file(&socket);
+
+    assert!(
+        !output.status.success(),
+        "a JSON-RPC error must exit non-zero"
+    );
+
+    for needle in ["10042", "Prism", "20087", "Finder"] {
```
mpiton, line 159: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Nothing asserts the mirrored error.data.message is printed only once

The fixture at lines 51-58 mirrors `message` into `data` exactly as the plugin's `rpc_error()` does (`screenshot/ipc.rs:47`, unconditional for every `screenshot_native` error), but the only assertions are four `available_windows` needles plus a non-zero exit. Deleting `obj.remove("message")` from `client/mod.rs:74` makes stderr print `window_id 999 not found` twice and every assertion in this test still passes. Half of bug 1's stated fix has no coverage.

**Fix:** Assert the count, not just the presence: `stderr.matches("window_id 999 not found").count()` must be 1. Put it before the needle loop so the dedup and the payload are pinned by the same test.

```suggestion
    assert_eq!(
        stderr.matches("window_id 999 not found").count(),
        1,
        "the `message` mirrored into `error.data` must not be printed twice.\n--- stderr ---\n{stderr}\n--- end ---"
    );

    for needle in ["10042", "Prism", "20087", "Finder"] {
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the `obj.remove("message")` dedup and this test are both added by this diff

</details>

crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -64,7 +64,29 @@ impl Client {
...
         if let Some(err) = response.error {
-            bail!("RPC error ({}): {}", err.code, err.message);
+            // Structured detail the plugin attaches under `error.data` is the
+            // only way a caller learns, say, the real window ids behind a
+            // WINDOW_NOT_FOUND. Render it under the message instead of
+            // dropping it (#149). `message` is mirrored into `data` by the
+            // plugin, so drop that key rather than printing it twice.
+            let detail = match err.data {
+                Some(serde_json::Value::Object(mut obj)) => {
+                    obj.remove("message");
```
mpiton: > [!NOTE]
> **🔵 Low · 🎯 Product** — data.message is dropped unconditionally, not only when it duplicates err.message

The removal is keyed to a convention that lives in one helper of the other crate (`screenshot/ipc.rs:43-58` mirrors `message` into `data`; every other `RpcError` in the plugin sets `data: None`). The CLI and the plugin ship as separately versioned crates, and `rpc_error` merges `extras` *after* the mirror, so any producer that puts a distinct `message` inside `data` - a future error type, or an older/newer plugin - loses that string silently in the one path that was just fixed to stop losing detail.

**Fix:** Remove the key only when it is the duplicate: `if obj.get("message").and_then(Value::as_str) == Some(err.message.as_str()) { obj.remove("message"); }`, leaving a divergent message visible. Dropping the mirror in `rpc_error` instead would work too and would delete this special case entirely, but that is a plugin-side contract change.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the unconditional remove is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: product-logic, quality · confidence 0.9</sub>

mpiton:
## 🧨 mpiton-review — PR #150

> [!CAUTION]
> ⚠️ **Review summary** — No blocking issue. 5 findings to consider. Posted as a comment: GitHub refuses `REQUEST_CHANGES` on your own PR.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/src/client/mod.rs` | 0 | 0 | 1 | 1 |
| `crates/tauri-plugin-pilot/src/screenshot/ipc.rs` | 0 | 0 | 2 | 0 |
| `crates/tauri-pilot-cli/tests/rpc_error_data.rs` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟡 **Rendered error.data ships other apps' window titles to stderr and to the MCP client** — `crates/tauri-pilot-cli/src/client/mod.rs:88`
2. 🟡 **tcc_denied is a permission flag, not a statement about the capture geometry** — `crates/tauri-plugin-pilot/src/screenshot/ipc.rs:316`
3. 🟡 **Nothing asserts the mirrored error.data.message is printed only once** — `crates/tauri-pilot-cli/tests/rpc_error_data.rs:107`
4. 🔵 **data.message is dropped unconditionally, not only when it duplicates err.message** — `crates/tauri-pilot-cli/src/client/mod.rs:74`

<details>

- 🟡 **handle_screenshot doc comment still lists scale_factor as an unconditional response field** — `crates/tauri-plugin-pilot/src/screenshot/ipc.rs:221` — This doc comment is the only description of the `screenshot_native` success payload anywhere in the repo, and it lists `scale_factor` next to `output_path`, `byte_size` and `backend` as if it were always populated. After this PR `compute_scale_factor` returns `None` when `tcc_denied` is true, so the field serializes as `null`; a caller who implements against this comment reads it as a number and b
  - *Fix:* Say in the payload description that `scale_factor` is `null` when `tcc_denied` is true, and why - the rationale already exists on `compute_scale_factor` (lines 511-522), but line 219-222 is the contract readers hit first. `docs/src/content/docs/reference/cli.md` mentions `screenshot_native` only in 
  - *Link to this PR:* the PR changes `compute_scale_factor` to `Option<f32>` (line 524) and feeds it into the payload at line 348, making the field nullable, but leaves the payload description at line 221 unchanged

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `e36396a418` |
| Base | `main` |
| Files changed | 4 (+191/-12) |
| Languages | rust |

</details>

crates/tauri-pilot-cli/src/client/mod.rs
mpiton: > Rendered error.data ships other apps' window titles to stderr and to the MCP client

Fixed in `cd4bdfd`, at the source rather than in the renderer. `enumerate_layer_zero_windows` reads `kCGWindowOwnerPID` and keeps `title` only when the window belongs to the host process; every other entry comes back with an empty string. Doing it in `window_id.rs` covers stderr and `tool_error` in one place, and `window_id` + `owner` are still enough to retarget. `window_not_found_returns_available_list` now asserts every entry is title-less — the test binary owns no on-screen window, so that holds on the macOS runner.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
mpiton: > tcc_denied is a permission flag, not a statement about the capture geometry

Took the geometry route in `cd4bdfd`. `compute_scale_factor` takes both axes and returns `None` unless the two ratios match after rounding to hundredths; `tcc_denied` is out of the derivation entirely. `capture_screencapture` runs with `-o` now, so the shadow stops padding the PNG past `kCGWindowBounds` — without it the permitted path reads 2.15 across and 2.23 down on a window that is plainly 2.0.

I couldn't measure the shadow, no Mac on this machine. `-o` is the man page's word, not an observation. If it turns out not to apply to `-l`, the axes never agree and the field is `null` everywhere: still the wrong answer, but not a lying one.

Kept the screen-sized claim in the doc comment. #149's 2.87 is 4344/1512, and 4344 px is a 16" panel at 2172 pt, so the flags really aren't honoured under denial — I reworded it to say that instead of implying `CGWindowList` was asked for the screen. One ceiling worth naming: two ratios agreeing isn't proof, so a window whose aspect happens to match the display's would still slip through on a denied capture. Reading the backing scale off `CGDisplayMode` is what removes the inference for good.

crates/tauri-pilot-cli/tests/rpc_error_data.rs
mpiton: > Nothing asserts the mirrored error.data.message is printed only once

Added in `cd4bdfd`, count first, then the needle loop. Checked that it bites: swap `obj.remove("message")` for a no-op and the test fails. The other half of the dedup has its own test now, `test_rpc_error_data_message_distinct_from_error_message_is_kept`.

crates/tauri-pilot-cli/src/client/mod.rs
mpiton: > data.message is dropped unconditionally, not only when it duplicates err.message

Keyed on the duplicate in `cd4bdfd`, the `obj.get("message") == Some(err.message.as_str())` form. Left `rpc_error`'s mirror alone: that one is a plugin-side contract change and the two crates version separately. The divergent case has a test.

mpiton:
On the finding outside the inline cap, `crates/tauri-plugin-pilot/src/screenshot/ipc.rs:221`:

> handle_screenshot doc comment still lists scale_factor as an unconditional response field

Fixed in `cd4bdfd`. The payload description says `scale_factor` is `null` when the capture's two axes disagree on the ratio against the window's logical bounds, and points at `compute_scale_factor` for the reasoning. The condition itself moved while I was there: it's the geometry now, not `tcc_denied` — see the reply on ipc.rs:316.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -509,34 +518,71 @@ fn read_png_metadata(path: &Path) -> Result<PngMetadata, RpcError> {
...
+///
+/// The geometry decides, not `tcc_denied` — that flag reports a permission
+/// state, and `capture_with_fallback` also raises it after a *granted* probe
+/// whose per-window `screencapture` call failed, where the fallback image is
+/// the window at 1x and the scale was derivable. The primary backend passes
+/// `screencapture -o` so its own PNG is not padded by the window's shadow.
+///
+/// The check compares two ratios, so a window whose aspect happens to match
+/// the screen's would still pass on a denied capture. Reading the display's
+/// backing scale from `CGDisplayMode` would remove the inference entirely.
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> `compute_scale_factor` compares only PNG dimensions with `CGWindowBounds`. A non-window fallback with the same aspect ratio can therefore pass and report a plausible scale.

Valid, fixed in `e8bbc1f`. Geometry alone can't tell a half-size screen from the window. `compute_scale_factor` now takes a `nominal_resolution` flag from the call site: `capture_cgwindow` requests `kCGWindowImageNominalResolution`, so a real capture from that backend is 1:1 and nothing else, and any other ratio means CoreGraphics ignored the flags and handed back the screen. That case returns `None`.

Passing the flag rather than the `ScreenshotBackend` enum because the enum is `#[cfg(target_os = "macos")]` while the helper compiles under `test` on Linux too.

Covered by `test_compute_scale_factor_screen_sized_capture_at_window_aspect_reports_no_scale` (4344x2744 against a 1086x686 pt window, both ratios 4.0) and `test_compute_scale_factor_nominal_window_capture_reports_one`.

crates/tauri-plugin-pilot/src/screenshot/ipc.rs
```diff
@@ -509,34 +518,71 @@ fn read_png_metadata(path: &Path) -> Result<PngMetadata, RpcError> {
...
+fn compute_scale_factor(
+    pixel_width: u32,
+    pixel_height: u32,
+    logical_width: f64,
+    logical_height: f64,
+) -> Option<f32> {
+    if logical_width <= 0.0 || logical_height <= 0.0 {
+        return Some(1.0);
+    }
+    let horizontal = rounded_ratio(pixel_width, logical_width)?;
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> P2: When a denied CGWindow fallback has the same aspect ratio as the target window, this geometry-only check reports a plausible scale instead of `null`.

Right, fixed in `e8bbc1f`. The dual-axis check narrows the hole but doesn't close it — a screen at exactly half the window's size in both directions still slips through with a clean 4.0.

`compute_scale_factor` now gets the capture's provenance. `capture_cgwindow` asks for `kCGWindowImageNominalResolution`, which means a legitimate PNG from it is 1:1; under TCC denial CoreGraphics drops the flags and returns the screen, so any other ratio from that backend is proof the pixels aren't the window and the scale comes back `None`.

Regression test: `test_compute_scale_factor_screen_sized_capture_at_window_aspect_reports_no_scale` in `crates/tauri-plugin-pilot/src/screenshot/ipc.rs`.

crates/tauri-pilot-cli/tests/rpc_error_data.rs
```diff
@@ -0,0 +1,184 @@
...
+            "code": -32602,
+            "message": "window_id 999 not found",
+            "data": {
+                "error": "WINDOW_NOT_FOUND",
+                "message": "the window closed between enumeration and capture"
+            }
+        }),
+    );
+
+    let stderr = stderr_of_failed_screenshot(&socket);
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> `stderr_of_failed_screenshot` only asserts the binary exits non-zero; it never verifies the binary contacted the mock server before the test calls `handle.join()`.

Valid, fixed in `e8bbc1f`. An arg-parse failure exits non-zero without ever connecting, and then `join()` waits on a thread parked in `accept()` forever — no test timeout to save it.

The helper now checks stderr for `RPC error` and panics there, before returning, so the failure surfaces with the captured stderr instead of hanging the run. Same guard the older `test_rpc_error_with_data_prints_available_windows_on_stderr` already had inline.

PR #170, opened 2026-09-10
fix(plugin): fail fast when the page cannot answer pilot callbacks

### LINKED ISSUE #153: navigate to an external origin breaks the session and still reports ok
`navigate` to an origin outside the app leaves the session permanently broken, and reports success while doing it.

## Reproduce

```
$ tauri-pilot navigate https://example.com
✓ ok

$ tauri-pilot title
Error: eval timed out after 10s

$ tauri-pilot text body
Error: eval timed out after 10s

$ tauri-pilot windows
main  https://example.com/          # still works
```

Every eval-based command now blocks for the full 10 s timeout and fails. Navigating back to the app URL also works but reports the same timeout error, because the response is sent from the page that is being torn down.

## Cause

`crates/tauri-plugin-pilot/js/bridge.js:867`:

```js
function navigate(options) {
  const url = options && options.url;
  if (url) window.location.href = url;
  return { ok: true };
}
```

The bridge is injected through `js_init_script()` and only exists on the app's own pages. Setting `window.location.href` to an external origin discards the bridge; the new document never has the pilot globals or the `plugin:pilot|__callback` IPC handler, so no eval callback can ever be delivered. The `{ ok: true }` is returned from a context that is already gone.

## Expected

Two things are wrong and both are worth fixing:

- **Reporting.** `navigate` should not claim `ok` for a navigation whose outcome it cannot observe. It should either resolve only once the destination has a working bridge, or return a result that says the bridge was dropped.
- **Recovery.** After landing on a bridge-less origin the CLI has no way back other than the user closing the app. Commands that need the bridge should fail fast with an actionable message (something like "no pilot bridge on the current origin (https://example.com/) — navigate back to the app origin") instead of hanging for 10 s each.

At minimum, detecting that the target URL is not the app origin and refusing (or warning loudly) would prevent a QA run from silently bricking itself mid-scenario.

mpiton:
Fixes #153.

## Summary

• Root cause: the bridge does run on a foreign origin, but Tauri's ACL rejects `plugin:pilot|__callback` there (`allow-eval-callbacks` only covers the app origin unless a capability lists the origin in `remote.urls`). A denied call sends nothing, so every eval waited out its 10 s timeout.
• The bridge sends a hello on each page load (`__callback`, reserved id 0, `location.href`). `EvalEngine` records the origins that said hello.
• A command on a page whose origin never said hello fails at once, and the error names the origins that can answer.
• `navigate` to an origin without a hello waits up to 3 s for one, then fails instead of reporting ok. Navigating back recovers.
• Pilot's own IPC no longer shows up in `network.getRequests`.

Known limits: hellos are tracked per origin, not per window, and a slow remote page allowed by `remote.urls` can miss the 3 s grace on its first visit.

#169 is stacked on this branch.

## Type

fix

---
## Summary by cubic
Fixes #153 so pilot commands fail fast instead of timing out when the current page can't call back. The bridge now sends a hello on each page load, letting the plugin track which origins can answer; commands on an unknown origin fail immediately, and `navigate` to one waits 3s before failing. The error names working origins, and navigating back restores the session.

**Bug Fixes**
- Bridge sends a hello with reserved id 0 on each page load; the plugin records the invoking webview's origin by scheme, host, and port, dropping hellos whose payload disagrees.
- Origin matching shares default ports and treats an `http` destination as matching its `https` upgrade.
- Commands on a page without a bridge return an error immediately, naming the origins that can answer.
- `navigate` parses absolute URLs without a page base, treats `javascript:` as same-document, and waits up to 3s for a destination hello before failing; a departing page that never callbacks doesn't stop that wait.
- Pilot's own IPC no longer appears in `network.getRequests`.
- Known limit: hellos are tracked per origin, not per window, and a slow remote page allowed by `remote.urls` may miss the 3s grace on first visit.

<sup>Written for commit 709a70a771336786e8950ad7a94fcc9a43e5f3d1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Navigation now verifies destination bridge availability and recovers when returning to supported pages.
  * Added controlled access to remote origins through capability-based URL allowlists.
  * Pilot IPC requests are excluded from application network traffic reporting while matching application URLs remain recorded.

* **Bug Fixes**
  * Improved cross-origin navigation, origin matching, bridge detection, and page-specific error handling.
  * Navigation now applies appropriate wait times for new and previously recognized origins.

* **Documentation**
  * Documented remote-origin permissions, bridge handshakes, navigation timeouts, recovery behavior, and network request filtering.

### DIFF
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -143,9 +143,14 @@
       const q = text.indexOf("?");
       path = q === -1 ? text : text.slice(0, q);
     }
-    // Tauri encodes the command in the IPC path (`.../plugin%3Apilot%7C__callback`).
-    // A query string that happens to contain the same text is still app traffic.
-    return path.indexOf("plugin%3Apilot%7C") !== -1;
+    let decoded;
+    try {
+      decoded = decodeURIComponent(path);
+    } catch (_) {
+      decoded = path;
+    }
+    // Exact IPC path for the two callback commands, not a substring match.
+    return decoded === "/plugin:pilot|callback" || decoded === "/plugin:pilot|__callback";
   }
 
   const _originalFetch = window.fetch.bind(window);
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -133,12 +133,27 @@
     return 0;
   }
 
+  function isPilotIpcUrl(url) {
+    const text = String(url);
+    let path;
+    try {
+      const base = (window.location && window.location.href) || "http://localhost/";
+      path = new URL(text, base).pathname;
+    } catch (_) {
+      const q = text.indexOf("?");
+      path = q === -1 ? text : text.slice(0, q);
+    }
+    // Tauri encodes the command in the IPC path (`.../plugin%3Apilot%7C__callback`).
+    // A query string that happens to contain the same text is still app traffic.
+    return path.indexOf("plugin%3Apilot%7C") !== -1;
+  }
+
   const _originalFetch = window.fetch.bind(window);
   window.fetch = function(input, init) {
     const method = (init && init.method) || (input && input.method) || "GET";
     const url = (typeof input === "string") ? input : (input && input.url) || String(input);
     // Pilot's own IPC (eval callbacks, the bridge hello) is not app traffic.
-    if (url.indexOf("plugin%3Apilot%7C") !== -1) return _originalFetch(input, init);
+    if (isPilotIpcUrl(url)) return _originalFetch(input, init);
     const timestamp = Date.now();
     const requestSize = bodySize(init && init.body);
     return _originalFetch(input, init).then(function(response) {
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -137,6 +137,8 @@
   window.fetch = function(input, init) {
     const method = (init && init.method) || (input && input.method) || "GET";
     const url = (typeof input === "string") ? input : (input && input.url) || String(input);
+    // Pilot's own IPC (eval callbacks, the bridge hello) is not app traffic.
+    if (url.indexOf("plugin%3Apilot%7C") !== -1) return _originalFetch(input, init);
     const timestamp = Date.now();
     const requestSize = bodySize(init && init.body);
     return _originalFetch(input, init).then(function(response) {
@@ -1435,4 +1437,14 @@
     storageClear: storageClear,
     formDump: formDump,
   };
+
+  // Tell the plugin this origin can answer (#153). The ACL denies
+  // `__callback` to origins without the pilot permission, and a denied page
+  // can run the bridge but never deliver a result, so its silence here is the
+  // signal. Id 0 is HELLO_ID in eval.rs, never used by an eval request.
+  try {
+    window.__TAURI_INTERNALS__
+      .invoke("plugin:pilot|__callback", { id: 0, result: location.href })
+      .catch(function() {});
+  } catch (_) {}
 })();
--- a/crates/tauri-plugin-pilot/src/eval.rs
+++ b/crates/tauri-plugin-pilot/src/eval.rs
@@ -3,7 +3,33 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use tokio::sync::oneshot;
+use tauri::Url;
+use tokio::sync::{oneshot, watch};
+
+/// Callback id the bridge sends its hello with when a page loads (#153).
+///
+/// Eval request ids start at 1 (see [`EvalEngine::new`]), so 0 never collides
+/// with a pending eval. `bridge.js` hardcodes the same value.
+pub(crate) const HELLO_ID: u64 = 0;
+
+/// Origins whose bridge said hello, each with the hello count at its latest hello.
+#[derive(Debug, Default)]
+struct Bridges {
+    hellos: u64,
+    latest: HashMap<String, u64>,
+}
+
+/// Key a URL by scheme, host and port.
+///
+/// `Url::origin` is opaque for custom schemes such as `tauri://`, which would
+/// make every app page its own origin, so the key is built by hand.
+fn origin_key(url: &Url) -> String {
+    match (url.host_str(), url.port()) {
+        (Some(host), Some(port)) => format!("{}://{host}:{port}", url.scheme()),
+        (Some(host), None) => format!("{}://{host}", url.scheme()),
+        (None, _) => format!("{}:", url.scheme()),
+    }
+}
 
 /// Error types for eval operations.
 #[derive(Debug, thiserror::Error)]
@@ -27,6 +53,7 @@ pub(crate) struct EvalEngine {
     pending: Arc<Mutex<PendingMap>>,
     next_id: Arc<AtomicU64>,
     last_snapshot: Arc<Mutex<Option<serde_json::Value>>>,
+    bridges: Arc<watch::Sender<Bridges>>,
 }
 
 impl EvalEngine {
@@ -35,7 +62,57 @@ impl EvalEngine {
             pending: Arc::new(Mutex::new(HashMap::new())),
             next_id: Arc::new(AtomicU64::new(1)),
             last_snapshot: Arc::new(Mutex::new(None)),
+            bridges: Arc::new(watch::Sender::new(Bridges::default())),
         }
+    }
+
+    /// Record the hello the bridge sends from `page` (its `location.href`).
+    ///
+    /// Only pages whose origin may call `__callback` can say hello, so the
+    /// hellos tell which origins answer bridge commands (#153).
+    pub fn bridge_hello(&self, page: &str) {
+        let Ok(url) = Url::parse(page) else {
+            tracing::warn!(page, "bridge hello with an invalid page URL");
+            return;
+        };
+        let key = origin_key(&url);
+        self.bridges.send_modify(|b| {
+            b.hellos += 1;
+            b.latest.insert(key, b.hellos);
+        });
+    }
+
+    /// Number of bridge hellos received so far.
+    pub fn hellos(&self) -> u64 {
+        self.bridges.borrow().hellos
+    }
+
+    /// Whether a bridge on the origin of `url` can answer.
+    ///
+    /// Also `true` before any hello: without one the engine cannot tell app
+    /// origins from foreign ones, so callers keep the plain eval path.
+    pub fn has_bridge(&self, url: &Url) -> bool {
+        let bridges = self.bridges.borrow();
+        bridges.hellos == 0 || bridges.latest.contains_key(&origin_key(url))
+    }
+
+    /// Origins whose bridge said hello, sorted.
+    pub fn bridge_origins(&self) -> Vec<String> {
+        let mut origins: Vec<String> = self.bridges.borrow().latest.keys().cloned().collect();
+        origins.sort();
+        origins
+    }
+
+    /// Wait for a hello from the origin of `url` newer than hello number `since`.
+    ///
+    /// Returns `false` when none arrives within `limit`.
+    pub async fn wait_bridge(&self, url: &Url, since: u64, limit: Duration) -> bool {
+        let key = origin_key(url);
+        let mut rx = self.bridges.subscribe();
+        let hello = rx.wait_for(|b| b.latest.get(&key).is_some_and(|&n| n > since));
+        tokio::time::timeout(limit, hello)
+            .await
+            .is_ok_and(|seen| seen.is_ok())
     }
 
     /// Store the last snapshot result for later diff comparison.
@@ -250,6 +327,46 @@ mod tests {
         assert!(
             script.contains("__r===undefined?'null':JSON.stringify(__r)"),
             "wrapped script must normalize undefined to the string 'null'; got: {script}"
+        );
+    }
+
+    #[test]
+    fn test_has_bridge_keys_by_scheme_host_and_port() {
+        let engine = EvalEngine::new();
+        let url = |text| Url::parse(text).expect("valid test URL");
+        // Before any hello the engine cannot tell origins apart.
+        assert!(engine.has_bridge(&url("https://example.com/")));
+
+        engine.bridge_hello("tauri://localhost/index.html");
+        // `Url::origin` is opaque for custom schemes; app pages must share a key.
+        assert!(engine.has_bridge(&url("tauri://localhost/settings")));
+        assert!(!engine.has_bridge(&url("https://example.com/")));
+
+        engine.bridge_hello("http://127.0.0.1:8080/");
+        assert!(!engine.has_bridge(&url("http://127.0.0.1:9090/")));
+        assert_eq!(
+            engine.bridge_origins(),
+            ["http://127.0.0.1:8080", "tauri://localhost"]
+        );
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_wait_bridge_ignores_hellos_before_since() {
+        let engine = EvalEngine::new();
+        let app = Url::parse("tauri://localhost/").expect("valid test URL");
+        engine.bridge_hello(app.as_str());
+        let since = engine.hellos();
+        assert!(
+            !engine
+                .wait_bridge(&app, since, Duration::from_secs(1))
+                .await,
+            "an old hello must not count as the new page's"
+        );
+        engine.bridge_hello(app.as_str());
+        assert!(
+            engine
+                .wait_bridge(&app, since, Duration::from_secs(1))
+                .await
         );
     }
 
--- a/crates/tauri-plugin-pilot/src/eval.rs
+++ b/crates/tauri-plugin-pilot/src/eval.rs
@@ -121,16 +121,11 @@ impl EvalEngine {
     ///
     /// Also `true` before any hello: without one the engine cannot tell app
     /// origins from foreign ones, so callers keep the plain eval path.
-    /// An `http` URL also matches a hello from the `https` upgrade of the
-    /// same host and port, so an HSTS hop still counts.
+    /// Scheme-specific: a previous `https` hello does not make `http` look
+    /// callable. [`wait_bridge`] still accepts a fresh `https` upgrade.
     pub fn has_bridge(&self, url: &Url) -> bool {
         let bridges = self.bridges.borrow();
-        if bridges.hellos == 0 {
-            return true;
-        }
-        origin_keys(url)
-            .iter()
-            .any(|key| bridges.latest.contains_key(key))
+        bridges.hellos == 0 || bridges.latest.contains_key(&origin_key(url))
     }
 
     /// Origins whose bridge said hello, sorted.
@@ -399,8 +394,8 @@ mod tests {
         engine.bridge_hello("https://example.com/");
         assert!(engine.has_bridge(&url("https://example.com:443/login")));
         assert!(
-            engine.has_bridge(&url("http://example.com/")),
-            "http dest must match an https hello (HSTS)"
+            !engine.has_bridge(&url("http://example.com/")),
+            "a previous https hello must not make http look callable"
         );
         assert!(!engine.has_bridge(&url("https://other.example/")));
 
@@ -409,6 +404,28 @@ mod tests {
         assert!(
             !engine.has_bridge(&url("file:///tmp/b.html")),
             "hostless pages must not share a key"
+        );
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_wait_bridge_http_dest_matches_fresh_https_hello() {
+        let engine = EvalEngine::new();
+        let https = Url::parse("https://example.com/").expect("valid test URL");
+        let http = Url::parse("http://example.com/").expect("valid test URL");
+        engine.bridge_hello(https.as_str());
+        let since = engine.hellos();
+        assert!(
+            !engine
+                .wait_bridge(&http, since, Duration::from_secs(1))
+                .await,
+            "an old https hello must not count as this navigation's upgrade"
+        );
+        engine.bridge_hello(https.as_str());
+        assert!(
+            engine
+                .wait_bridge(&http, since, Duration::from_secs(1))
+                .await,
+            "a hello after since from the https upgrade must count"
         );
     }
 
--- a/crates/tauri-plugin-pilot/src/eval.rs
+++ b/crates/tauri-plugin-pilot/src/eval.rs
@@ -23,11 +23,41 @@ struct Bridges {
 ///
 /// `Url::origin` is opaque for custom schemes such as `tauri://`, which would
 /// make every app page its own origin, so the key is built by hand.
-fn origin_key(url: &Url) -> String {
-    match (url.host_str(), url.port()) {
+/// Default ports are always written (`https://host` and `https://host:443`
+/// share a key). Hostless URLs (`file:`, `data:`) keep the rest of the URL
+/// so they do not inherit each other's hello.
+pub(crate) fn origin_key(url: &Url) -> String {
+    match (
+        url.host_str().filter(|host| !host.is_empty()),
+        url.port_or_known_default(),
+    ) {
         (Some(host), Some(port)) => format!("{}://{host}:{port}", url.scheme()),
         (Some(host), None) => format!("{}://{host}", url.scheme()),
-        (None, _) => format!("{}:", url.scheme()),
+        (None, _) => match url.as_str().split_once('#') {
+            Some((without_fragment, _)) => without_fragment.to_owned(),
+            None => url.as_str().to_owned(),
+        },
+    }
+}
+
+/// Origin keys `url` can match, including an `http` → `https` upgrade.
+///
+/// HSTS (and similar) hops store the hello under `https://host` while
+/// `navigate` still asked for `http://host`. Same host and port, new scheme.
+fn origin_keys(url: &Url) -> Vec<String> {
+    let key = origin_key(url);
+    if url.scheme() != "http" {
+        return vec![key];
+    }
+    let mut https = url.clone();
+    if https.set_scheme("https").is_err() {
+        return vec![key];
+    }
+    let https_key = origin_key(&https);
+    if https_key == key {
+        vec![key]
+    } else {
+        vec![key, https_key]
     }
 }
 
@@ -66,7 +96,7 @@ impl EvalEngine {
         }
     }
 
-    /// Record the hello the bridge sends from `page` (its `location.href`).
+    /// Record a hello from `page` (the invoking webview's URL).
     ///
     /// Only pages whose origin may call `__callback` can say hello, so the
     /// hellos tell which origins answer bridge commands (#153).
@@ -91,9 +121,16 @@ impl EvalEngine {
     ///
     /// Also `true` before any hello: without one the engine cannot tell app
     /// origins from foreign ones, so callers keep the plain eval path.
+    /// An `http` URL also matches a hello from the `https` upgrade of the
+    /// same host and port, so an HSTS hop still counts.
     pub fn has_bridge(&self, url: &Url) -> bool {
         let bridges = self.bridges.borrow();
-        bridges.hellos == 0 || bridges.latest.contains_key(&origin_key(url))
+        if bridges.hellos == 0 {
+            return true;
+        }
+        origin_keys(url)
+            .iter()
+            .any(|key| bridges.latest.contains_key(key))
     }
 
     /// Origins whose bridge said hello, sorted.
@@ -105,11 +142,16 @@ impl EvalEngine {
 
     /// Wait for a hello from the origin of `url` newer than hello number `since`.
     ///
-    /// Returns `false` when none arrives within `limit`.
+    /// Returns `false` when none arrives within `limit`. An `http` destination
+    /// also succeeds when the hello comes from the `https` upgrade of the
+    /// same host and port.
     pub async fn wait_bridge(&self, url: &Url, since: u64, limit: Duration) -> bool {
-        let key = origin_key(url);
+        let keys = origin_keys(url);
         let mut rx = self.bridges.subscribe();
-        let hello = rx.wait_for(|b| b.latest.get(&key).is_some_and(|&n| n > since));
+        let hello = rx.wait_for(move |b| {
+            keys.iter()
+                .any(|key| b.latest.get(key).is_some_and(|&n| n > since))
+        });
         tokio::time::timeout(limit, hello)
             .await
             .is_ok_and(|seen| seen.is_ok())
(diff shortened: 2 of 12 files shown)


crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -137,6 +137,8 @@
   window.fetch = function(input, init) {
     const method = (init && init.method) || (input && input.method) || "GET";
     const url = (typeof input === "string") ? input : (input && input.url) || String(input);
+    // Pilot's own IPC (eval callbacks, the bridge hello) is not app traffic.
+    if (url.indexOf("plugin%3Apilot%7C") !== -1) return _originalFetch(input, init);
```
mpiton: > [!WARNING]
> **🟠 High · 🧪 Tests** — No test that plugin IPC is omitted from network.getRequests

The PR and changelog claim eval callbacks and the bridge hello no longer appear in `network.getRequests`, but nothing runs the new `plugin%3Apilot%7C` skip. Existing `bridge.*.test.mjs` files stub `fetch` only so the IIFE loads, and `lib.rs` string-checks the hello invoke, not this filter. Deleting the skip or recording before it would still be green.

**Fix:** Add a Node test beside the other bridge suites (`crates/tauri-plugin-pilot/js/bridge.network.test.mjs`) that loads real `bridge.js`, stubs `window.fetch` to resolve, issues one `plugin%3Apilot%7C__callback` URL and one app URL, and asserts `networkRequests()` contains only the app request.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -580,26 +562,160 @@ async fn handle_eval_method(
...
+        engine.resolve(id, Err("page has no pilot bridge".to_owned()));
     }
 
+    let limit = if engine.has_bridge(&dest) {
+        DEFAULT_TIMEOUT
+    } else {
+        BRIDGE_GRACE
+    };
+    if engine.wait_bridge(&dest, since, limit).await {
+        Ok(serde_json::json!({"ok": true}))
```
mpiton, line 676: > [!WARNING]
> **🟠 High · 🧪 Tests** — No test that navigate succeeds when a new origin hellos in time

`handle_navigate` returns `{ok: true}` when `wait_bridge` succeeds for a destination that has never said hello, using `BRIDGE_GRACE`. The new tests cover no hello (error), same-origin (dest already known), and navigating back to an origin that already hellod (`DEFAULT_TIMEOUT`). A first visit to a `remote.urls` origin that hellos within 3s is never dispatched, so that success path can break while every new test stays green.

**Fix:** In the `handler.rs` tests, start from `engine_with_app_bridge`, navigate to a new origin, have `eval_fn` resolve `{ok: true}` from the app page then `handle_callback(0, dest, None)` after a short paused sleep, and assert the dispatch returns `{ok: true}` before `DEFAULT_TIMEOUT`.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.86</sub>

docs/src/content/docs/reference/cli.md
```diff
@@ -914,6 +914,12 @@ tauri-pilot navigate "http://localhost:1420/settings"
 tauri-pilot navigate "/"
 ```
 
+Bridge commands only work on origins allowed to call the plugin back: the app
+origin, plus any origin a capability lists in `remote.urls`. When the
+destination is another origin, `navigate` still loads it but fails after 3
+seconds, and later commands fail at once with an error that names the page.
+Run `tauri-pilot navigate` with an app URL to get the session back.
```
mpiton: > [!WARNING]
> **🟠 High · 📄 Docs** — CLI navigate docs say another origin always fails after 3s

The new `navigate` paragraph first says origins in `remote.urls` can call back, then says a destination that is another origin always fails after 3 seconds and later commands fail at once. `handle_navigate` returns ok when `has_bridge(&dest)` is already true, and `wait_bridge` can succeed within `BRIDGE_GRACE` on a first visit; the `BRIDGE_GRACE` comment also states a slow allowed page still answers later commands once its hello lands.

**Fix:** Describe the actual contract: `navigate` still loads any URL; it waits up to 3s only when the destination origin has never said hello, and fails only if none arrives. Spell out that a `remote.urls` origin can succeed (or miss the 3s window on first visit, with later commands recovering), and that later fail-fast applies to origins that cannot call back.

```suggestion
Bridge commands only work on origins allowed to call the plugin back: the app
origin, plus any origin listed in a capability's `remote.urls` that also
grants `pilot:default`. `navigate` still loads any URL. If the destination
origin has never said hello, it waits up to 3 seconds for one and fails
without it. A slow page allowed by `remote.urls` can miss that window on
its first visit; later commands succeed once the hello arrives. On an
origin that cannot call back, later commands fail at once with an error
that names the page and the origins that work. Run `tauri-pilot navigate`
with an app URL to get the session back.
```

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.92</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -635,6 +751,12 @@ pub(crate) fn handle_callback(
     result: Option<String>,
     error: Option<String>,
 ) {
+    if id == HELLO_ID {
+        if let Some(page) = result {
+            engine.bridge_hello(&page);
+        }
+        return;
```
mpiton, line 825: > [!IMPORTANT]
> **🟡 Medium · 🔒 Security** — Bridge hello trusts client-supplied URL instead of the webview origin

`HELLO_ID` records the IPC `result` string as the page URL. Any origin allowed to call `__callback` (the app origin, or one listed in `remote.urls`) can invoke `plugin:pilot|__callback` with a spoofed href and insert origins it does not own into `bridges.latest`. `has_bridge` / `wait_bridge` then treat those origins as able to answer, so `navigate` can report ok without a hello from the destination, later commands hang for 10s instead of failing fast, and the origin map grows without bound.

**Fix:** Have `callback` and `__callback` take the invoking `WebviewWindow`, read its URL with the same `catch_unwind` path as `current_url` in `lib.rs`, and pass that into `bridge_hello`. Ignore client `result` when `id == HELLO_ID`.

<details>
<summary>🔎 Details</summary>

- **References:** CWE-345, CWE-346

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.84</sub>

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -3,7 +3,33 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+    hellos: u64,
+    latest: HashMap<String, u64>,
+}
+
+/// Key a URL by scheme, host and port.
+///
+/// `Url::origin` is opaque for custom schemes such as `tauri://`, which would
+/// make every app page its own origin, so the key is built by hand.
+fn origin_key(url: &Url) -> String {
+    match (url.host_str(), url.port()) {
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — origin_key treats implicit and explicit default ports as different origins

`Url::port()` is None when the port is the scheme default, so `https://example.com/` keys as `https://example.com` while `https://example.com:443/` keys as `https://example.com:443`. Hello uses `location.href` (browsers strip :80/:443); `navigate` dest comes from `page.join(url)` and keeps an explicit default port. `has_bridge`/`wait_bridge` then miss the hello: a same-origin or `remote.urls` navigate to `https://host:443/...` waits out the 3s grace and fails even though the page can answer.

**Fix:** Key with `url.port_or_known_default()` so implicit and explicit default ports share a key (`https://example.com` and `https://example.com:443` both become `https://example.com:443`). Custom schemes such as `tauri://` still have no default and stay unchanged.

```suggestion
    match (url.host_str(), url.port_or_known_default()) {
```

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.93</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -580,26 +562,160 @@ async fn handle_eval_method(
...
+    let (id, rx, page) = send_script(&script, engine, eval_fn, window)?;
+    let dest = page
+        .as_ref()
+        .zip(params.and_then(|p| p.get("url")?.as_str()))
+        .and_then(|(page, url)| page.join(url).ok())
+        // A `javascript:` URL runs in the current page instead of leaving it.
+        .filter(|dest| dest.scheme() != "javascript");
+    let (Some(page), Some(dest)) = (page, dest) else {
+        return wait(engine, id, rx, DEFAULT_TIMEOUT).await;
+    };
```
> **🟡 Medium · 🐛 Correctness** — handle_navigate hangs 10s when dest cannot be resolved

`dest` is `page.as_ref().zip(url).and_then(join)`, then `javascript:` is dropped, and any `None` skips the new #153 logic for a 10s `wait`. `current_url` already returns `None` when `wv.url()` panics or errors — the reason this PR wrapped it in `catch_unwind` — so an absolute `https://…` target is discarded even though it does not need a base. After the app origin has said hello, a foreign page cannot callback, so `javascript:` navigate (CLI does not filter it) and any navigate issued while `url()` is unavailable hang for `DEFAULT_TIMEOUT` instead of fail-fast / `wait_bridge`.

**Fix:** Parse the `url` param with `tauri::Url::parse` first and only `page.join` for relative input so absolute destinations survive a missing page URL. Map `javascript:` to `page.clone()` (same document) instead of filtering dest to `None`. If `page` is known and `!has_bridge(page)` and dest is still `None`, resolve the pending eval and return `no_bridge_error` as `eval_bridge` does; if dest is `Some` and page is `None`, still `wait_bridge` for dest.

<sub>🤖 <b>mpiton-review</b> · lang-rust · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -580,26 +562,160 @@ async fn handle_eval_method(
...
+    } else {
+        BRIDGE_GRACE
+    };
+    if engine.wait_bridge(&dest, since, limit).await {
+        Ok(serde_json::json!({"ok": true}))
+    } else {
+        Err(no_bridge_error(
+            engine,
+            &format!("navigated to {dest}, but no pilot bridge answered there within {limit:?}"),
+        ))
```
> **🟡 Medium · 🐛 Correctness** — navigate waits on the requested origin, so a redirect looks like no bridge

`wait_bridge(&dest, since, limit)` and the earlier `has_bridge(&dest)` use the origin of the pre-navigation URL from `page.join(url)`. Hellos are keyed from the loaded document's `location.href`. An HSTS or www hop (`http://allowed.example` → `https://allowed.example`, both in `remote.urls`) therefore stores the hello under a different `origin_key`. `navigate` fails after `BRIDGE_GRACE` even though the final origin can answer; the next command succeeds because `eval_bridge` keys off `current_url`.

**Fix:** Succeed `wait_bridge` when a hello newer than `since` arrives from dest or from dest with the scheme upgraded to `https` (and the same host/port). On timeout, re-check `has_bridge` for that https variant — or re-read the webview URL and test `has_bridge` on the actual origin — before returning `no_bridge_error`.

<sub>🤖 <b>mpiton-review</b> · lang-rust · confidence 0.82</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -1758,4 +1880,143 @@ mod tests {
...
+            &Recorder::new(),
+        )
+        .await
+        .expect_err("navigate to a foreign origin must not report ok");
+
+        assert!(
+            start.elapsed() < DEFAULT_TIMEOUT,
+            "took {:?}",
+            start.elapsed()
+        );
```
mpiton, line 2021: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Foreign-origin navigate test passes if navigate fails immediately

`test_dispatch_navigate_to_foreign_origin_does_not_report_ok` only checks that the call errors and that elapsed time is under 10s. If `handle_navigate` returned `Err` as soon as dest had no hello, skipping `wait_bridge` and `BRIDGE_GRACE`, elapsed would be ~0 and this test would still pass. The 3s grace on the fail path is therefore unpinned.

**Fix:** Under `start_paused`, assert `start.elapsed()` is `BRIDGE_GRACE` (or at least `BRIDGE_GRACE` and still less than `DEFAULT_TIMEOUT`) so an immediate error fails the test.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.92</sub>

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -35,9 +62,59 @@ impl EvalEngine {
...
+        origins
+    }
+
+    /// Wait for a hello from the origin of `url` newer than hello number `since`.
+    ///
+    /// Returns `false` when none arrives within `limit`.
+    pub async fn wait_bridge(&self, url: &Url, since: u64, limit: Duration) -> bool {
+        let key = origin_key(url);
+        let mut rx = self.bridges.subscribe();
+        let hello = rx.wait_for(|b| b.latest.get(&key).is_some_and(|&n| n > since));
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — wait_bridge is not tested for hellos from a different origin

`wait_bridge` is supposed to fire only when `latest[origin_key] > since`. `test_wait_bridge_ignores_hellos_before_since` uses a single origin, and `test_has_bridge_keys_by_scheme_host_and_port` never calls `wait_bridge`. A predicate that only checked `hellos > since` would still pass both, so a later hello from the app origin could make `navigate` to a foreign dest report ok.

**Fix:** Add a paused test: after a `tauri://localhost` hello, take `since = hellos()`, send another hello from that app origin, and assert `wait_bridge(https://example.com, since, 1s)` is false.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.84</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -580,26 +562,160 @@ async fn handle_eval_method(
...
+        data: None,
+    })?;
+    let since = engine.hellos();
+    let (id, rx, page) = send_script(&script, engine, eval_fn, window)?;
+    let dest = page
+        .as_ref()
+        .zip(params.and_then(|p| p.get("url")?.as_str()))
+        .and_then(|(page, url)| page.join(url).ok())
+        // A `javascript:` URL runs in the current page instead of leaving it.
+        .filter(|dest| dest.scheme() != "javascript");
```
> **🟡 Medium · 🧪 Tests** — javascript: navigate destination filter is untested

`handle_navigate` drops a `javascript:` dest and waits on the current page's callback, because that URL never leaves the page. CLI `navigate` still forwards those URLs (only MCP rejects them). No test hits the filter, so treating `javascript:` as a foreign origin and waiting out `BRIDGE_GRACE` would not fail any test.

**Fix:** Dispatch `navigate` with `url: "javascript:void(0)"` from `engine_with_app_bridge`, resolve the eval callback with `{ok: true}`, and assert success without waiting `BRIDGE_GRACE`.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.78</sub>

docs/src/content/docs/guides/architecture.md
```diff
@@ -103,6 +103,12 @@ The eval function dynamically resolves the target window on each call — it cap
...
 
+### Origins without a callback
+
+The bridge is injected into every page, but Tauri's ACL lets `__callback` through only from origins that hold the pilot permission: the app origin, plus any origin a capability lists in `remote.urls`. On any other origin the script runs and its result is dropped, and a denied call sends the plugin no signal. To find out which origins can answer, the bridge sends a hello on each page load: `__callback` with id `0`, which no eval request uses, and its `location.href` as the result. The `EvalEngine` records the origin (scheme, host and port) of each hello.
+
+The eval function returns the URL of the page it sent the script to. When that page's origin never said hello, the command fails at once instead of waiting 10 seconds, and the error names the origins that did. `navigate` goes one step further. The old page answers before the webview leaves it, so for a destination origin with no hello yet, `navigate` waits up to 3 seconds for a hello from the new page and fails without one. Until the first hello arrives, the plugin cannot tell origins apart and every command takes the plain path.
```
mpiton, line 110: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — Architecture omits known limits of origin hellos and the 3s grace

The new section states that `navigate` waits up to 3 seconds for a first hello and fails without one, with no further caveat. `BRIDGE_GRACE`'s comment and the PR both call out that a slow remote page allowed by `remote.urls` can miss that window on its first visit while later commands succeed once the hello lands, and that hellos are tracked per origin, not per window. Operators will treat a failed first `navigate` to an allowed origin as a bad capability.

**Fix:** After that paragraph, add the two known limits: hellos are per origin, not per window; a slow `remote.urls` page can miss the 3s grace on its first visit, and later commands succeed once its hello arrives.

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.84</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #170

> [!CAUTION]
> 🚫 **Changes requested** — 3 blocking issues (0 critical, 3 high) out of 12 findings.

**Actionable comments posted: 11**

|:---|---:|:---|
| 🟠 High | 3 | must fix before merge |
| 🟡 Medium | 9 | fix now or file a follow-up |
| 🔵 Low | 0 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/handler.rs` | 0 | 1 | 5 | 0 |
| `crates/tauri-plugin-pilot/src/eval.rs` | 0 | 0 | 2 | 0 |
| `crates/tauri-plugin-pilot/js/bridge.js` | 0 | 1 | 0 | 0 |
| `docs/src/content/docs/reference/cli.md` | 0 | 1 | 0 | 0 |
| `docs/src/content/docs/guides/architecture.md` | 0 | 0 | 1 | 0 |
| `docs/src/content/docs/guides/plugin-setup.md` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟠 **No test that plugin IPC is omitted from network.getRequests** — `crates/tauri-plugin-pilot/js/bridge.js:141`
2. 🟠 **No test that navigate succeeds when a new origin hellos in time** — `crates/tauri-plugin-pilot/src/handler.rs:628`
3. 🟠 **CLI navigate docs say another origin always fails after 3s** — `docs/src/content/docs/reference/cli.md:921`
4. 🟡 **Bridge hello trusts client-supplied URL instead of the webview origin** — `crates/tauri-plugin-pilot/src/handler.rs:758`
5. 🟡 **origin_key treats implicit and explicit default ports as different origins** — `crates/tauri-plugin-pilot/src/eval.rs:27`

<details>

- 🟡 **Setup guide never shows how to allow a remote origin** — `docs/src/content/docs/guides/plugin-setup.md:67` — `no_bridge_error` tells the user to allow the origin in a capability's `remote.urls`, and the CLI/architecture pages repeat that key. The permissions snippet in `plugin-setup.md` (copied in `getting-started.md` and the README) still only has `permissions: ["core:default", "pilot:default"]`, so a reader following setup cannot act on the new error. Listing `remote.urls` on a capability that does not
  - *Fix:* Extend the existing capability example with a `remote.urls` array on the same capability that already includes `pilot:default`, and add one sentence that this is required to drive a foreign origin. Mirror that snippet in `docs/src/content/docs/getting-started.md` and the README capability callout.
  - *Link to this PR:* this PR's fail-fast error and CLI/architecture text advertise `remote.urls` as the recovery path but leave the permissions guide unchanged

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `5330a60e5e` |
| Base | `main` |
| Files changed | 8 (+487/-48) |
| Languages | rust, javascript |
| Reviewers | documentation, lang-javascript, lang-rust, maintainability, product-logic, quality, security, tests |

</details>

crates/tauri-plugin-pilot/js/bridge.js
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Restrict the exclusion to actual Pilot IPC requests.

Fixed in `82f8b5d`. `isPilotIpcUrl` only looks at the path, so a query that contains `plugin%3Apilot%7C` is still recorded. `bridge.network.test.mjs` covers both the IPC skip and that query case.

docs/src/content/docs/reference/cli.md
```diff
@@ -914,6 +914,12 @@ tauri-pilot navigate "http://localhost:1420/settings"
 tauri-pilot navigate "/"
 ```
 
+Bridge commands only work on origins allowed to call the plugin back: the app
+origin, plus any origin a capability lists in `remote.urls`. When the
+destination is another origin, `navigate` still loads it but fails after 3
+seconds, and later commands fail at once with an error that names the page.
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Qualify the three-second navigation timeout.

Fixed in `82f8b5d`. CLI and changelog now say the 3s grace is only for a dest that has never said hello. A known origin still uses the 10s wait for the new page.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -177,13 +177,24 @@ fn make_eval_fn<R: tauri::Runtime>(app: &tauri::AppHandle<R>) -> EvalFn {
                 .cloned()
                 .ok_or_else(|| "No webview available".to_owned())?
         };
+        // Read before the eval so the URL names the page the script lands on.
+        let page = current_url(&target);
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Concurrent commands can navigate the selected window after `current_url` returns but before `target.eval` runs.

That's a real TOCTOU, but this PR is fail-fast for a single CLI connection. Serializing eval vs navigate per window is a bigger change than the origin check, so I'm leaving it out of this one.

crates/tauri-plugin-pilot/src/server/mod.rs
```diff
@@ -9,7 +9,10 @@ use tokio::io::{AsyncBufReadExt, AsyncReadExt, AsyncWriteExt, BufReader};
 
 /// A function that evaluates JS in the webview.
 /// The first argument is an optional window label (`None` means "use default window").
-pub(crate) type EvalFn = Arc<dyn Fn(Option<&str>, String) -> Result<(), String> + Send + Sync>;
+/// Returns the URL of the page the script was sent to, or `None` when the
+/// webview cannot report it.
+pub(crate) type EvalFn =
+    Arc<dyn Fn(Option<&str>, String) -> Result<Option<tauri::Url>, String> + Send + Sync>;
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When the webview cannot report its URL, `handle_navigate` takes the `None` fallback and returns the old page's callback as success.

Fixed in `82f8b5d` for the case that matters: an absolute dest is parsed without a page URL and we `wait_bridge` on it. If both page and dest are unknown we still wait on the eval callback, because there is nothing else to check.

crates/tauri-plugin-pilot/src/handler.rs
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When `navigate` receives a `javascript:` URL on a foreign page, this branch waits for the denied callback for 10 seconds.

Fixed in `82f8b5d`. `javascript:` dest is the current page. Same-origin dest with no bridge fail-fasts the same way `eval_bridge` does.

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -3,7 +3,33 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+
+/// Key a URL by scheme, host and port.
+///
+/// `Url::origin` is opaque for custom schemes such as `tauri://`, which would
+/// make every app page its own origin, so the key is built by hand.
+fn origin_key(url: &Url) -> String {
+    match (url.host_str(), url.port()) {
+        (Some(host), Some(port)) => format!("{}://{host}:{port}", url.scheme()),
+        (Some(host), None) => format!("{}://{host}", url.scheme()),
+        (None, _) => format!("{}:", url.scheme()),
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Hostless URLs are collapsed to one origin key, so separate opaque pages such as `file:` pages can inherit each other’s bridge status.

Fixed in `82f8b5d`. Hostless keys are the URL minus fragment, so `file:///tmp/a.html` and `file:///tmp/b.html` don't share a hello.

crates/tauri-plugin-pilot/src/eval.rs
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Navigating to a URL with an explicit default port can report that no bridge answered.

Fixed in `82f8b5d`. `origin_key` uses `port_or_known_default`, so `https://host` and `https://host:443` are the same key.

crates/tauri-plugin-pilot/js/bridge.js
mpiton: > No test that plugin IPC is omitted from network.getRequests

Fixed in `82f8b5d`. `bridge.network.test.mjs` loads real `bridge.js`, hits an IPC path and an app URL, and asserts only the app request is stored. A second case keeps a query that contains `plugin%3Apilot%7C`.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > No test that navigate succeeds when a new origin hellos in time

Fixed in `82f8b5d`. `test_dispatch_navigate_to_new_origin_succeeds_when_hello_arrives` navigates to a fresh origin, sends that dest's hello after a paused sleep, and asserts `{ok: true}` before `DEFAULT_TIMEOUT`.

docs/src/content/docs/reference/cli.md
mpiton: > CLI navigate docs say another origin always fails after 3s

Fixed in `82f8b5d`. The CLI paragraph now matches the real contract: 3s only on a first hello, `remote.urls` can succeed or miss that window, later fail-fast is for origins that cannot call back.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > Bridge hello trusts client-supplied URL instead of the webview origin

Fixed in `82f8b5d`. `callback` / `__callback` take the invoking `WebviewWindow`, hello uses `current_url`, and the IPC `result` is ignored. `test_hello_uses_webview_url_not_client_payload` covers a spoofed href.

crates/tauri-plugin-pilot/src/eval.rs
mpiton: > origin_key treats implicit and explicit default ports as different origins

Fixed in `82f8b5d`. Keys go through `port_or_known_default`, so `https://example.com/` and `https://example.com:443/` match.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > handle_navigate hangs 10s when dest cannot be resolved

Fixed in `82f8b5d`. Absolute URLs parse without a page base. `javascript:` dest is the current page. Known page with no bridge fail-fasts when dest is missing or same-origin. Known dest with no page URL still `wait_bridge`s.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > navigate waits on the requested origin, so a redirect looks like no bridge

Fixed in `82f8b5d`. `has_bridge` / `wait_bridge` also match dest with the scheme upgraded to `https` (same host/port), so an HSTS hop counts.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > Foreign-origin navigate test passes if navigate fails immediately

Fixed in `82f8b5d`. That test now asserts elapsed is at least `BRIDGE_GRACE` and still under `DEFAULT_TIMEOUT`.

crates/tauri-plugin-pilot/src/eval.rs
mpiton: > wait_bridge is not tested for hellos from a different origin

Fixed in `82f8b5d`. `test_wait_bridge_ignores_hello_from_another_origin` hellos from `tauri://localhost` after `since` and asserts `wait_bridge(https://example.com)` is false.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > javascript: navigate destination filter is untested

Fixed in `82f8b5d`. `test_dispatch_navigate_javascript_url_stays_on_page` dispatches `javascript:void(0)`, gets `{ok: true}` from the current page, and finishes under `BRIDGE_GRACE`.

docs/src/content/docs/guides/architecture.md
mpiton: > Architecture omits known limits of origin hellos and the 3s grace

Fixed in `82f8b5d`. Architecture now states hellos are per origin, not per window, and that a slow `remote.urls` page can miss the 3s grace on first visit.

mpiton:
Addressed the review on `82f8b5d`.

The setup-guide finding that wasn't inlined: `plugin-setup.md`, `getting-started.md`, and the README capability snippet now include `remote.urls` on the same capability as `pilot:default`, plus a line that this is required to drive a foreign origin.

One cubic note I didn't change: serializing eval vs navigate per window. Real TOCTOU, but it's a bigger change than this fail-fast PR and the CLI is single-connection. Ready for re-review.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -580,26 +562,207 @@ async fn handle_eval_method(
...
+        (Some(page), Some(dest)) if origin_key(page) == origin_key(dest) => {
+            if engine.has_bridge(page) {
+                wait(engine, id, rx, DEFAULT_TIMEOUT).await
+            } else {
+                Err(fail_no_bridge(engine, id, page))
+            }
+        }
+        (Some(page), Some(dest)) => {
+            if engine.has_bridge(page) {
+                let result = wait(engine, id, rx, DEFAULT_TIMEOUT).await?;
```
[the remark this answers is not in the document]
mpiton: @coderabbitai

> Continue to `wait_dest_bridge` when the page callback times out

Fixed in `709a70a`. The departing-page wait uses `BRIDGE_GRACE`. Timeout (and a callback from a dest that still has no hello) falls through to `wait_dest_bridge`. Other eval errors still fail the command. `test_dispatch_navigate_succeeds_when_page_callback_times_out_but_dest_hellos` covers it.

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -35,9 +92,71 @@ impl EvalEngine {
...
+    /// An `http` URL also matches a hello from the `https` upgrade of the
+    /// same host and port, so an HSTS hop still counts.
+    pub fn has_bridge(&self, url: &Url) -> bool {
+        let bridges = self.bridges.borrow();
+        if bridges.hellos == 0 {
+            return true;
+        }
+        origin_keys(url)
+            .iter()
+            .any(|key| bridges.latest.contains_key(key))
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When an HTTP destination does not upgrade to HTTPS, this unconditional alias reuses a previous HTTPS hello

Fixed in `709a70a`. `has_bridge` is scheme-specific again. `wait_bridge` still accepts a *fresh* `https` hello after `since`, so a real HSTS hop on this navigation counts and an old https hello does not. `test_wait_bridge_http_dest_matches_fresh_https_hello` pins that.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -634,7 +797,18 @@ pub(crate) fn handle_callback(
     id: u64,
     result: Option<String>,
     error: Option<String>,
+    page: Option<&tauri::Url>,
 ) {
+    if id == HELLO_ID {
+        // The invoking webview's URL is the origin that was allowed to call
+        // `__callback`. The IPC `result` is client-supplied and is ignored.
+        if let Some(page) = page {
+            engine.bridge_hello(page.as_str());
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When an id-0 IPC request is handled after the webview navigates, `current_url` can identify the new page

Fixed in `709a70a`. Hello still records the webview URL, not the payload. If the payload parses as a different origin, we drop the hello instead of tagging the page the webview already left. `test_hello_dropped_when_webview_url_does_not_match_payload` covers it.

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -133,10 +133,27 @@
...
+    try {
+      const base = (window.location && window.location.href) || "http://localhost/";
+      path = new URL(text, base).pathname;
+    } catch (_) {
+      const q = text.indexOf("?");
+      path = q === -1 ? text : text.slice(0, q);
+    }
+    // Tauri encodes the command in the IPC path (`.../plugin%3Apilot%7C__callback`).
+    // A query string that happens to contain the same text is still app traffic.
+    return path.indexOf("plugin%3Apilot%7C") !== -1;
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When an application requests a path such as `/api/plugin%3Apilot%7C__callback`

Fixed in `709a70a`. `isPilotIpcUrl` matches only `/plugin:pilot|callback` and `/plugin:pilot|__callback` after decoding. The `/api/...` path is recorded; `bridge.network.test.mjs` has that case.

mpiton:
@coderabbitai resolve

Follow-up review is in `709a70a`. The two older CodeRabbit threads already had `@coderabbitai resolve` as replies; posting it here because those commands are ignored on review-thread replies.

Issue #160, opened 2026-09-10
storage get exits 0 for a missing key


mpiton:
`storage get` on a key that does not exist prints `(not found)` and exits 0, so a scenario cannot distinguish "missing" from "present but empty" by exit code.

## Reproduce

```
$ tauri-pilot storage clear
✓ ok
$ tauri-pilot storage get does-not-exist
(not found)
$ echo $?
0
```

## Expected

Exit non-zero when the key is absent, the way the `assert` commands and `wait` already do, so `storage get missing-key || echo absent` works in a shell script and a TOML scenario step fails on a missing key.

If a zero exit is deliberate (treating "absent" as a valid answer rather than an error), then it should at least be documented, and `--json` should emit `null` rather than a human string so the two cases are machine-distinguishable.

PR #183, opened 2026-09-12
fix(cli): storage get exits 1 on missing key (#160)

mpiton:
• Exit code 1 distinguishes missing key from present-but-empty
• "(not found)" prints to stderr, leaving stdout for the value
• Regression test: missing key, empty value, present value
• `--json` still emits `{found: false}` on missing key
• JSON-RPC and MCP responses unchanged

---
## Summary by cubic
`storage get` on a missing key now exits 1 instead of 0, so missing keys are no longer confused with present-but-empty values and can be used in shell conditionals. The `(not found)` notice moved to stderr, leaving stdout for the value only.

**Notes**
- Scripts relying on exit 0 for missing keys need to handle exit 1.
- `--json` still prints `{"found": false}` but exits 1; connection and RPC errors exit 1 too.
- The exit is tied to the `storage get` command itself, so `storage set` and `storage clear` still exit 0.
- Regression tests cover missing, empty, and present keys in text and `--json` output.

<sup>Written for commit d9f99ba208d5783bc42507febb4b0ef03b501bca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - `storage get` now exits with status 1 when a key is missing or a connection/RPC error occurs.
  - Missing-key notices and errors are sent to stderr.
  - JSON output reports missing keys as `{"found": false}` while preserving JSON-RPC/MCP behavior.
  - Existing keys, including empty values, continue to exit successfully.

- **Documentation**
  - Updated CLI and command documentation to describe missing-key, JSON, and error output behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `storage get` on a missing key now exits 1, like `assert` and `wait`, so
+  `storage get key || echo absent` works and a missing key is no longer
+  confused with a key holding an empty string. The `(not found)` notice moved
+  to stderr, leaving stdout for the value. `--json` still prints
+  `{"found": false}` and exits 1 too. The JSON-RPC and MCP responses are
+  unchanged. [#160]
+
 - `wait --gone` now returns `{ gone: true }` (`--json`, MCP, JSON-RPC;
   CLI text: `✓ gone`) and times out with
   `Timeout waiting for <selector> to disappear`. Success used to
@@ -782,3 +789,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#156]: https://github.com/mpiton/tauri-pilot/issues/156
 [#157]: https://github.com/mpiton/tauri-pilot/issues/157
 [#159]: https://github.com/mpiton/tauri-pilot/issues/159
+[#160]: https://github.com/mpiton/tauri-pilot/issues/160
--- a/crates/tauri-pilot-cli/src/main.rs
+++ b/crates/tauri-pilot-cli/src/main.rs
@@ -141,8 +141,7 @@ async fn main() -> Result<()> {
     }
 
     format_result(output_kind, &result, args.json)?;
-    // Only `storage get` answers with `found`; a missing key exits 1 (#160).
-    if matches!(output_kind, OutputKind::Storage) && result["found"] == false {
+    if matches!(output_kind, OutputKind::StorageGet) && result["found"] == false {
         std::process::exit(1);
     }
     Ok(())
@@ -156,6 +155,7 @@ enum OutputKind {
     Network,
     Watch,
     Storage,
+    StorageGet,
     Forms,
     Windows,
     Record,
@@ -171,6 +171,10 @@ impl From<&Command> for OutputKind {
             Command::Logs { .. } => OutputKind::Logs,
             Command::Network { .. } => OutputKind::Network,
             Command::Watch { .. } => OutputKind::Watch,
+            Command::Storage(StorageArgs {
+                action: StorageAction::Get { .. },
+                ..
+            }) => OutputKind::StorageGet,
             Command::Storage(..) => OutputKind::Storage,
             Command::Forms(..) => OutputKind::Forms,
             Command::Windows => OutputKind::Windows,
@@ -208,12 +212,11 @@ fn format_result(kind: OutputKind, result: &serde_json::Value, emit_json: bool) 
                 output::format_text(result);
             } else if result.get("entries").is_some() {
                 output::format_storage(result);
-            } else if result.get("found").is_some() {
-                output::format_storage_value(result);
             } else {
                 output::format_text(result);
             }
         }
+        OutputKind::StorageGet => output::format_storage_value(result),
         OutputKind::Forms => output::format_forms(result),
         OutputKind::Windows => output::format_windows(result),
         OutputKind::Record => {
--- a/crates/tauri-pilot-cli/src/main.rs
+++ b/crates/tauri-pilot-cli/src/main.rs
@@ -141,6 +141,10 @@ async fn main() -> Result<()> {
     }
 
     format_result(output_kind, &result, args.json)?;
+    // Only `storage get` answers with `found`; a missing key exits 1 (#160).
+    if matches!(output_kind, OutputKind::Storage) && result["found"] == false {
+        std::process::exit(1);
+    }
     Ok(())
 }
 
--- a/crates/tauri-pilot-cli/tests/storage_get_exit.rs
+++ b/crates/tauri-pilot-cli/tests/storage_get_exit.rs
@@ -1,9 +1,10 @@
 //! Regression test for issue #160: `storage get` on a missing key must exit
-//! non-zero, while a key holding an empty string still exits 0.
+//! non-zero, while a key holding a value (even an empty string) exits 0. The
+//! other `storage` subcommands must keep exiting 0.
 //!
 //! Same harness as `snapshot_save_json.rs`: a one-shot mock JSON-RPC unix
-//! socket server answers the single `storage.get` request, then the binary runs
-//! against that socket via `assert_cmd`.
+//! socket server answers the single request, then the binary runs against that
+//! socket via `assert_cmd`.
 
 #![cfg(unix)]
 
@@ -12,9 +13,19 @@ use std::os::unix::net::UnixListener;
 use std::path::PathBuf;
 use std::process::Output;
 use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::mpsc;
 use std::thread;
+use std::time::Duration;
 
 use assert_cmd::Command;
+
+/// How long to wait for the mock server once the binary has exited.
+///
+/// The binary only exits after reading the reply, so the server is done or a
+/// few instructions away from it; the margin covers slow CI hosts. A binary
+/// that exits before connecting leaves the server blocked on `accept()`, and
+/// this bound turns that into a test failure instead of a hung suite.
+const SERVER_DONE_TIMEOUT: Duration = Duration::from_secs(10);
 
 static SOCK_COUNTER: AtomicUsize = AtomicUsize::new(0);
 
@@ -28,57 +39,59 @@ fn unique_socket_path(tag: &str) -> PathBuf {
     ))
 }
 
-/// Run `storage get <key>` against a mock server answering with `result`.
-fn run_storage_get(tag: &str, result: serde_json::Value, json: bool) -> Output {
-    let socket = unique_socket_path(tag);
+/// Run `storage <args>` against a mock server that expects `method` and
+/// answers with `result`.
+fn run_storage(args: &[&str], method: &'static str, result: serde_json::Value) -> Output {
+    let socket = unique_socket_path("storage");
     let _ = std::fs::remove_file(&socket);
     let listener = UnixListener::bind(&socket).expect("bind mock socket");
-    let handle = thread::spawn(move || {
+    let (done_tx, done_rx) = mpsc::channel();
+    thread::spawn(move || {
         let (stream, _) = listener.accept().expect("accept");
         let mut reader = BufReader::new(stream.try_clone().expect("clone stream"));
         let mut writer = stream;
         let mut line = String::new();
         reader.read_line(&mut line).expect("read line");
         let req: serde_json::Value = serde_json::from_str(line.trim()).expect("parse request");
-        assert_eq!(req["method"], "storage.get");
+        assert_eq!(req["method"], method);
         let resp = serde_json::json!({"jsonrpc": "2.0", "id": req["id"], "result": result});
         let mut bytes = serde_json::to_vec(&resp).expect("serialize");
         bytes.push(b'\n');
         writer.write_all(&bytes).expect("write");
         writer.flush().expect("flush");
+        let _ = done_tx.send(());
     });
 
-    let mut args = vec!["--socket", socket.to_str().expect("socket path is UTF-8")];
-    if json {
-        args.push("--json");
-    }
-    args.extend(["storage", "get", "some-key"]);
     let output = Command::cargo_bin("tauri-pilot")
         .expect("cargo_bin")
-        .args(&args)
+        .args([
+            "--socket",
+            socket.to_str().expect("socket path is UTF-8"),
+            "storage",
+        ])
+        .args(args)
         .output()
         .expect("run tauri-pilot");
 
-    // The binary connected and sent a request unless it failed before that
-    // (e.g. arg-parse error); do not block on accept() in that case.
-    if output.status.code() == Some(2) {
-        let _ = std::fs::remove_file(&socket);
+    // Disconnected means the server panicked (e.g. wrong method); timeout means
+    // the binary never connected.
+    let done = done_rx.recv_timeout(SERVER_DONE_TIMEOUT);
+    let _ = std::fs::remove_file(&socket);
+    if let Err(err) = done {
         panic!(
-            "binary exited before connecting.\n--- stderr ---\n{}",
+            "mock server did not answer {method}: {err}\n--- stderr ---\n{}",
             String::from_utf8_lossy(&output.stderr)
         );
     }
-    handle.join().expect("mock server join");
-    let _ = std::fs::remove_file(&socket);
     output
 }
 
 #[test]
 fn storage_get_missing_key_exits_1_with_empty_stdout() {
-    let output = run_storage_get(
-        "storage-missing",
+    let output = run_storage(
+        &["get", "some-key"],
+        "storage.get",
         serde_json::json!({"found": false}),
-        false,
     );
 
     assert_eq!(output.status.code(), Some(1), "missing key must exit 1");
@@ -92,10 +105,10 @@ fn storage_get_missing_key_exits_1_with_empty_stdout() {
 
 #[test]
 fn storage_get_missing_key_json_exits_1() {
-    let output = run_storage_get(
-        "storage-missing-json",
+    let output = run_storage(
+        &["get", "some-key", "--json"],
+        "storage.get",
         serde_json::json!({"found": false}),
-        true,
     );
 
     assert_eq!(output.status.code(), Some(1), "missing key must exit 1");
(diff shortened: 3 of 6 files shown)


docs/src/content/docs/reference/cli.md
```diff
@@ -1200,6 +1200,11 @@ tauri-pilot storage <subcommand> [OPTIONS]
 $ tauri-pilot storage get "auth_token"
 eyJhbGciOiJIUzI1NiJ9...
 
+# A missing key prints "(not found)" on stderr and exits 1
+$ tauri-pilot storage get "missing" || echo absent
+(not found)
+absent
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — Reference treats exit 1 as "key missing", but connection and RPC errors also exit 1

`main` returns `anyhow::Result`, so a stale socket, a stopped app or an RPC timeout also exits 1 (with `Error: ...` on stderr). A script copying the new `storage get "missing" || echo absent` example prints `absent` when the app is simply down. The reference also never says that `--json` exits 1 on a missing key. That is only in the CHANGELOG, and a JSON consumer would not expect it alongside a structured `{"found": false}`.

**Fix:** State in the example (or in an **Exit codes** table like the one under `assert`) that exit 1 also covers connection and RPC failures, and that stderr `(not found)` vs `Error: ...` tells them apart. Also show the `--json` missing-key case.

```suggestion
# A missing key prints "(not found)" on stderr and exits 1.
# Connection and RPC errors exit 1 too, with "Error: ..." on stderr.
$ tauri-pilot storage get "missing" || echo absent
(not found)
absent

# With --json a missing key prints {"found": false} and still exits 1
$ tauri-pilot storage get "missing" --json
{
  "found": false
}
```

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: documentation, product-logic · confidence 0.85</sub>

crates/tauri-pilot-cli/tests/storage_get_exit.rs
```diff
@@ -0,0 +1,116 @@
...
+    // The binary connected and sent a request unless it failed before that
+    // (e.g. arg-parse error); do not block on accept() in that case.
+    if output.status.code() == Some(2) {
+        let _ = std::fs::remove_file(&socket);
+        panic!(
+            "binary exited before connecting.\n--- stderr ---\n{}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+    }
+    handle.join().expect("mock server join");
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Pre-connect guard only catches exit 2, so any other early exit hangs on handle.join()

When `main` returns `Err`, the process exits 1. I checked this: `--socket /tmp/does-not-exist.sock storage get k` exits 1. A panic exits 101. Neither matches `code() == Some(2)`, so `handle.join()` waits forever on a mock thread that is still blocked in `accept()`. Exit 1 is also what two of these tests expect on success, so the status code can't tell "never connected" from "missing key". `ci.yml` sets no `timeout-minutes`, so a regression shows up as a job that hangs for hours instead of a failed test.

**Fix:** Put a time limit on the wait instead of relying on the exit code. Create a `std::sync::mpsc::channel()` before spawning and call `send(())` from the mock thread after `flush()`. Then replace the `Some(2)` check and `handle.join()` with `done_rx.recv_timeout(std::time::Duration::from_secs(10))`, and panic with the captured stderr on `Err`. If `assert_eq!(req["method"], ...)` fails inside the mock, the channel disconnects, so that failure shows up too.

crates/tauri-pilot-cli/tests/storage_get_exit.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut line = String::new();
+        reader.read_line(&mut line).expect("read line");
+        let req: serde_json::Value = serde_json::from_str(line.trim()).expect("parse request");
+        assert_eq!(req["method"], "storage.get");
+        let resp = serde_json::json!({"jsonrpc": "2.0", "id": req["id"], "result": result});
+        let mut bytes = serde_json::to_vec(&resp).expect("serialize");
+        bytes.push(b'\n');
+        writer.write_all(&bytes).expect("write");
+        writer.flush().expect("flush");
+    });
```
mpiton, line 63: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Third copy of the one-shot mock JSON-RPC socket test harness

`SOCK_COUNTER`, `unique_socket_path` and the thread that accepts, reads one line and replies are copied from `snapshot_save_json.rs` and `rpc_error_data.rs`, as the file header says. The copies have already drifted: the guard against blocking on accept() is `!status.success()` in one file, `!stderr.contains(..)` in another and `status.code() == Some(2)` here. Any fix to the harness now has to be made three times.

**Fix:** Move `unique_socket_path` and a `spawn_one_shot_server(socket: &Path, respond: impl FnOnce(&serde_json::Value) -> serde_json::Value + Send + 'static) -> JoinHandle<()>` into `crates/tauri-pilot-cli/tests/common/mod.rs`, with one shared accept() guard. Add `mod common;` to all three test files and keep only the storage-specific argument building in `run_storage_get`.

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -141,6 +141,10 @@ async fn main() -> Result<()> {
     }
 
     format_result(output_kind, &result, args.json)?;
+    // Only `storage get` answers with `found`; a missing key exits 1 (#160).
+    if matches!(output_kind, OutputKind::Storage) && result["found"] == false {
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — The storage get missing-key rule is re-derived from the response shape

`OutputKind` is built from `&args.command`, so the CLI knows it ran `StorageAction::Get` before the RPC call. The exit still guesses it from the payload (`result["found"] == false`), like the `result.get("found").is_some()` branch in `format_result` and the `found` re-check in `format_storage_value` (output.rs:241). The comment holds only because of the `OutputKind::Storage` guard, since `wait` also answers with `found`. Any future storage response that carries `found` inherits exit 1 silently.

**Fix:** Split an `OutputKind::StorageGet` variant out in `From<&Command>` (`Command::Storage(StorageArgs { action: StorageAction::Get { .. }, .. })`) with its own `format_result` arm calling `output::format_storage_value`, then gate the exit on `matches!(output_kind, OutputKind::StorageGet) && result["found"] == false` and drop the comment.

<sub>🤖 <b>mpiton-review</b> · maintainability · confidence 0.8</sub>

CHANGELOG.md
```diff
@@ -26,6 +26,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `storage get` on a missing key now exits 1, like `assert` and `wait`, so
+  `storage get key || echo absent` works and a missing key is no longer
+  confused with a key holding an empty string. The `(not found)` notice moved
+  to stderr, leaving stdout for the value. `--json` still prints
+  `{"found": false}` and exits 1 too. The JSON-RPC and MCP responses are
+  unchanged. [#160]
```
mpiton, line 34: > [!NOTE]
> **🔵 Low · 🎯 Product** — #160's TOML-scenario half is not addressed: `run` has no storage step

Issue #160 expects two things: `storage get missing-key || echo absent` works in a shell, and 'a TOML scenario step fails on a missing key'. `scenario::dispatch_step` has no storage action, so a storage step bails with `unknown step action`, and `run` never reaches the new exit check. A scenario still cannot fail on a missing key, yet this entry marks #160 as fixed.

**Fix:** Either add a `storage-get` action to `dispatch_step` that calls `storage.get` with the step's `key` and bails when `found` is false, or say in the PR that the scenario part is out of scope and keep #160 open or file a follow-up. Is dropping it deliberate?

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** The PR title and this CHANGELOG entry mark #160 as fixed, and the scenario behaviour is half of #160's Expected section.

</details>

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.65</sub>

crates/tauri-pilot-cli/tests/storage_get_exit.rs
```diff
@@ -0,0 +1,116 @@
...
+fn storage_get_empty_value_exits_0() {
+    let output = run_storage_get(
+        "storage-empty",
+        serde_json::json!({"found": true, "value": ""}),
+        false,
+    );
+
+    assert!(output.status.success(), "present-but-empty key must exit 0");
+    assert_eq!(output.stdout, b"\n");
+}
```
mpiton, line 129: > [!NOTE]
> **🔵 Low · 🧪 Tests** — PR body promises a present-value regression test that the file does not contain

The description lists 'missing key, empty value, present value', but the file only covers missing (text), missing (`--json`) and empty. No test runs a non-empty hit such as `{"found": true, "value": "dark"}` (should print `dark` and exit 0), and none checks that `--json` exits 0 on a hit.

**Fix:** Add `storage_get_present_value_exits_0` via `run_storage_get` with `{"found": true, "value": "dark"}`, asserting exit 0 and stdout `dark\n`, plus a `--json` variant asserting exit 0. Or drop 'present value' from the PR body.

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.8</sub>

crates/tauri-pilot-cli/src/main.rs
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Exit-1 gate is untested for storage set/list/clear, which share OutputKind::Storage

The gate runs for all four `storage` subcommands, but the tests only run `storage get`. Two rewrites would keep all three tests green: `!result["found"].as_bool().unwrap_or(false)`, the pattern this PR just removed from `format_storage_value`, or `result["found"] != true`. With either one, `storage set` (`{ok: true}`), `storage list` and `storage clear` exit 1 when they succeed, and any `set -e` script using them breaks.

**Fix:** In `tests/storage_get_exit.rs`, change the helper to take the subcommand args and the expected RPC method. Add a case that runs `storage set k v` against `{"ok": true}` and one that runs `storage clear` against `{"cleared": true}`, and assert exit 0 in both.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the gate is added by this diff and its only tests exercise `storage get`

</details>

mpiton (commented):
## 🧨 mpiton-review — PR #183

> [!NOTE]
> 💬 **Comments** — No blocking issue. 7 findings to consider.

**Actionable comments posted: 7**

|:---|---:|:---|
| 🔵 Low | 5 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/tests/storage_get_exit.rs` | 0 | 0 | 1 | 2 |
| `crates/tauri-pilot-cli/src/main.rs` | 0 | 0 | 0 | 2 |

### 🎯 Start here

1. 🟡 **Reference treats exit 1 as "key missing", but connection and RPC errors also exit 1** — `docs/src/content/docs/reference/cli.md:1206`
2. 🟡 **Pre-connect guard only catches exit 2, so any other early exit hangs on handle.join()** — `crates/tauri-pilot-cli/tests/storage_get_exit.rs:71`
3. 🔵 **Third copy of the one-shot mock JSON-RPC socket test harness** — `crates/tauri-pilot-cli/tests/storage_get_exit.rs:49`
4. 🔵 **The storage get missing-key rule is re-derived from the response shape** — `crates/tauri-pilot-cli/src/main.rs:145`
5. 🔵 **#160's TOML-scenario half is not addressed: `run` has no storage step** — `CHANGELOG.md:34`

<details>

| | |
|:---|:---|
| Commit reviewed | `dfb06fc6cb` |
| Base | `main` |
| Files changed | 6 (+144/-13) |
| Languages | rust |

</details>

crates/tauri-pilot-cli/tests/storage_get_exit.rs
```diff
@@ -0,0 +1,116 @@
...
+    args.extend(["storage", "get", "some-key"]);
+    let output = Command::cargo_bin("tauri-pilot")
+        .expect("cargo_bin")
+        .args(&args)
+        .output()
+        .expect("run tauri-pilot");
+
+    // The binary connected and sent a request unless it failed before that
+    // (e.g. arg-parse error); do not block on accept() in that case.
+    if output.status.code() == Some(2) {
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> The fail-fast guard only handles exit code 2 (clap parse errors), so any other pre-connect exit [...] leaves the mock server thread blocked in `listener.accept()`

Fixed in d9f99ba, but not with the `Some(0) | Some(1)` check: exit 1 is also what `main` returns when the connect fails, so it can't tell "never connected" from "missing key". The mock thread now signals a channel after `flush()`, and the helper waits with `recv_timeout(SERVER_DONE_TIMEOUT)` (10s) instead of `join()`. A timeout or a disconnected channel panics with the binary's stderr. Checked by pointing `--socket` at a path nobody listens on: the test fails after 10s instead of hanging.

crates/tauri-pilot-cli/tests/storage_get_exit.rs
```diff
@@ -0,0 +1,116 @@
...
+#[test]
+fn storage_get_empty_value_exits_0() {
+    let output = run_storage_get(
+        "storage-empty",
+        serde_json::json!({"found": true, "value": ""}),
+        false,
+    );
+
+    assert!(output.status.success(), "present-but-empty key must exit 0");
+    assert_eq!(output.stdout, b"\n");
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> There is no test for a present non-empty value, which is the primary happy path this change protects

Fixed in d9f99ba. `storage_get_present_value_exits_0` answers `{"found": true, "value": "dark"}` and asserts exit 0, stdout exactly `dark\n`, and no `(not found)` on stderr. `storage_get_present_value_json_exits_0` covers the `--json` hit.

docs/src/content/docs/reference/cli.md
mpiton: > Reference treats exit 1 as "key missing", but connection and RPC errors also exit 1

Fixed in d9f99ba with the suggested block. The example now says connection and RPC errors exit 1 with `Error: ...` on stderr, and shows the `--json` missing-key output (`{"found": false}`, exit 1).

crates/tauri-pilot-cli/tests/storage_get_exit.rs
mpiton: > Pre-connect guard only catches exit 2, so any other early exit hangs on handle.join()

Fixed in d9f99ba as described. The mock thread sends on an `mpsc` channel after `flush()`, and `run_storage` replaces the `Some(2)` check and `join()` with `done_rx.recv_timeout(SERVER_DONE_TIMEOUT)`, panicking with the captured stderr on `Err`. A failed `assert_eq!(req["method"], ...)` in the mock drops the sender, so that shows up as `Disconnected` right away. With `--socket` pointed at a dead path the test now fails after 10s instead of hanging.

crates/tauri-pilot-cli/tests/storage_get_exit.rs
mpiton: > Third copy of the one-shot mock JSON-RPC socket test harness

Worth extracting, but not in this PR. It means rewriting `snapshot_save_json.rs` and `rpc_error_data.rs`, which this fix doesn't touch, and neither of those can hang: `snapshot_save_json.rs:84` bails on `!status.success()` and `rpc_error_data.rs:106` on stderr missing `RPC error`, so both only reach `join()` after the binary got a reply. This copy was the one that could hang, and d9f99ba fixes it with the channel timeout. Moving `unique_socket_path` and a `spawn_one_shot_server` with that timeout into `tests/common/mod.rs` makes a good follow-up.

crates/tauri-pilot-cli/src/main.rs
mpiton: > The storage get missing-key rule is re-derived from the response shape

Fixed in d9f99ba. `From<&Command>` maps `Command::Storage(StorageArgs { action: StorageAction::Get { .. }, .. })` to a new `OutputKind::StorageGet` with its own `format_result` arm calling `output::format_storage_value`. The `found` branch is gone from the `Storage` arm, the exit is gated on `OutputKind::StorageGet`, and the comment is dropped.

CHANGELOG.md
mpiton: > #160's TOML-scenario half is not addressed: `run` has no storage step

Deliberate. A scenario can't fail on a missing key because it can't read storage at all: `dispatch_step` has no storage action, so this is a new step type, not part of the exit-code fix. Filed #184 for a `storage-get` step that bails on `found: false`. The CHANGELOG entry only describes the CLI change, and the PR has no closing keyword, so #160 stays open until closed by hand. #184 tracks the scenario part either way.

crates/tauri-pilot-cli/tests/storage_get_exit.rs
mpiton: > PR body promises a present-value regression test that the file does not contain

Fixed in d9f99ba. Added `storage_get_present_value_exits_0` (`{"found": true, "value": "dark"}`, exit 0, stdout `dark\n`) and `storage_get_present_value_json_exits_0` (exit 0, stdout equals the reply), so the PR body is accurate now.

crates/tauri-pilot-cli/src/main.rs
mpiton: > Exit-1 gate is untested for storage set/list/clear, which share OutputKind::Storage

Fixed in d9f99ba. The helper is now `run_storage(args, method, result)`, and `storage_set_exits_0` / `storage_clear_exits_0` run `storage set some-key dark` against `{"ok": true}` and `storage clear` against `{"cleared": true}`. I checked they catch the rewrite you describe: gating on `OutputKind::Storage | OutputKind::StorageGet` with `result["found"] != true` fails both. The real gate is on `OutputKind::StorageGet` now, so set, list and clear can't reach it.

PR #186, opened 2026-09-12
fix(bridge): snapshot omits value="0" on plain <li> (#162)

mpiton:
snapshot emitted `value="0"` on every plain `<li>` because HTMLLIElement.value reflects the value attribute and returns 0 when it is absent or not an integer.

## Changes

- elementValue() reads an `<li>` value attribute as written; absent or empty gives no value
- snapshot no longer carries a value key for plain list items; `<li value="fr">` shows `fr`
- value command on a plain `<li>` still returns an empty string; `<li value="3">` now returns the string "3" instead of the number 3 that assert value rejected
- Tests: no value key for bare or empty `<li value>`, attribute kept as written when set, value command cases in bridge.value.test.mjs

## Notes

- diff --ref on pre-fix snapshot files reports each affected `<li>` as changed once
- Save reference again with updated bridge code

---
## Summary by cubic
Fixes `snapshot` printing `value="0"` on every plain `<li>` because `HTMLLIElement.value` reflects the missing attribute as 0. `snapshot` and `value` now read the `value` attribute as written.

- `snapshot` omits the `value` key for `<li>` without a `value` attribute; `<li value="3">` still carries `value="3"`.
- `value` returns the attribute string (`"3"`, `"0"`), and an empty string for bare `<li>`.
- `diff --ref` against files saved before this fix reports each plain `<li>` as changed until the reference is saved again.

<sup>Written for commit 29a56f7e93b2c685c573fc264a8007cecb7f1223. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/186?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed snapshots incorrectly reporting `value="0"` for list items without an explicit `value` attribute.
  * Preserved explicit numeric list-item values, including `0`, as strings.
  * Maintained existing value handling for multi-select controls and other elements.

* **Documentation**
  * Added an unreleased changelog entry documenting the fix and related regression coverage.

PR #185, opened 2026-09-12
fix(cli): record stop without start exits 1 and writes no file (#161)

mpiton:
Fixes #161.

- `record stop` with no recording in progress now fails with error code -32602 and message "No recording in progress. Run `record start` first"
- CLI exits 1 without writing output file
- Prevents misleading empty `[]` recordings that hide missing `record start`
- Concurrent stop calls protected by recorder lock—only one can succeed
- Plugin-side check, so JSON-RPC and MCP tools return same error
- Regression test: `test_dispatch_record_stop_without_start_errors`
- Updated CHANGELOG, SKILL.md, and CLI reference

---
## Summary by cubic
`record stop` now fails with `No recording in progress` and exits 1 without writing an output file, instead of silently saving `[]` when `record start` was never called.

**Bug Fixes**
- The check lives in the plugin, so JSON-RPC `record.stop` and MCP `record_stop` return the same error.
- The recorder lock makes concurrent or repeated `record stop` calls safe—only one succeeds.

<sup>Written for commit 276b07a87682468959eaba73c6823d93409bd480. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - `record stop` now reports an error when no recording is active.
  - The command exits with status 1 and does not create an output file in this situation.
  - The same behavior is now consistent across CLI, JSON-RPC, and MCP interfaces.

- **Documentation**
  - Updated command documentation to describe the error message, exit status, and output-file behavior.
  - Added a changelog entry for this correction.

PR #182, opened 2026-09-11
fix(bridge): report wait --gone as gone, not found

mpiton:
## Summary

- `wait --gone` now prints `✓ gone` on success and times out with `Timeout waiting for <selector> to disappear`.
- Appear waits are unchanged (`✓ found` / `Timeout waiting for <selector>`).

## Motivation

Success returned `{ found: true }` (CLI: `✓ found`) and the timeout said `Timeout waiting for <selector>`, both of which describe the opposite of waiting for an element to disappear.

Closes #159

## Changes

- Bridge `waitFor` resolves `{ gone: true }` on the gone path and appends `to disappear` to the timeout.
- CLI `format_text` treats `gone` as a status flag (before `found`) so `{ gone: true }` prints `✓ gone`.
- Node tests for immediate gone, mutation gone, gone timeout, and the appear regressions.
- Changelog updated.

`--json` and MCP now return `{ "gone": true }` instead of `{ "found": true }` for gone success.

## Test Plan

- [x] `cargo test --workspace` passes
- [x] `cargo clippy --workspace -- -D warnings` passes
- [ ] Tested manually with a Tauri app (if applicable)

Also: `node --test crates/tauri-plugin-pilot/js/*.test.mjs` (79 passed).

## Checklist

- [x] No `.unwrap()` outside of tests
- [x] All new files are under 150 lines
- [x] Error handling uses `thiserror` (plugin) or `anyhow` (CLI)
- [x] Commit messages follow conventional commits (`feat:`, `fix:`, `refactor:`, etc.)

## Type

fix

---
## Summary by cubic
Fixes `wait --gone` so success and timeout messages describe waiting for an element to disappear instead of appear, closing #159.

- `wait --gone` now resolves `{ gone: true }` and prints `✓ gone`; timeouts report `Timeout waiting for <selector> to disappear`.
- `--json`, MCP, and JSON-RPC responses change from `{ found: true }` to `{ gone: true }` on gone success — a breaking shape change for consumers parsing those fields.
- Adds behavioral tests for immediate gone, mutation gone, gone timeout, appear via mutation, appear timeout, and the unchanged appear path.

<sup>Written for commit 4e5b8feef2ab1de5b4ff923432442ecbd285826f. Summary will update on new commits.</sup>

[...]

PR #189, opened 2026-09-12
fix(bridge): capture uncaught errors and unhandled rejections (#188)

### LINKED ISSUE #188: `logs --level error` cannot see uncaught errors or unhandled rejections
## Bug Description

The bridge captures console output by wrapping `console.log/warn/error/info`, and nothing else. Uncaught exceptions and unhandled promise rejections never pass through `console.*` — the webview prints those itself — so they never reach the log buffer.

That makes `logs --level error` blind to exactly the failures it is documented to find. From `README.md` ("For AI Agents", step 6):

> 6. `tauri-pilot logs --level error` — check for JS errors

and the command table:

> | `logs` | Capture and display console output |

An app whose script throws on load, or that has a rejected promise, reports `No logs captured`. An agent following the documented workflow concludes the page is healthy.

## Steps to Reproduce

1. Start a Tauri app with the plugin enabled (`devtools-pilot`, debug build).
2. `tauri-pilot eval "setTimeout(() => { throw new TypeError('boom'); }, 0)"`
3. `tauri-pilot logs --level error` → nothing.
4. `tauri-pilot eval "Promise.reject(new RangeError('boom2'))"` → also nothing.
5. For contrast, `tauri-pilot eval "console.error('boom3')"` → captured.

## Expected

Uncaught errors and unhandled rejections show up in `logs --level error`, like they do in devtools.

## Actual

Only explicit `console.*` calls are recorded. The throw and the rejection are invisible to `logs`.

## Notes

Confirmed in a clean `__PILOT__` instance (a same-origin child iframe with no app extensions loaded), so this is not an interaction with page code:

| probe | in `consoleLogs()` |
|---|---|
| `console.log(...)` | yes |
| uncaught `throw` | **no** |
| unhandled rejection | **no** |

The host app's own `window.onerror` hook logged both of the errors pilot missed, at the same moment, so the errors definitely happened.

`grep -n "addEventListener" js/bridge.js` shows the only `error` listener is the XHR one; there is no `window.onerror` / `unhandledrejection` hook.

Environment: plugin + CLI 0.7.3, Windows 11, WebView2.

I have a fix with tests ready and will open a PR referencing this issue.

RAMU324E:
Fixes #188.

## Problem

The bridge builds its log buffer by wrapping `console.log/warn/error/info`, and nothing else. Uncaught exceptions and unhandled promise rejections never pass through `console.*` — the webview prints those itself — so they never enter the buffer.

`README.md` documents `tauri-pilot logs --level error` as step 6 of the agent workflow, "check for JS errors". Today that command cannot see a page that throws on load; it answers `No logs captured`, and an agent following the documented workflow concludes the page is fine.

I confirmed this in a clean `__PILOT__` instance (a same-origin child iframe with no app code loaded), so it is not an interaction with page scripts:

| probe | in `consoleLogs()` before this PR |
|---|---|
| `console.log(...)` | yes |
| uncaught `throw` | **no** |
| unhandled rejection | **no** |

## Change

Add `window` listeners for `error` and `unhandledrejection` that push level-`error` entries through the same buffer. Because they share the buffer they inherit `--level`, `--since-id`, `--since`, `--last`, `MAX_LOGS` trimming and `clear` for free, and the id sequence stays monotonic so `sinceId` polling cannot skip an uncaught error.

- Only script errors are recorded. Failed resource loads (``/`<script>` 404) raise a bare `Event` with no `message`; they do not bubble to `window`, but the guard is there anyway so a synthetic one cannot log an empty entry.
- `source` is the first non-bridge stack frame when the event carries an `Error`, otherwise `filename:lineno:colno`.
- Rejection entries read `Unhandled rejection: <name>: <message>`, with the stack in `source`.
- The listener registration is guarded by `typeof window.addEventListener === 'function'` so the existing dependency-free test mocks keep loading the bridge.

Entry construction moves into a shared `pushLog()`. The console wrapper still calls `extractSource()` from its own frame, since that helper skips a fixed number of stack frames to reach the caller — there is a regression test pinning that.

## Tests

New `js/bridge.errors.test.mjs`, written first (RED → GREEN), following the `bridge.network.test.mjs` pattern of eval-ing the real `bridge.js` into a minimal global mock:

- an uncaught error lands in the buffer at level `error`, with `filename:lineno:colno` as source
- an unhandled rejection lands at level `error`
- a rejection with a non-`Error` reason is still recorded
- a failed resource load is **not** reported as a JS error
- ids stay monotonic across sources, and `sinceId` does not skip an uncaught error
- the bridge still loads when `window` has no `addEventListener`
- console entries still record the calling site (guards the `pushLog` refactor)

```
node --test crates/tauri-plugin-pilot/js/*.test.mjs   # 89 pass, 0 fail
cargo test --workspace                                # 347 pass, 0 fail
cargo clippy --workspace -- -D warnings               # clean
cargo fmt --all -- --check                            # clean
```

Also verified the event shapes against the real runtime rather than only the mock — WebView2 on Windows 11, in a live app: the `error` event carries a string `message` plus `filename`/`lineno`/`colno`, and `event.error` / `event.reason` are real `Error`s with a string `stack`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Fixes #188 so `tauri-pilot logs --level error` captures uncaught errors and unhandled rejections. Previously these bypassed the console wrapper and a page that threw on load reported `No logs captured`; now they enter the same log buffer as console entries and respect `--level`, `--since-id`, `--since`, `--last`, and `clear`.

**What changed**
- Adds `window` `error` and `unhandledrejection` listeners that log at `error` level through a shared `pushLog()`.
- Ignores resource-load failures, which fire an `error` event with no message.
- Picks the source from the first non-bridge stack frame with a `:line:col` location, falling back to `filename:lineno:colno`; the human-readable logs renderer prints `source` when it is set.
- Keeps uncaught errors on the same id sequence as console entries, so `sinceId` polling does not skip them.
- Records unusual rejection reasons legibly without throwing: cross-realm `Error`s, strings, symbols, functions, `NaN`/`Infinity`/`undefined`, circular and null-prototype objects, revoked proxies, and getters that throw; objects that spoof the `Error` tag keep their own fields, and non-string `Error` `name`/`message` values stay intact.
- Adds `bridge.errors.test.mjs` covering all of the above plus monotonic ids, and restores the mocked console and global state after tests.

<sup>Written for commit d343541476d85ffa642517703bc71193c863412f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Uncaught script errors and unhandled promise rejections are now captured in error logs.
  - Error logs can include a readable source location when available.
  - Rejections are clearly labeled as unhandled rejections.
- **Bug Fixes**
  - Error details remain readable for circular or otherwise unusual values.
  - Logging is more resilient when error properties cannot be accessed safely.
  - Error entries now use the most relevant available stack information.
- **Documentation**
  - Updated logs documentation to describe error and rejection capture.

### DIFF
--- a/crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
+++ b/crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
@@ -1,0 +1,133 @@
+// Uncaught errors and unhandled rejections must reach the log buffer (#188).
+//
+// README step 6 of the AI-agent workflow is `tauri-pilot logs --level error`
+// to "check for JS errors", but the bridge only wrapped console.*. A page that
+// throws never calls console.error itself — the browser prints that — so the
+// one command documented for finding JS errors could not see them.
+//
+// bridge.js is an IIFE that attaches its API to `window.__PILOT__`. We load the
+// real file into a minimal global mock so these tests exercise the shipping
+// code, not a re-implementation.
+//
+// Run: node --test crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, join } from "node:path";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const BRIDGE_SRC = readFileSync(join(here, "bridge.js"), "utf8");
+
+const REAL_CONSOLE = {
+  log: console.log.bind(console),
+  warn: console.warn.bind(console),
+  error: console.error.bind(console),
+  info: console.info.bind(console),
+};
+
+// The bridge wraps fetch/XHR on load, so the mock has to carry both even
+// though these tests only care about the error listeners.
+function baseGlobals() {
+  Object.assign(console, REAL_CONSOLE);
+  globalThis.location = { href: "https://app.example/" };
+  globalThis.XMLHttpRequest = function () {};
+  globalThis.XMLHttpRequest.prototype.open = function () {};
+  globalThis.XMLHttpRequest.prototype.send = function () {};
+  globalThis.XMLHttpRequest.prototype.addEventListener = function () {};
+  globalThis.XMLHttpRequest.prototype.removeEventListener = function () {};
+  globalThis.document = { querySelector() { return null; } };
+  return {
+    fetch() { return Promise.resolve({ status: 200, headers: { get() { return "0"; } } }); },
+    location: globalThis.location,
+  };
+}
+
+function loadBridge() {
+  const listeners = Object.create(null);
+  globalThis.window = Object.assign(baseGlobals(), {
+    addEventListener(type, handler) {
+      (listeners[type] || (listeners[type] = [])).push(handler);
+    },
+    dispatch(type, event) {
+      for (const handler of listeners[type] || []) handler(event);
+    },
+  });
+
+  (0, eval)(BRIDGE_SRC);
+  return { pilot: globalThis.window.__PILOT__, win: globalThis.window };
+}
+
+test("an uncaught error lands in the log buffer at level error", () => {
+  const { pilot, win } = loadBridge();
+  win.dispatch("error", {
+    message: "Uncaught TypeError: x is not a function",
+    filename: "https://app.example/main.js",
+    lineno: 42,
+    colno: 7,
+  });
+
+  const errors = pilot.consoleLogs({ level: "error" });
+  assert.equal(errors.length, 1);
+  assert.equal(errors[0].args[0], "Uncaught TypeError: x is not a function");
+  assert.match(errors[0].source, /main\.js:42:7/);
+});
+
+test("an unhandled rejection lands in the log buffer at level error", () => {
+  const { pilot, win } = loadBridge();
+  const reason = new Error("boom");
+  reason.stack = "Error: boom\n    at https://app.example/main.js:9:1";
+  win.dispatch("unhandledrejection", { reason });
+
+  const errors = pilot.consoleLogs({ level: "error" });
+  assert.equal(errors.length, 1);
+  assert.match(errors[0].args[0], /Unhandled rejection: .*boom/);
+});
+
+test("a rejection with a non-Error reason is still recorded", () => {
+  const { pilot, win } = loadBridge();
+  win.dispatch("unhandledrejection", { reason: "plain string" });
+
+  const errors = pilot.consoleLogs({ level: "error" });
+  assert.equal(errors.length, 1);
+  assert.match(errors[0].args[0], /Unhandled rejection: plain string/);
+});
+
+test("a failed resource load is not reported as a JS error", () => {
+  const { pilot, win } = loadBridge();
+  // Resource errors (img/script 404) fire the same event type on window but
+  // carry no `message`; they are not JS errors and would be noise in `logs`.
+  win.dispatch("error", { target: { tagName: "IMG" } });
+
+  assert.deepEqual(pilot.consoleLogs({ level: "error" }), []);
+});
+
+test("uncaught errors share the id sequence with console entries", () => {
+  const { pilot, win } = loadBridge();
+  console.log("first");
+  win.dispatch("error", { message: "Uncaught Error: second", filename: "a.js", lineno: 1, colno: 1 });
+
+  const all = pilot.consoleLogs({});
+  assert.equal(all.length, 2);
+  assert.ok(all[1].id > all[0].id, "ids must stay monotonic across sources");
+  // sinceId polling must not skip the uncaught error
+  assert.equal(pilot.consoleLogs({ sinceId: all[0].id }).length, 1);
+});
+
+test("the bridge still loads when window has no addEventListener", () => {
+  globalThis.window = baseGlobals();
+
+  (0, eval)(BRIDGE_SRC);
+  assert.ok(globalThis.window.__PILOT__, "bridge must not throw without addEventListener");
+});
+
+test("console entries still record the calling site", () => {
+  // pushLog() is shared with the error listeners, but extractSource() skips a
+  // fixed frame count, so the console wrapper must keep calling it directly.
+  const { pilot } = loadBridge();
+  console.log("hello");
+
+  const [entry] = pilot.consoleLogs({});
+  assert.match(entry.source, /bridge\.errors\.test\.mjs/);
+});
--- a/crates/tauri-plugin-pilot/js/bridge.js
+++ b/crates/tauri-plugin-pilot/js/bridge.js
@@ -86,20 +86,75 @@
     info: console.info.bind(console),
   };
 
+  function pushLog(level, args, source) {
+    const entry = {
+      id: ++_logIdCounter,
+      timestamp: Date.now(),
+      level: level,
+      args: args.map(serializeArg),
+      source: source || null,
+    };
+    _logs.push(entry);
+    if (_logs.length > MAX_LOGS) _logs.shift();
+    return entry;
+  }
+
   ['log', 'warn', 'error', 'info'].forEach(level => {
     console[level] = function(...args) {
-      const entry = {
-        id: ++_logIdCounter,
-        timestamp: Date.now(),
-        level: level,
-        args: args.map(serializeArg),
-        source: extractSource(),
-      };
-      _logs.push(entry);
-      if (_logs.length > MAX_LOGS) _logs.shift();
+      // extractSource() must be called from this frame: it skips a fixed
+      // number of stack frames to reach the caller.
+      pushLog(level, args, extractSource());
       _originalConsole[level].apply(console, args);
     };
   });
+
+  function describeReason(reason) {
+    if (reason instanceof Error) {
+      // The stack goes in `source`; keep the message readable in `args`.
+      return reason.name + ': ' + reason.message;
+    }
+    if (typeof reason === 'string') return reason;
+    try {
+      return JSON.stringify(reason);
+    } catch (_) {
+      return String(reason);
+    }
+  }
+
+  function stackSource(error) {
+    if (!error || typeof error.stack !== 'string') return null;
+    const lines = error.stack.split('\n');
+    for (let i = 1; i < lines.length; i++) {
+      const line = lines[i];
+      if (line && line.trim() && !line.includes('__PILOT__')) return line.trim();
+    }
+    return null;
+  }
+
+  // Uncaught errors and unhandled rejections never pass through console.* --
+  // the browser prints those itself. Without these listeners the documented
+  // `logs --level error` workflow cannot see the failures it exists to find.
+  if (typeof window.addEventListener === 'function') {
+    window.addEventListener('error', event => {
+      // Only script errors carry a message. Failed resource loads (<img>,
+      // <script> 404) raise a bare Event that does not bubble to window, but
+      // guard anyway rather than logging an empty entry.
+      if (!event || typeof event.message !== 'string') return;
+      const where = event.filename
+        ? event.filename + ':' + (event.lineno || 0) + ':' + (event.colno || 0)
+        : null;
+      pushLog('error', [event.message], stackSource(event.error) || where);
+    });
+
+    window.addEventListener('unhandledrejection', event => {
+      const reason = event && event.reason;
+      pushLog(
+        'error',
+        ['Unhandled rejection: ' + describeReason(reason)],
+        stackSource(reason)
+      );
+    });
+  }
 
   function consoleLogs(options) {
     let result = _logs.slice();


RAMU324E:
Thanks — all three findings were valid. Verified each against the actual runtime before changing anything, and pushed 6d55115.

**1. `JSON.stringify` returning a non-string** (flagged by both reviewers)

Confirmed:

| reason | `JSON.stringify` | was recorded as |
|---|---|---|
| `undefined`, function, `Symbol` | `undefined` | `Unhandled rejection: undefined` |
| `NaN`, `Infinity` | `"null"` | `Unhandled rejection: null` |

So a bare `Promise.reject()` produced an entry that says nothing. Now falls back to `String()` when the result is not a string.

**2. Cross-realm `Error`** (cubic P2)

Confirmed with `vm.runInNewContext('new TypeError("from another realm")')`:

```
instanceof Error : false
toString tag     : [object Error]
JSON.stringify   : {}
name / message   : TypeError / from another realm
```

An entirely readable error was being recorded as `Unhandled rejection: {}`. This one matters more than it looks for the case that motivated the PR — the app I hit this on runs user scripts inside iframes. Now brand-checked with `Object.prototype.toString`, keeping `instanceof` first.

One consequence worth noting: `String()` itself throws on a null-prototype object, so the fallback needed its own guard. There is a test for that.

**3. Missing coverage of the `event.error` stack branch** (cubic P3)

Correct, every dispatch omitted `error`, so only the `filename:lineno:colno` fallback was exercised even though the stack-first behaviour is what the PR claims. Added a test that dispatches an error event carrying an `Error` and asserts the stack frame wins.

**Tests**: 5 added, 12 in the file. I checked the three bug-fix tests actually fail against the previous commit — cross-realm, non-serializable reasons, and null-prototype all went red, so they are real regression tests rather than tests written around the new code.

```
node --test crates/tauri-plugin-pilot/js/*.test.mjs   # 94 pass, 0 fail
cargo test --workspace                                # 347 pass, 0 fail
cargo clippy --workspace -- -D warnings               # clean
cargo fmt --all -- --check                            # clean
```

#191 has been rebased onto this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

RAMU324E:
Second round, from cubic review on the stacked PR (#191) — two of its findings land in this diff, so the fixes are here.

**a74207e — survive hostile rejection reasons.** A revoked proxy throws on the `instanceof` prototype walk *and* on `Object.prototype.toString` (internal IsArray), and `Symbol.toStringTag` is a plain writable accessor free to throw. Either one threw out of `describeReason` before `pushLog` ran, losing the only record the rejection ever happened — the exact failure this PR exists to prevent. The same applied to reading `.stack`. Every read on the path to a log entry now goes through a guard (`safeTag`, `safeRead`, `safeString`, a try/catch around `instanceof`). Separately, the brand check alone promoted any object merely wearing the Error tag to `undefined: undefined`; the `name: message` shape is now used only when one of them really reads back as a string.

**73ba3ab — test hygiene.** `Object.assign(console, REAL_CONSOLE)` routes through the installed accessor, so after a bridge load it pushed a native function onto that bridge chain rather than restoring a clean console. Now `Object.defineProperty`, plus a globals snapshot restored in `after()`.

Four new tests, all red before the fix: revoked proxy recorded; throwing tag getter recorded; throwing `stack` getter keeps the message with `source === null`; a tag-spoofing object keeps its own fields. 98 JS tests pass on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,136 @@
...
+      // Symbol.toStringTag is writable, so the brand check alone would promote
+      // any object wearing the tag to "undefined: undefined" and throw away
+      // what it actually carried. Take the shortcut only when the fields exist.
+      if (typeof name === 'string' || typeof message === 'string') {
+        const label = typeof name === 'string' ? name : 'Error';
+        // The stack goes in `source`; keep the message readable in `args`.
+        return typeof message === 'string' && message
+          ? label + ': ' + message
+          : label;
+      }
```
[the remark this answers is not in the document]
RAMU324E: Valid — fixed in f397248. Requiring a string was the wrong test. An Error subclass is free to put a number or an object in `message`, and dropping it left the entry as a bare "Error" with the only description of the failure gone.

The string check was there to separate a real Error from an object merely wearing the `Symbol.toStringTag` brand, and for that "is it defined" is the right question: a spoof carries neither field and still falls through to the JSON path. Defined fields now go through `safeString`, so a throwing `toString` cannot take the entry down either.

Test covers `{name: "HttpError", message: 404}` → `HttpError: 404`, a non-string `name`, and a `message` object falling back to the prototype `name`. Red before the fix.

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,76 @@
...
-      if (_logs.length > MAX_LOGS) _logs.shift();
+      // extractSource() must be called from this frame: it skips a fixed
+      // number of stack frames to reach the caller.
+      pushLog(level, args, extractSource());
       _originalConsole[level].apply(console, args);
     };
   });
 
+  function describeReason(reason) {
+    if (reason instanceof Error) {
```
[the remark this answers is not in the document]
RAMU324E: Fixed in 6d55115 — `Object.prototype.toString.call(v) === "[object Error]"` as a realm-independent brand check, then hardened further in a74207e so a hostile reason cannot throw out of it. Test: "a rejection with a cross-realm Error keeps its name and message", built with `runInNewContext` and asserting `foreign instanceof Error === false` as a precondition so the test cannot silently stop testing the cross-realm case.

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -0,0 +1,133 @@
...
+    },
+  });
+
+  (0, eval)(BRIDGE_SRC);
+  return { pilot: globalThis.window.__PILOT__, win: globalThis.window };
+}
+
+test("an uncaught error lands in the log buffer at level error", () => {
+  const { pilot, win } = loadBridge();
+  win.dispatch("error", {
```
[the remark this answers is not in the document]
RAMU324E: Fixed in 6d55115 — "an error event with an Error prefers the stack over filename:lineno". It dispatches an error event carrying an Error whose stack points at `widget.js:12:5` alongside a `filename`/`lineno`/`colno` of `main.js:42:7`, and asserts the source is the stack frame and does not match `main.js`.

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,76 @@
...
+      // The stack goes in `source`; keep the message readable in `args`.
+      return reason.name + ': ' + reason.message;
+    }
+    if (typeof reason === 'string') return reason;
+    try {
+      return JSON.stringify(reason);
```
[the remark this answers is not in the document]
RAMU324E: Fixed in 6d55115 — the JSON result is checked for `typeof === "string"` and falls back to `safeString`. Test walks `undefined`, a named function, a symbol, NaN, Infinity and null, so `Promise.reject()` with no reason records "Unhandled rejection: undefined" as text rather than losing the entry.

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,138 @@
...
+    // Same hostile-reason problem: reading .stack must not throw.
+    if (!error) return null;
+    const stack = safeRead(error, 'stack');
+    if (typeof stack !== 'string') return null;
+    const lines = stack.split('\n');
+    for (let i = 1; i < lines.length; i++) {
+      const line = lines[i];
+      if (line && line.trim() && !line.includes('__PILOT__')) return line.trim();
+    }
+    return null;
```
mpiton, line 193: > [!WARNING]
> **🟠 High · 🐛 Correctness** — stackSource picks frames by index: wrong frame on WebKit, message text on V8

The loop starts at index 1 to jump over V8's `Name: message` header, but JavaScriptCore (WebKitGTK on Linux, WKWebView on macOS) emits no header: line 0 is already the throwing frame (`handler@https://app/widget.js:12:5`). On two of the three Tauri webviews `source` therefore names the caller's frame instead of the throw site, or `null` for a single-frame stack, and for `error` events that wrong frame outranks the accurate `filename:lineno:colno` fallback. On V8 the header is not one line either: a rejection with `new Error("Invalid config:\n  expected foo")` records `source: "expected foo"`. Both reproduced with the real bridge loaded the way `bridge.errors.test.mjs` does. The PR was verified on WebView2 only, the one engine where a fixed skip works.

**Fix:** Select frames by shape, not by count: scan from index 0 and keep the first trimmed line ending in a `:line:col` location, which V8 (`at fn (url:1:2)`) and JSC (`fn@url:1:2`) frames match and header/message lines do not. The replacement below keeps all 17 tests green. Add two cases to `bridge.errors.test.mjs`: a JSC-shaped stack `handler@https://app.example/widget.js:12:5\nglobal code@https://app.example/main.js:40:1` asserting `widget.js:12:5`, and a real `new Error("line1\nline2")` reason asserting `source` ends in `bridge.errors.test.mjs:<line>:<col>)`. Both fail on the PR head.

```suggestion
    const lines = stack.split('\n');
    // V8 opens with "Name: message" (possibly multi-line); JavaScriptCore
    // (WebKitGTK, WKWebView) starts at the throwing frame. Keep the first
    // line that carries a location instead of skipping a fixed count.
    for (let i = 0; i < lines.length; i++) {
      const line = lines[i].trim();
      if (/:\d+:\d+\)?$/.test(line) && !line.includes('__PILOT__')) return line;
    }
    return null;
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** stackSource is added by this diff and is the only source of `source` for both new listeners

</details>

<sub>🤖 <b>mpiton-review</b> · 4 reviewers agree: lang-javascript, product-logic, quality, tests · confidence 1.0</sub>

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -0,0 +1,320 @@
...
+});
+
+test("console entries still record the calling site", () => {
+  // pushLog() is shared with the error listeners, but extractSource() skips a
+  // fixed frame count, so the console wrapper must keep calling it directly.
+  const { pilot } = loadBridge();
+  console.log("hello");
+
+  const [entry] = pilot.consoleLogs({});
+  assert.match(entry.source, /bridge\.errors\.test\.mjs/);
```
mpiton, line 165: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Calling-site test cannot fail: every eval'd bridge frame contains the test file name

V8 names the eval call site inside every frame of eval'd code, so the console wrapper's own frame reads `at console.<computed> [as log] (eval at loadBridge (file:///.../bridge.errors.test.mjs:91:10), <anonymous>:106:7)` and already matches `/bridge\.errors\.test\.mjs/`. Verified by mutation: moving `extractSource()` into `pushLog` (`source: source || extractSource()`, wrapper calls `pushLog(level, args)`), the exact refactor this test says it guards, leaves this test green while `source` records the wrapper frame instead of the caller.

**Fix:** Assert the frame is the caller's own: it ends in `bridge.errors.test.mjs:<line>:<col>)`, whereas an eval'd frame ends in `<anonymous>:<line>:<col>)` and contains `eval at`. Verified: the replacement passes on the PR head and fails on the mutant.

```suggestion
  assert.match(entry.source, /bridge\.errors\.test\.mjs:\d+:\d+\)?$/);
  assert.doesNotMatch(entry.source, /eval at/, "bridge frames name the eval site; source must be the caller's own frame");
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the test is added by this diff

</details>

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -0,0 +1,320 @@
...
+
+test("an unhandled rejection lands in the log buffer at level error", () => {
+  const { pilot, win } = loadBridge();
+  const reason = new Error("boom");
+  reason.stack = "Error: boom\n    at https://app.example/main.js:9:1";
+  win.dispatch("unhandledrejection", { reason });
+
+  const errors = pilot.consoleLogs({ level: "error" });
+  assert.equal(errors.length, 1);
+  assert.match(errors[0].args[0], /Unhandled rejection: .*boom/);
```
mpiton, line 118: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Rejection entries' source is never asserted; dropping stackSource(reason) stays green

The PR states rejections carry the first stack frame in `source`, and this test sets `reason.stack` for that purpose, but only `args[0]` is checked. Verified by mutation: removing the `stackSource(reason)` argument from the `unhandledrejection` listener passes all 17 tests (the only `source` assertion on the rejection path is the `null` case at line 283, which the mutant also satisfies).

**Fix:** Assert the stack frame set on line 113 lands in `source`.

```suggestion
  assert.match(errors[0].args[0], /Unhandled rejection: .*boom/);
  assert.match(errors[0].source, /main\.js:9:1/, "the first stack frame must land in source");
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the test is added by this diff

</details>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,138 @@
...
+  }
```
mpiton, line 194: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — stackSource re-implements extractSource's stack walk with a different start index

`stackSource` and `extractSource` (line 68) both split a stack string, skip a fixed number of leading frames, reject frames containing `__PILOT__` and return the first trimmed line. They differ only in the start index (1 vs 3) and in whether a whitespace-only line counts, so the next change to the frame-filtering rule has to be made twice and will drift silently.

**Fix:** Once `stackSource` selects frames by shape (see the correctness finding on lines 188-193), extract one `firstAppFrame(stack, skipFrames)` holding the scan and the `__PILOT__` filter, and have `extractSource()` call it too: its own `i = 3` skip is the same V8-only assumption and should live in one place.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** stackSource is added by this diff next to the existing extractSource

</details>

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -0,0 +1,320 @@
...
+  }
+});
+
+test("a rejection with a circular object does not throw", () => {
+  const { pilot, win } = loadBridge();
+  const circular = {};
+  circular.self = circular;
+
+  assert.doesNotThrow(() => win.dispatch("unhandledrejection", { reason: circular }));
+  assert.equal(pilot.consoleLogs({ level: "error" }).length, 1);
```
mpiton, line 228: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Hostile-reason tests only count entries; recording 'undefined' passes all four

The circular (228), null-prototype (238), revoked proxy (253) and throwing toStringTag (267) tests assert `doesNotThrow` plus `length === 1` and never read `args[0]`. Verified by mutation: making both `describeReason` fallbacks return `undefined`, which records `Unhandled rejection: undefined` (the exact bug the second commit fixed), keeps all four green.

**Fix:** Pin the recorded text in each: circular and null-prototype record `Unhandled rejection: [object Object]`, the revoked proxy `Unhandled rejection: [unprintable]`, the throwing toStringTag `Unhandled rejection: {}` (JSON.stringify ignores symbol keys, so the getter is never hit). Replace the length check at 228 with the assertion below and mirror it at 238, 253 and 267.

```suggestion
  assert.equal(pilot.consoleLogs({ level: "error" })[0].args[0], "Unhandled rejection: [object Object]");
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the tests are added by this diff

</details>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -86,21 +86,138 @@
...
+  function isError(value) {
+    // instanceof is realm-bound, so an Error thrown from an iframe fails it.
+    // The brand check catches those, and Error subclasses keep the same tag.
+    try {
+      if (value instanceof Error) return true;
+    } catch (_) {
+      // instanceof walks the prototype chain, which a revoked proxy refuses.
+    }
+    return safeTag(value) === '[object Error]';
+  }
```
mpiton, line 150: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Hostile-reason hardening spread over five helpers and six try/catch blocks

`safeTag`, `safeString`, `safeRead` and the try inside `isError` add ~35 lines and six independent try/catch layers so that a revoked proxy, a throwing `Symbol.toStringTag` getter or a null-prototype object still yields a partially readable string; eight of the sixteen new tests exist only to pin those shapes. A rejection reason in a WebView under automation is an `Error`, a string or a plain object, and one try/catch around each listener body already delivers the stated guarantee that the entry is never lost.

**Fix:** Keep the two guards that cover real cases: the `Object.prototype.toString` brand check (cross-realm Errors from an iframe) and the try around `JSON.stringify` (circular reasons such as an axios error payload). Drop `safeTag`, `safeRead` and `safeString`, and wrap each listener body in a single try/catch that pushes `'Unhandled rejection: [unprintable]'` on failure with `pushLog` outside the try. That also removes the revoked-proxy, throwing-tag-getter, null-prototype and toStringTag-spoof tests from `bridge.errors.test.mjs`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the whole helper cluster is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · maintainability · confidence 0.7</sub>

mpiton (changes requested):
## 🧨 mpiton-review — PR #189

> [!CAUTION]
> 🚫 **Changes requested** — 1 blocking issue (0 critical, 1 high) out of 9 findings.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/js/bridge.js` | 0 | 1 | 0 | 2 |
| `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs` | 0 | 0 | 2 | 1 |
| `.github/workflows/ci.yml` | 0 | 0 | 1 | 0 |
| `crates/tauri-pilot-cli/src/output.rs` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟠 **stackSource picks frames by index: wrong frame on WebKit, message text on V8** — `crates/tauri-plugin-pilot/js/bridge.js:193`
2. 🟡 **Calling-site test cannot fail: every eval'd bridge frame contains the test file name** — `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs:165`
3. 🟡 **Rejection entries' source is never asserted; dropping stackSource(reason) stays green** — `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs:118`
4. 🔵 **stackSource re-implements extractSource's stack walk with a different start index** — `crates/tauri-plugin-pilot/js/bridge.js:194`
5. 🔵 **Hostile-reason tests only count entries; recording 'undefined' passes all four** — `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs:228`

<details>
<summary>🧭 Findings outside the diff or over the inline cap (3)</summary>

- 🟡 **CI never runs for this PR: the workflow path filter excludes js and mjs files** — `.github/workflows/ci.yml:13` — The `check` job triggers only on `**/*.rs`, `**/Cargo.toml`, `**/Cargo.lock` and `ci.yml`. This PR changes only `bridge.js` and adds `bridge.errors.test.mjs`, so no job runs: not `cargo test --workspace` (which holds the `BRIDGE_JS` substring assertions in `lib.rs`), and not the new node tests, which no workflow invokes at all. The regression guard for #188 exists only on the author's machine, and
  - *Fix:* Add `'**/*.js'` and `'**/*.mjs'` to both `paths` lists, and add a step running `node --test crates/tauri-plugin-pilot/js/*.test.mjs` so the eleven bridge test files execute in CI.
  - *Link to this PR:* this is the first change whose entire content is JS plus a node test, so the filter silently skips the only verification the fix has
- 🟡 **User-visible change to what `logs` returns ships without a CHANGELOG entry or doc update** — `CHANGELOG.md:27` — Every fix merged ahead of this branch (#187, #186, #185, #183, #182) adds an `[Unreleased] > Fixed` bullet, and this PR changes the `logs` contract for every consumer (CLI, `--follow`, JSON-RPC `console.getLogs`, MCP `logs`): uncaught errors and `Unhandled rejection: …` entries now appear at level `error`, share the 500-entry buffer and the id sequence, yet none of the five commits touches CHANGEL
  - *Fix:* Add a `### Fixed` bullet under `[Unreleased]` ending in `[#188]` stating that uncaught errors and unhandled rejections are now captured at level `error` (with the `Unhandled rejection: ` prefix and the `source` fallback), and extend the `logs` wording in `docs/src/content/docs/reference/cli.md` (lin
  - *Link to this PR:* this diff changes what the documented `logs` command captures, and the repo records every user-visible fix in CHANGELOG.md
- 🔵 **format_logs drops source, so uncaught errors print with no location** — `crates/tauri-pilot-cli/src/output.rs:168` — The human-readable renderer writes `[time] level args` and never prints `source`. For console entries the arguments usually carry context, but a new uncaught-error entry holds only `event.message`, so `tauri-pilot logs --level error`, the step README:186 documents for finding JS errors, shows `error Uncaught TypeError: x is not a function` with no file or line. The location the listeners compute i
  - *Fix:* Append the location when present, e.g. write ` (source)` after `args_str` for entries that carry a `source`, or at least for `level == "error"`, and extend the `format_logs` unit test at output.rs:904 with an entry whose `source` is set.
  - *Link to this PR:* the diff adds entries whose only location information lives in the field this renderer discards

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `f397248837` |
| Base | `main` |
| Files changed | 2 (+446/-9) |
| Languages | javascript |
| Reviewers | lang-javascript, maintainability, product-logic, quality, security, tests |

</details>

mpiton:
@RAMU324E thanks for the PR, the direction is right (error + unhandledrejection into the log buffer). The blocking note is stackSource picking frames by index (wrong frame on WebKit, message text on V8); the other comments are test strength and dropping the hostile-reason helpers.

I’ll wait until 25 Sep. If you’re busy I can push the remaining fixes to this branch (maintainers can edit) and keep your authorship. Just say if you’d rather I take it.

RAMU324E:
Thanks for the review! I'm a bit busy recently, so please feel free to
push the remaining fixes directly to this branch. I appreciate your help!

Mathieu ***@***.***> 于2026年9月18日周五 23:18写道：

mpiton:
Pushed the remaining review fixes to this branch (`d343541`), keeping your commits and authorship.

- `stackSource` / `extractSource` now share `firstAppFrame`: first `:line:col` frame, not a fixed skip. JSC-shaped stacks and V8 multi-line messages have tests.
- Dropped `safeTag` / `safeRead` / `safeString`. Cross-realm Errors and circular reasons stay; a listener `try/catch` records `[unprintable]` if a reason is hostile.
- Calling-site and rejection `source` assertions are actually able to fail now.
- Human-readable `logs` prints `source` when set. CHANGELOG, README, and the CLI reference mention uncaught errors / unhandled rejections.

I did not change `.github/workflows/ci.yml` here (GitHub blocked a workflow update on the fork). After merge I will add `**/*.js` / `**/*.mjs` to the path filters and a `node --test crates/tauri-plugin-pilot/js/*.test.mjs` step on `main`.

#191 is still stacked on the previous tip of this branch, so it will need a rebase once this merges.

mpiton:
Follow-up: GitHub reports this PR **conflicting** with `main` (radio `check` + screenshot scrollbar copy). A rebase of this fork branch cannot be pushed here (`workflow` scope on `.github/workflows/ci.yml` history from `main`).

The same commits, rebased onto current `main`, are in https://github.com/mpiton/tauri-pilot/pull/208 — original five commits still authored by you, plus the review-fix commit. Happy to close this one in favor of #208 once that lands, or to keep both until you say otherwise.

Issue #165, opened 2026-09-10
Test suite leaves stale /tmp/tauri-pilot-test-*.sock files


mpiton:
The test suite leaves its Unix sockets behind in `/tmp`.

## Reproduce

```
$ cargo test --workspace
$ ls /tmp/tauri-pilot-test-*.sock | wc -l
35
```

The count grows with every run; nothing ever removes them.

## Expected

Tests that bind a socket should unlink it on teardown — a guard type holding the path and unlinking in `Drop` covers both the normal exit and a panicking test. Alternatively, bind inside a per-test `tempfile::TempDir` so cleanup is automatic.

Low impact, but it accumulates on any machine that runs the suite regularly, and a stale socket with a colliding name can confuse a later run.

PR #193, opened 2026-09-14
fix(plugin): unix socket guard unlinks socket file on drop (#165)

mpiton:
Socket guard compared listener fd inode (sockfs inode) with socket path inode (filesystem inode). These never match, so guard never removed socket file.

Each cargo test run left a /tmp/tauri-pilot-test-*.sock behind. CLI auto-detection picks newest tauri-pilot-*.sock in /tmp when no $XDG_RUNTIME_DIR socket exists, so could bind dead socket instead of running app.

Guard now reads inode from socket path immediately after bind, ensuring it only removes the socket it created. Regression test verifies guard removes its own socket and preserves sockets created by other instances at same path.

---
## Summary by cubic
Fixes the Unix socket guard so it removes the socket file on drop instead of leaving stale files that CLI auto-detection could pick, and closes a race that could let a starting instance swap in its own socket.

- The guard now reads the inode from the socket path immediately after bind, before any fallible setup, so it only unlinks the socket it created.
- The guard is dropped before the listener closes, so a starting instance can't see the socket as stale and replace it between the inode check and unlink.
- On a normal app quit, the guard never drops (the server task lives on Tauri's static runtime), so the socket file stays until the next bind removes it as stale.
- Adds a regression test verifying the guard removes its own socket and preserves sockets bound by other instances at the same path.

<sup>Written for commit 0c7f5cda73a298f6256245c90c300717c6f6b849. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Unix socket cleanup now occurs reliably when the server ends.
  - Stale socket files are removed when safely rebinding.
  - Cleanup no longer removes a socket that has been reused by another instance.
  - Socket cleanup behavior is now protected against timing-related races.

- **Documentation**
  - Updated the changelog to clarify Unix socket cleanup behavior and document the related issue.


mpiton (commented):
## 🧨 mpiton-review — PR #193

> [!IMPORTANT]
> 💬 **Review posted** — No blocking issue. 3 findings to consider.

|:---|---:|:---|
| 🟡 Medium | 1 | fix now or file a follow-up |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/server/unix.rs` | 0 | 0 | 1 | 2 |

### 🎯 Start here

1. 🟡 **SocketGuard doc promises cleanup on normal shutdown, but a quitting app never drops it** — `crates/tauri-plugin-pilot/src/server/unix.rs:15`
2. 🔵 **SocketGuard is built three fallible steps after bind, apart from the Drop check** — `crates/tauri-plugin-pilot/src/server/unix.rs:190`
3. 🔵 **Negative case only holds while `_listener` pins the unlinked socket's inode** — `crates/tauri-plugin-pilot/src/server/unix.rs:523`

<details>

| | |
|:---|:---|
| Commit reviewed | `59dbf1c017` |
| Base | `main` |
| Files changed | 2 (+42/-18) |
| Languages | rust |

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -6,14 +6,13 @@ use crate::recorder::Recorder;
...
-use std::os::unix::io::AsRawFd;
 use std::os::unix::net::SocketAddr;
 use std::sync::Arc;
 use tokio::net::UnixListener;
 
 /// RAII guard that removes the socket file on drop (normal shutdown or panic).
-/// Stores the inode at bind time so it only unlinks its own socket, not one
-/// created by an overlapping instance.
+/// Stores the socket file's inode at bind time so it only unlinks its own
+/// socket, not one created by an overlapping instance.
```
> **🟡 Medium · 📄 Docs** — SocketGuard doc promises cleanup on normal shutdown, but a quitting app never drops it

`init` moves the guard into the `server::run` task on `tauri::async_runtime`, a static runtime that is never shut down, and `App::run` ends in tao's `process::exit`, so quitting an app never runs `Drop` and `tauri-pilot-<id>.sock` stays in `$XDG_RUNTIME_DIR`. `resolve_socket` scans that directory first and returns the newest socket without probing it, so the CLI can still pick a quit app's dead socket. Now that `Drop` really unlinks, readers will trust the '(normal shutdown or panic)' promise on this line and rule that case out.

**Fix:** Say when the guard actually drops (the server task ends, is aborted or panics) and that a normal quit leaves the file until the next bind at that path removes it as stale; one clause in the #165 CHANGELOG entry would help too. Removing the socket on quit (keep the guard in plugin state and drop it from the plugin's `on_event` on `RunEvent::Exit`) is a separate fix for a follow-up issue.

```suggestion
/// RAII guard that removes the socket file on drop.
///
/// The guard drops when the server task ends, is aborted, or panics. Quitting
/// the app never drops it: the task lives on Tauri's static runtime, which is
/// never shut down, so the file stays until the next bind at the same path
/// removes it as stale.
///
/// Stores the socket file's inode at bind time so it only unlinks its own
/// socket, not one created by an overlapping instance.
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR makes `Drop` unlink the file for the first time and rewrites lines 14-15 of this doc comment, leaving the unchanged '(normal shutdown or panic)' claim on line 13 as the one readers will now rely on.
- **References:** https://docs.rs/tauri/2/tauri/struct.App.html#method.run

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -198,7 +185,9 @@ fn bind_pathname(
     listener.set_nonblocking(true)?;
 
     tracing::info!(version = env!("CARGO_PKG_VERSION"), path = %socket_path.display(), "tauri-pilot socket listening");
-    let inode = inode_from_raw_fd(listener.as_raw_fd());
+    // Read the inode from the path: `fstat` on the listener fd reports its
+    // sockfs inode, which never matches the socket file (#165).
+    let inode = std::fs::metadata(socket_path)?.ino();
```
> **🔵 Low · 🧱 Maintainability** — SocketGuard is built three fallible steps after bind, apart from the Drop check

#165 came from the guard's key and its `Drop` check reading the inode from two different sources. The key is still computed out here in `bind_pathname` while `Drop` does its own `metadata(...).ino()`, and the guard only exists after `set_permissions(...)?`, `set_nonblocking(...)?` and the new `metadata(...)?`, so an error in any of them drops the listener and leaves the socket file behind. None of these calls fails in practice today; this is about keeping the pair from drifting apart again.

**Fix:** Add `SocketGuard::new(path: &Path) -> std::io::Result<Self>` that reads `std::fs::metadata(path)?.ino()`, and call it as soon as `match first_bind` yields `listener`, before `set_permissions`. A later setup error then drops the guard, which unlinks the file while the listener is still open, and `bind_pathname` returns `(listener, guard)`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** The diff rewrites the inode capture at this spot and adds a fallible `?` between bind and guard construction.

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: lang-rust, maintainability · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -514,4 +504,28 @@ mod tests {
...
+
+        let (listener, guard) = bind(&address).expect("bind test socket");
+        drop(listener);
+        drop(guard);
+        let own_left = socket.exists();
+        let _ = std::fs::remove_file(&socket);
+        assert!(!own_left, "guard must unlink the socket it bound");
+
+        // Another instance re-bound the path: that socket is not ours to remove.
+        let (_listener, guard) = bind(&address).expect("bind test socket");
```
mpiton, line 544: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Negative case only holds while `_listener` pins the unlinked socket's inode

The name `_listener` reads as an unused binding, but the second half of the test depends on it: it keeps the first socket's inode allocated after `remove_file`, so the rebind gets a different inode. Closing that listener before the rebind hands the new socket the same inode every time on ext4 `/tmp` (50/50 runs here, 0/50 while it stays open), so tidying the binding to `let (_, guard)` makes the guard delete `_other` and the test fails on ext4 hosts only.

**Fix:** Keep the binding and add a comment explaining it: the first listener must stay alive past the rebind so the kernel cannot give its inode to the new socket.

```suggestion
        // Another instance re-bound the path: that socket is not ours to remove.
        // `_listener` must outlive the rebind: it pins the unlinked socket's
        // inode. Closed early, ext4 hands that inode to the new socket and the
        // guard deletes it.
        let (_listener, guard) = bind(&address).expect("bind test socket");
```

crates/tauri-plugin-pilot/src/server/unix.rs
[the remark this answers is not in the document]
mpiton: Valid, but only part of it is fixable in this PR. 0c7f5cd moves the inode read into `SocketGuard::new`, called right after bind, so B's remove and rebind can no longer land during the chmod and non-blocking setup. What's left is the gap between bind and the `stat`, and it comes from the stale-removal path not being serialized: in the same race B's `remove_file` already deletes A's fresh socket and leaves A unreachable, guard or no guard. That needs a lock around bind, tracked in #195. Today the guard never drops on a normal quit (#194), only when the server task ends early, so the wrong unlink mostly matters once #194 lands.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: Applied the suggested doc comment in 0c7f5cd, with a pointer to #194, and added a clause to the #165 CHANGELOG entry. The quit-time cleanup itself is #194.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: Done in 0c7f5cd. `SocketGuard::new` reads `metadata(path)?.ino()`, and `bind_pathname` creates the guard right after the match, before `set_permissions`, so a setup error drops the guard (and unlinks the file) while the listener is still open.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: Kept the binding and added the comment from the suggestion in 0c7f5cd.

mpiton:
@coderabbitai on the two outside-diff findings in `unix.rs`:

25-28 (drop the guard before the listener): fixed in 0c7f5cd. `run` now rebinds the guard after the tokio listener and passes `&listener` to `accept_loop`, so on cancel or unwind the guard unlinks while the listener is still open, and a starting instance's `connect` succeeds instead of reporting the socket as stale. `bind` stays synchronous and still returns `AddrInUse`. I didn't change the test: the swap has to land between `metadata` and `remove_file` inside `Drop`, and there's no way to get a competitor in there without a test-only hook. I checked the order with a standalone async fn polled once and then dropped: before, listener then guard; now, guard then listener.

48-51 (check-and-unlink isn't atomic): valid in general, but a lock in `Drop` wouldn't close it. The only processes that would take the lock are `bind_pathname` callers, and they already see a live socket while the listener is open and leave it alone. Anything else that swaps the file ignores an advisory lock, and there's no unlink-if-inode-matches call to close the window from the guard; such a process runs as the same user and can delete the socket at any time anyway. The `SocketGuard` doc (`unix.rs:20-21`) only claims it won't unlink a socket from an overlapping instance, so I left it. The bind-side lock is tracked in #195.

PR #202, opened 2026-09-17
fix(plugin): pin bridge commands to the origin they checked

### LINKED ISSUE #173: Bridge commands can run on a page that replaced the one they checked
Since #169, `eval_bridge` checks that the target page has a bridge that can call back before it sends the script:

https://github.com/mpiton/tauri-pilot/blob/568a519d4cf2d4dbc14b85dc1bcc911b42cd718b/crates/tauri-plugin-pilot/src/handler.rs#L684-L692

The check reads `target.url()`, but `eval` only queues the script. It runs on whatever document the webview shows when the event loop gets to it. If the page changes in between, the script runs on a page nobody checked.

Ways to get there:

- Another connection sends `navigate` to a foreign origin. Each connection runs on its own task ([server/unix.rs L275](https://github.com/mpiton/tauri-pilot/blob/568a519d4cf2d4dbc14b85dc1bcc911b42cd718b/crates/tauri-plugin-pilot/src/server/unix.rs#L275)), so nothing orders the two requests.
- The page redirects itself, e.g. a login flow bouncing through an identity provider.
- Someone clicks a link in the window while an agent drives it.

On a foreign page the bridge is still injected but the ACL drops `__callback` (the #153 root cause). So `click` clicks, `fill` types into the other site's form, and the request then waits out the 10 s `DEFAULT_TIMEOUT` and fails with a generic timeout instead of the no-bridge error.

I haven't reproduced this end to end. It comes from reading the code, and cubic flagged it on #169. The gap between the check and the eval is small, so it takes bad timing in every case.

## Expected

A bridge command runs on the page it checked, or fails fast with the no-bridge error.

## Possible fix

Serializing requests per window would cover the concurrent `navigate` case, not redirects or user clicks. Checking inside the page covers all three:

- Pass the checked origin to `wrap_script` and have the wrapper compare it with the page's own before running the command. Build the page side with the same rules as [`origin_key`](https://github.com/mpiton/tauri-pilot/blob/568a519d4cf2d4dbc14b85dc1bcc911b42cd718b/crates/tauri-plugin-pilot/src/eval.rs#L29), since `location.origin` can be `"null"` for `file:` and custom schemes.
- On a mismatch the wrapper can't report back, since the callback is denied there. So on timeout, re-read `target.url()` and return the no-bridge error when the origin moved.

mpiton:
## Summary

- Pin each bridge command to the origin `eval_bridge` checked
- Refuse the command in-page if `location` is a different origin
- On timeout, re-read `target.url()` and return the no-bridge error when the origin moved

## Motivation

`eval` only queues the script. After `has_bridge` reads `target.url()`, a concurrent `navigate`, a redirect, or a user click can replace the page. The command then runs on a foreign origin whose ACL drops `__callback`, waits out `DEFAULT_TIMEOUT`, and fails with a generic timeout. Closes #173.

## Changes

- `wrap_script` embeds the checked origin and compares a page-side key built like `origin_key` (not `location.origin`, which is `"null"` for `file:` and custom schemes)
- `eval_bridge` fails fast if the URL already moved after eval, and reclassifies a later timeout
- Fake webviews expose a live URL so tests can navigate mid-request
- Changelog entry for the unreleased fix

## Test Plan

## Checklist

---
## Summary by cubic
Fixes a race condition where a bridge command could run on a page that replaced the one it checked, causing generic timeouts or executing on a foreign origin. Closes #173.

**Bug Fixes**
- Commands now pin to the origin that passed the check, matching scheme, host, and explicit port so an empty `location.port` can't pass a pin for a non-default checked port; the wrapper refuses to run if the page changed.
- `navigate` is not pinned so it still works from foreign pages; relative destinations resolve against the checked page.
- If the origin moved after eval, the RPC fails immediately; a later timeout returns an error naming the checked and new page instead of a generic eval timeout.
- Fake webviews now expose a live URL so tests can navigate mid-request.

<sup>Written for commit 7a6be8a0f18a3137536d615ee2d064297213567c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/202?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Bridge commands now verify that the page remains on the expected origin before and after execution.
  - Clear errors identify origin changes and missing bridge connections.
  - Relative navigation destinations are resolved from the current page URL.
  - Navigation continues to support JavaScript URLs and destination-bridge recovery.

- **Documentation**
  - Updated architecture and CLI documentation to describe origin verification, navigation behavior, and related errors.

- **Bug Fixes**
  - Improved handling of navigation races, redirects, and timeout classification.


crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -582,7 +582,8 @@ async fn handle_navigate(
     let target = target(webviews, window)?;
     // Read before the eval so the URL names the page the script lands on.
     let page = target.url();
-    let (id, rx) = send_script(&script, engine, target.as_ref())?;
+    let origin = page.as_ref().map(origin_key);
+    let (id, rx) = send_script(&script, engine, target.as_ref(), origin.as_deref())?;
```
mpiton: > [!WARNING]
> **🟠 High · 🎯 Product** — Navigate is origin-pinned, so recovery can refuse while a redirect is in flight

`handle_navigate` now passes `origin_key(page)` into `wrap_script`, so the wrapper returns without assigning `window.location.href` if `location` has already left that origin. Architecture and CHANGELOG (#153) make `navigate` the way back from a foreign page; a login bounce or concurrent redirect is the same #173 race, so recovery no-ops while `wait_dest_bridge` waits for a hello that never comes. Absolute dests still work on whichever document eval lands on; relative dests must not be resolved against the new page.

**Fix:** Do not pin `handle_navigate`. Pass `None` into `send_script`, and put the already computed absolute `dest` (from `navigate_destination`) into the script so a relative url cannot resolve against a page that moved.

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.88</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -686,14 +689,55 @@ async fn eval_bridge(
...
             &format!("no pilot bridge on the current page ({page})"),
         ));
     }
-    let (id, rx) = send_script(script, engine, target.as_ref())?;
-    wait(engine, id, rx, timeout).await
+    let origin = checked.as_ref().map(origin_key);
+    let (id, rx) = send_script(script, engine, target.as_ref(), origin.as_deref())?;
+    if let Some(page) = origin_moved_page(checked.as_ref(), target.url().as_ref()) {
+        return Err(fail_no_bridge(engine, id, page));
+    }
```
> **🟡 Medium · 🐛 Correctness** — Origin-move path always claims the new page has no bridge

`eval_bridge` maps any `origin_key` change to `no pilot bridge on the current page`, both on the fail-fast return after `send_script` and later in `timeout_or_moved_origin`, without calling `has_bridge` on the new URL. A click that races with a navigation to another origin that already said hello (a host in `remote.urls`) therefore names that page as having no bridge while `bridge_origins()` still lists it. The command must still fail — the wrapper already refused it — but the message tells the caller to navigate to a page that already answers.

**Fix:** Keep failing when the origin moved. Use a distinct message that the command was pinned to the checked origin and the page changed, instead of `fail_no_bridge` / `no_bridge_error` on the new URL. Do not wait for a callback: the wrapper returned without running the command.

<sub>🤖 <b>mpiton-review</b> · lang-rust · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/webview.rs
```diff
@@ -312,14 +341,27 @@ pub(crate) mod fake {
...
+            webviews.set_url("main", Some("https://other.test/"));
+            assert_eq!(
+                webviews
+                    .target(None)
+                    .expect("main window")
+                    .url()
+                    .map(|url| url.to_string()),
+                Some("https://other.test/".to_owned())
+            );
+        }
```
mpiton, line 374: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — url_is_live_after_set_url does not hold the target across set_url

`FakeTarget` now looks up the URL by label so a target already held across `eval` sees `set_url`. This test calls `set_url` and then `target()`, which also passed when `FakeTarget` snapshotted the URL inside `target()`. Only the handler fail-fast test would catch that regression.

**Fix:** Resolve `target` first, call `set_url`, then assert on that same `TargetWindow`.

```suggestion
        fn url_is_live_after_set_url() {
            let webviews = FakeWebviews::window("main", Some("https://app.test/"));
            let target = webviews.target(None).expect("main window");
            webviews.set_url("main", Some("https://other.test/"));
            assert_eq!(
                target.url().map(|url| url.to_string()),
                Some("https://other.test/".to_owned())
            );
        }
```

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.96</sub>

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -61,6 +61,23 @@ fn origin_keys(url: &Url) -> Vec<String> {
...
+/// http(s), ws(s) and ftp; changing them would let a page slip the wrapper
+/// after a default-port URL was checked.
+const PAGE_ORIGIN_EXPR: &str = concat!(
+    "(function(){var h=location.hostname,s=location.protocol.slice(0,-1),p=location.port;",
+    "if(h){if(!p){if(s==='https'||s==='wss')p='443';",
+    "else if(s==='http'||s==='ws')p='80';",
+    "else if(s==='ftp')p='21';}",
+    "return p?s+'://'+h+':'+p:s+'://'+h;}",
+    "return location.href.split('#')[0];})()",
+);
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧱 Maintainability** — PAGE_ORIGIN_EXPR reimplements origin_key with a hardcoded port table

`wrap_script` only receives the already-stringified `origin_key`, so `PAGE_ORIGIN_EXPR` has to rebuild that key from `location`, including a scheme-to-port table (443/80/21) copied from `Url::port_or_known_default`. `origin_key` uses the crate function; the next `origin_key` or `url` crate change desyncs the guard and either blocks a same-origin command or lets a foreign page run it.

**Fix:** Pass the checked `Url` into `wrap_script` and `send_script` instead of `Option<&str>`. Inject scheme, hostname, and `port_or_known_default` (or the hostless href) and compare those fields to `location`. Delete `PAGE_ORIGIN_EXPR`.

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -360,13 +388,64 @@ mod tests {
...
+        assert!(
+            script.contains(&encoded),
+            "wrapper must embed the checked origin; got: {script}"
+        );
+        let guard = script.find("return").expect("early return on mismatch");
+        let body = script.find("element.click()").expect("user script");
+        assert!(
+            guard < body,
+            "origin guard must run before the command; got: {script}"
+        );
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Origin-guard test passes without a mismatch return

`script.find("return")` matches `return p?s+'://'+h+':'+p...` inside `PAGE_ORIGIN_EXPR`, not the mismatch `if (...) return`. Combined with `script.contains(PAGE_ORIGIN_EXPR)` (the same constant `wrap_script` interpolates), the test still passes if the wrapper only mentions the origin and the expression and then always runs `element.click()`. Handler tests never execute that JS either, so a no-op guard would not fail CI.

**Fix:** Assert the concrete `if({PAGE_ORIGIN_EXPR}!=={encoded})return` substring appears before the user script. Add a node mock-`location` case (alongside the existing `js/*.test.mjs` files) so `PAGE_ORIGIN_EXPR` equals `origin_key` for `https://host/`, `http://host/`, `tauri://localhost/`, and `file:///tmp/a.html#frag`.

```suggestion
        assert!(
            script.contains(&format!("if({PAGE_ORIGIN_EXPR}!=={encoded})return;")),
            "wrapper must refuse the command on origin mismatch; got: {script}"
        );
        let guard = script
            .find(&format!("if({PAGE_ORIGIN_EXPR}!=={encoded})return;"))
            .expect("early return on mismatch");
        let body = script.find("element.click()").expect("user script");
        assert!(
            guard < body,
            "origin guard must run before the command; got: {script}"
        );
```

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.93</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -686,14 +689,55 @@ async fn eval_bridge(
...
+
+/// The page `now` if it is a different origin from `checked`.
+fn origin_moved_page<'a>(
+    checked: Option<&tauri::Url>,
+    now: Option<&'a tauri::Url>,
+) -> Option<&'a tauri::Url> {
+    let now = now?;
+    let checked = checked?;
+    (origin_key(checked) != origin_key(now)).then_some(now)
 }
```
mpiton, line 776: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Same-origin path change never exercised for origin_moved_page

`origin_moved_page` compares `origin_key`, so a path-only change on the app origin must not become a no-bridge error. Every new dispatch test jumps from `APP_PAGE` (`tauri://localhost/`) to `FOREIGN_PAGE` (`https://example.com/`), so replacing that comparison with full URL equality would still pass, and an in-app navigation during `click`/`title` would then fail fast as no-bridge.

**Fix:** Add a sibling of `test_dispatch_timeout_on_same_origin_stays_eval_timeout` that `set_url`s `main` to `tauri://localhost/settings` in `on_eval` (and one after a short delay) and still expects the eval-timeout error, not `no pilot bridge`.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.92</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -677,7 +678,9 @@ fn fail_no_bridge(engine: &EvalEngine, id: u64, page: &tauri::Url) -> RpcError {
 /// Send a bridge script and wait for its callback.
 ///
 /// Refuses without running the script when the page has no bridge that can
-/// call back, instead of waiting out `timeout` (#153).
+/// call back, instead of waiting out `timeout` (#153). The wrapper also pins
+/// the origin that passed that check, so a navigation in the gap before eval
+/// cannot run the command on a page nobody checked (#173).
```
> **🟡 Medium · 📄 Docs** — Architecture and CLI origin docs still describe the racy check-then-eval flow

`docs/src/content/docs/guides/architecture.md` still says the handler reads `target.url()` and then sends the script, and that on any other origin "the script runs and its result is dropped". This PR exists because `eval` only queues the script: `wrap_script` now embeds the checked origin key, a mismatch does not run the command, and a timeout after the origin moved is the no-bridge error. The `navigate` notes in `docs/src/content/docs/reference/cli.md` have the same gap, so readers of the #153 contract will still reason from the check that #173 proved insufficient.

**Fix:** Extend architecture.md § Origins without a callback, and add a sentence on the cli.md `navigate` origin contract: the wrapper compares a page-side key built like `origin_key` (not `location.origin`); a mismatch does not run the command; if the URL already moved after eval the RPC fails immediately with no-bridge, otherwise a later timeout is reclassified as that error.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR adds the origin pin and timeout reclassification; #170 documented the sibling hello check in those same two pages alongside CHANGELOG.

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.92</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -709,13 +753,17 @@ fn target<'a>(
 }
 
 /// Register a callback, then eval `script` wrapped in the ADR-001 pattern.
+///
+/// `origin` is the page that passed the bridge check. The wrapper refuses to
+/// run the command if the document that eventually evals it is elsewhere.
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — `send_script` rustdoc claims origin already passed the bridge check

`handle_navigate` always passes `origin_key(page)` into `send_script` before it branches on `has_bridge`, including when the current page has no hello so `navigate` can still leave. The new rustdoc says `origin` is "the page that passed the bridge check", which is true only for `eval_bridge`. A later edit that dropped the pin on the no-bridge path would re-open #173 for `navigate`.

**Fix:** Document `origin` as the `origin_key` of the URL this eval was aimed at, which the wrapper compares before running the command, including when `navigate` runs on a page with no bridge.

```suggestion
/// `origin` is the `origin_key` of the URL this eval was aimed at. The wrapper
/// refuses to run the command if the document that eventually evals it is elsewhere.
```

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.86</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #202

> [!IMPORTANT]
> 💬 **Review posted** — 1 blocking issue (0 critical, 1 high) out of 8 findings.

**Actionable comments posted: 8**

|:---|---:|:---|
| 🟡 Medium | 6 | fix now or file a follow-up |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/handler.rs` | 0 | 1 | 3 | 1 |
| `crates/tauri-plugin-pilot/src/eval.rs` | 0 | 0 | 2 | 0 |
| `crates/tauri-plugin-pilot/src/webview.rs` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟠 **Navigate is origin-pinned, so recovery can refuse while a redirect is in flight** — `crates/tauri-plugin-pilot/src/handler.rs:586`
2. 🟡 **Origin-move path always claims the new page has no bridge** — `crates/tauri-plugin-pilot/src/handler.rs:703`
3. 🟡 **url_is_live_after_set_url does not hold the target across set_url** — `crates/tauri-plugin-pilot/src/webview.rs:363`
4. 🟡 **PAGE_ORIGIN_EXPR reimplements origin_key with a hardcoded port table** — `crates/tauri-plugin-pilot/src/eval.rs:79`
5. 🟡 **Origin-guard test passes without a mismatch return** — `crates/tauri-plugin-pilot/src/eval.rs:426`

<details>

| | |
|:---|:---|
| Commit reviewed | `5675959d0f` |
| Base | `main` |
| Files changed | 4 (+340/-51) |
| Languages | rust |

</details>

crates/tauri-plugin-pilot/src/webview.rs
```diff
@@ -128,18 +128,23 @@ impl<R: tauri::Runtime> TargetWindow for tauri::WebviewWindow<R> {
...
 
     /// In-memory [`Webviews`] for handler tests.
     ///
     /// Without a label, `target` picks `main`, then the first window by label.
     /// Every evaluated script is recorded, then the responder runs: that is
     /// where a test plays the bridge and resolves the engine.
-    #[derive(Default)]
+    /// `url()` is live: [`Self::set_url`] is visible to a `target` already
+    /// held across an await, matching a real webview that navigated.
+    #[derive(Clone, Default)]
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When an eval responder captures a cloned `FakeWebviews`, the shared responder `Arc` forms a reference cycle

Fixed in `5724ee9`. `Clone` shares windows/scripts and starts with an empty responder. `clone_shares_windows_not_responder` covers it.

CHANGELOG.md
```diff
@@ -31,6 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bridge commands no longer run on a page that replaced the one they
+  checked. The wrapper compares origins before the command, and a
+  timeout after the origin moved returns the no-bridge error instead
+  of a generic eval timeout. [#173]
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> The new `[#173]` changelog reference has no matching definition

Fixed in `5724ee9`. Added `[#173]: https://github.com/mpiton/tauri-pilot/issues/173`.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: @mpiton

> Navigate is origin-pinned, so recovery can refuse while a redirect is in flight

Fixed in `5724ee9`. `handle_navigate` passes `None` into `send_script` and assigns the absolute dest from `navigate_destination`. `javascript:` urls stay as-is.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: @mpiton

> Origin-move path always claims the new page has no bridge

Fixed in `5724ee9`. Fail-fast and timeout now return `command was pinned to {checked} and the page changed to {now}` instead of `fail_no_bridge`. `test_dispatch_after_origin_change_does_not_claim_new_page_has_no_bridge` covers a dest that already said hello.

crates/tauri-plugin-pilot/src/webview.rs
mpiton: @mpiton

> url_is_live_after_set_url does not hold the target across set_url

Fixed in `5724ee9`. The test now resolves `target` first, then `set_url`, then asserts on that same `TargetWindow`.

crates/tauri-plugin-pilot/src/eval.rs
mpiton: @mpiton

> PAGE_ORIGIN_EXPR reimplements origin_key with a hardcoded port table

Fixed in `5724ee9`. `wrap_script` takes `&Url` and injects scheme/host/`port_or_known_default` (or the hostless href). `PAGE_ORIGIN_EXPR` is gone.

crates/tauri-plugin-pilot/src/eval.rs
mpiton: @mpiton

> Origin-guard test passes without a mismatch return

Fixed in `5724ee9`. The wrapper test asserts `if({guard})return;` before `element.click()`. `origin-guard.test.mjs` mocks location for https/http/tauri/file.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: @mpiton

> Same-origin path change never exercised for origin_moved_page

Fixed in `5724ee9`. Two siblings of the same-origin timeout test set `tauri://localhost/settings` in `on_eval` and after a delay, and still expect eval timeout.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: @mpiton

> Architecture and CLI origin docs still describe the racy check-then-eval flow

Fixed in `5724ee9`. architecture.md Origins without a callback and the cli.md navigate notes now cover the field comparison, mismatch no-op, origin-moved error, and unpinned navigate with an absolute dest.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: @mpiton

> `send_script` rustdoc claims origin already passed the bridge check

Fixed in `5724ee9`. rustdoc now says `origin` is the URL this eval was aimed at, and that navigate passes `None`.

crates/tauri-plugin-pilot/src/eval.rs
```diff
@@ -61,6 +61,43 @@ fn origin_keys(url: &Url) -> Vec<String> {
...
+    match (
+        url.host_str().filter(|host| !host.is_empty()),
+        url.port_or_known_default(),
+    ) {
+        (Some(host), Some(port)) => {
+            let protocol = json(&format!("{}:", url.scheme()));
+            let host = json(host);
+            let port = json(&port.to_string());
+            format!(
+                "location.protocol!=={protocol}||location.hostname!=={host}||(location.port||{port})!=={port}"
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Line 81 substitutes the checked port when `location.port` is empty.

Fixed in `7a6be8a`. The guard now compares `location.port` to `Url::port()` (empty for a default port). `https://host:8443/` vs `https://host/` is in origin-guard.test.mjs and the eval unit test.

mpiton:
@coderabbit resolve

PR #203, opened 2026-09-17
fix(plugin): fail press when the window never gains focus

### LINKED ISSUE #175: press injects keys into whatever app has focus when the WM refuses the focus request
`press` reports `ok` and injects OS-level keystrokes even when the window manager refused to focus the target window, so the keys land in whatever application currently has focus.

Found during end-to-end QA of 0.8.0 on Linux/X11 (plugin and CLI 0.7.3, commit 568a519). This is not a regression from the `Webviews` refactor — the same focus-then-inject shape is present at a5d60f3.

## Reproduce

Give focus to another application (a browser window here), then drive the app under test:

```
$ xdotool getactivewindow getwindowname
Nouvel onglet - Google Chrome

$ tauri-pilot press "a"
✓ ok

$ xdotool getactivewindow getwindowname
Nouvel onglet - Google Chrome          # focus never moved

$ tauri-pilot text "#last-keydown"
CLEARED                                # the app never saw the key
```

Now force focus by hand and repeat:

```
$ xdotool windowactivate --sync $(xdotool search --name "Pilot DnD Test" | head -1)
$ tauri-pilot press "a"
✓ ok
$ tauri-pilot text "#last-keydown"
a (trusted)                            # works once the window really has focus
```

During this session the browser tab title changed from "Issues · …" to "Nouvel onglet" while `press "ctrl+1"` / `press "Enter"` were being sent, so the keystrokes were genuinely delivered to the other application rather than being dropped.

Global shortcuts registered through `tauri-plugin-global-shortcut` still fire, because the OS dispatches them regardless of focus. That masks the problem: `press "ctrl+1"` looks like it works while ordinary keys silently go elsewhere.

## Cause

`crates/tauri-plugin-pilot/src/handler.rs:482`:

```rust
match target.focus() {
    Ok(()) => {
        // Only wait if the WM actually accepted the focus request —
        // a failed focus call won't transfer focus, so sleeping
        // would just delay the press for nothing.
        tokio::time::sleep(Duration::from_millis(FOCUS_SETTLE_MS)).await;
    }
    Err(e) => { ... }
}
```

`target.focus()` calls Tauri's `set_focus()` (`crates/tauri-plugin-pilot/src/webview.rs:129`), which reports whether the activation *request* was dispatched, not whether focus was transferred. Under X11 focus-stealing prevention the window manager can ignore the request and `set_focus()` still returns `Ok(())`. The handler reads that as "the WM accepted", sleeps `FOCUS_SETTLE_MS`, and injects through enigo anyway.

The existing `Err` branch is the right instinct but only covers the case where the runtime call itself fails.

## Impact

- Any `press` in a QA run can silently do nothing while reporting success, which makes keyboard scenarios unreliable on a developer machine where other windows are in front.
- Synthetic keystrokes go to an unrelated application. Depending on what has focus that can be a terminal, an editor, a chat window or a password field. A QA tool should not be able to type into something it is not driving.

## Suggested fix

After `set_focus()`, poll `is_focused()` (`Window::is_focused` / `WebviewWindow::is_focused`, available in Tauri 2) until it returns true or a short deadline expires, and fail with a clear error instead of injecting when focus never arrived — something like `cannot press: window 'main' did not gain focus (another application has it)`. The `FOCUS_SETTLE_MS` sleep can become the polling budget, so the common case is not slower.

Add `is_focused` to the `Webviews` seam so the behaviour is testable with the mock.

mpiton:
## Summary

- Fail `press` when the target window never actually receives OS focus
- Poll `is_focused()` after `set_focus()` instead of treating a successful request as a transfer
- Add `is_focused` / `label` to the webview seam so the refusal path is testable

## Motivation

`set_focus()` reports that the activation request was dispatched, not that focus moved. On X11 with focus-stealing prevention the WM can ignore the request, `press` still reports `ok`, and the keys land in whichever app currently has focus. Closes #175.

## Changes

- After `focus()`, poll `is_focused()` for up to the existing 80ms settle budget
- Return `-32603` with `cannot press: window '<label>' did not gain focus (another application has it)` instead of injecting
- `FakeWebviews` tracks per-window focus; unfocused by default so tests cannot inject keys by accident
- Changelog and CLI `press` docs

## Test Plan

## Checklist

---
## Summary by cubic
Fixes `press` so it fails when the target window never gains OS focus, instead of injecting keys into whichever application currently has focus. `set_focus` only reports that the activation request was dispatched, which the window manager may ignore on X11; `press` now polls `is_focused()` and returns an error if the window does not gain focus within a single monotonic 80ms deadline, so slow focus queries count against the budget instead of extending the wait. Press-only items are gated so `--no-default-features` clippy keeps passing.

**Bug Fixes**
- Adds `is_focused` and `label` to the `TargetWindow` seam; on Windows focus is read from the OS foreground window since the WebView2 child HWND keeps tao's flag false, and query errors fail immediately rather than counting as unfocused.
- `FakeWebviews` now simulates per-window focus and starts unfocused, so tests must call `set_focused` before a press can succeed.
- Updates CLI reference and changelog.

<sup>Written for commit 11fac23e99d86e894ceb34d7369bef756a07a44a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `press` safety by confirming the target window has focus before sending keystrokes.
  * Returns an error and skips key injection when focus is refused, preventing input from reaching another application.
  * Added focus timeout and query-failure handling.

* **Documentation**
  * Updated CLI documentation to describe focus verification and error behavior.


crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -534,6 +536,41 @@ async fn handle_press(
...
+    let budget = Duration::from_millis(FOCUS_SETTLE_MS);
+    let poll = Duration::from_millis(FOCUS_POLL_MS);
+    let mut waited = Duration::ZERO;
+    loop {
+        if matches!(target.is_focused(), Ok(true)) {
+            return Ok(());
+        }
+        let remaining = budget.saturating_sub(waited);
+        if remaining.is_zero() {
+            return Err(press_unfocused_error(target.label()));
```
mpiton: > [!WARNING]
> **🟠 High · 🐛 Correctness** — Windows is_focused is false while the webview holds keyboard focus

`wait_until_focused` treats `WebviewWindow::is_focused()` as “this window will receive OS key events.” On Windows that getter is tao 0.35.3 `has_active_focus()` (`is_active && is_focused`, driven by `WM_NCACTIVATE` plus `WM_SETFOCUS`/`WM_KILLFOCUS` on the parent HWND). WebView2 is a child HWND; when it has keyboard focus the parent gets `WM_KILLFOCUS`, so `is_focused()` is false even though the app is foreground and `enigo` keys would land in the webview. tauri-runtime-wry 2.11.4 already special-cases this for events (single-webview mode discards the window blur and synthesizes `WindowEvent::Focused` from WebView2 `GotFocus`/`LostFocus`) but `WindowMessage::IsFocused` still reads the raw parent-HWND flag. `set_focus()` is also a no-op when the window is already the foreground window, so the 80ms poll cannot recover. A normal Windows `press` against the app the user is looking at then fails with “another application has it.”

**Fix:** Do not use parent-HWND `WM_SETFOCUS` as the fail-closed signal. On Windows, treat the target as focused when it is the OS foreground window (`GetForegroundWindow` equals the window HWND, or that HWND is an ancestor of `GetFocus`). Keep `is_focused()` as the gate only where it matches OS foreground (GTK `is_active`, macOS `isKeyWindow`).

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -534,6 +536,41 @@ async fn handle_press(
...
+/// `TargetWindow: Sync` (the box is owned across each sleep).
+#[cfg(feature = "press")]
+async fn wait_until_focused(target: Box<dyn TargetWindow + '_>) -> Result<(), RpcError> {
+    let budget = Duration::from_millis(FOCUS_SETTLE_MS);
+    let poll = Duration::from_millis(FOCUS_POLL_MS);
+    let mut waited = Duration::ZERO;
+    loop {
+        if matches!(target.is_focused(), Ok(true)) {
+            return Ok(());
+        }
```
> **🟡 Medium · 🐛 Correctness** — wait_until_focused treats is_focused errors as unfocused

`matches!(target.is_focused(), Ok(true))` collapses `Err(_)` into the same path as `Ok(false)`. `TargetWindow::is_focused` is documented to return a runtime error when focus cannot be queried (Tauri's getter fails with `FailedToSendMessage` / `FailedToReceiveMessage` if the window is gone or the event-loop proxy is dead). `press` then burns the full 80ms budget and reports that another application has focus, hiding the query failure.

**Fix:** Match on `target.is_focused()`: `Ok(true)` returns success, `Ok(false)` keeps polling, `Err(e)` returns `-32603` immediately with the query error and window label.

```suggestion
        match target.is_focused() {
            Ok(true) => return Ok(()),
            Ok(false) => {}
            Err(e) => {
                return Err(RpcError {
                    code: RPC_INTERNAL_ERROR,
                    message: format!(
                        "cannot press: failed to query focus for window '{}': {e}",
                        target.label()
                    ),
                    data: None,
                });
            }
        }
```

<sub>🤖 <b>mpiton-review</b> · 3 reviewers agree: lang-rust, maintainability, quality · confidence 1.0</sub>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -534,6 +536,41 @@ async fn handle_press(
...
+/// Takes the window by value so the future stays `Send` without requiring
+/// `TargetWindow: Sync` (the box is owned across each sleep).
+#[cfg(feature = "press")]
+async fn wait_until_focused(target: Box<dyn TargetWindow + '_>) -> Result<(), RpcError> {
+    let budget = Duration::from_millis(FOCUS_SETTLE_MS);
+    let poll = Duration::from_millis(FOCUS_POLL_MS);
+    let mut waited = Duration::ZERO;
+    loop {
+        if matches!(target.is_focused(), Ok(true)) {
+            return Ok(());
```
> **🟡 Medium · 🧪 Tests** — is_focused query errors are untested, so fail-closed can regress

`wait_until_focused` only proceeds on `Ok(true)`, so an `is_focused()` `Err` is treated like `Ok(false)` and becomes `press_unfocused_error` after the budget. `FakeTarget::is_focused` always returns `Ok`, and no test double returns `Err`, so a change that treated a query failure as focused would inject keys and every new test would still pass.

**Fix:** In `handler.rs` `mod tests`, add a `TargetWindow` whose `is_focused` returns `Err` and assert `wait_until_focused` does not return `Ok` (today that is the unfocused `-32603` after the budget). If `FakeWebviews` is the intended seam, give it a way to fail the query so a `dispatch("press", ...)` case can prove injection is skipped.

CHANGELOG.md
```diff
@@ -31,6 +31,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `press` now waits until the target window actually has OS focus after
+  `set_focus`, and returns an error instead of injecting when the window
+  manager refuses the request. Previously `set_focus` succeeding was treated
+  as focus transferred, so on X11 with focus-stealing prevention the keys
+  could land in another application while the RPC still reported `ok`. [#175]
```
mpiton, line 38: > [!NOTE]
> **🔵 Low · 📄 Docs** — CHANGELOG [#175] has no matching link definition

The new Fixed bullet ends with `[#175]`, but the reference list at the bottom of CHANGELOG.md has no `[#175]:` URL. Every other issue citation in this file has a matching definition, so this entry will not hyperlink while the rest do.

**Fix:** Add `[#175]: https://github.com/mpiton/tauri-pilot/issues/175` to the footer list, immediately after `[#173]`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the new Unreleased bullet added by this diff is the first `[#175]` citation

</details>

mpiton (commented):
## 🧨 mpiton-review — PR #203

> [!IMPORTANT]
> 💬 **Review posted** — 1 blocking issue (0 critical, 1 high) out of 4 findings.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/handler.rs` | 0 | 1 | 2 | 0 |

### 🎯 Start here

1. 🟠 **Windows is_focused is false while the webview holds keyboard focus** — `crates/tauri-plugin-pilot/src/handler.rs:554`
2. 🟡 **wait_until_focused treats is_focused errors as unfocused** — `crates/tauri-plugin-pilot/src/handler.rs:551`
3. 🟡 **is_focused query errors are untested, so fail-closed can regress** — `crates/tauri-plugin-pilot/src/handler.rs:550`
4. 🔵 **CHANGELOG [#175] has no matching link definition** — `CHANGELOG.md:38`

> Inline comments were posted individually because GitHub refuses REQUEST_CHANGES on your own PR.

<details>

| | |
|:---|:---|
| Commit reviewed | `95ff775fb9` |
| Base | `main` |
| Files changed | 4 (+262/-34) |
| Languages | rust |

</details>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -534,6 +536,41 @@ async fn handle_press(
...
+///
+/// Takes the window by value so the future stays `Send` without requiring
+/// `TargetWindow: Sync` (the box is owned across each sleep).
+#[cfg(feature = "press")]
+async fn wait_until_focused(target: Box<dyn TargetWindow + '_>) -> Result<(), RpcError> {
+    let budget = Duration::from_millis(FOCUS_SETTLE_MS);
+    let poll = Duration::from_millis(FOCUS_POLL_MS);
+    let mut waited = Duration::ZERO;
+    loop {
+        if matches!(target.is_focused(), Ok(true)) {
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When the focus query fails, this branch silently treats the error as an unfocused window and returns the misleading “another application has it” message after the full timeout.

Fixed in 46a212f. `wait_until_focused` matches on `is_focused()` now: `Err` returns `-32603` immediately with the query error and window label. `test_wait_until_focused_errors_when_focus_query_fails` covers it.

CHANGELOG.md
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> The new changelog entry references `[#175]`, but there is no matching `[#175]:` definition in the link section.

Fixed in 46a212f. Added `[#175]: https://github.com/mpiton/tauri-pilot/issues/175` after `[#173]`.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > Windows is_focused is false while the webview holds keyboard focus

Fixed in 46a212f. On Windows, `TargetWindow::is_focused` now uses `GetForegroundWindow` (and `GetAncestor(GetFocus, GA_ROOT)`) instead of tao's parent-HWND `WM_SETFOCUS` flag. A focused `WebView2` child still counts as the key target.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > wait_until_focused treats is_focused errors as unfocused

Fixed in 46a212f. `Ok(true)` succeeds, `Ok(false)` keeps polling, `Err` returns `-32603` immediately with `cannot press: failed to query focus for window '{label}': {e}`.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > is_focused query errors are untested, so fail-closed can regress

Fixed in 46a212f. `FakeWebviews::set_focus_query_error` makes `is_focused` return `Err`. `test_wait_until_focused_errors_when_focus_query_fails` and `test_dispatch_press_skips_injection_when_focus_query_fails` assert that path, not the unfocused timeout.

CHANGELOG.md
mpiton: > CHANGELOG [#175] has no matching link definition

Fixed in 46a212f. `[#175]` is defined next to `[#173]` in the footer.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -534,6 +536,53 @@ async fn handle_press(
...
+            Ok(true) => return Ok(()),
+            Ok(false) => {}
+            Err(e) => {
+                return Err(RpcError {
+                    code: RPC_INTERNAL_ERROR,
+                    message: format!(
+                        "cannot press: failed to query focus for window '{}': {e}",
+                        target.label()
+                    ),
+                    data: None,
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Measure the focus budget with a monotonic deadline.

Fixed in d569279. `wait_until_focused` now uses `tokio::time::Instant` so `is_focused()` query time counts against `FOCUS_SETTLE_MS`, not just the sleeps between polls.

mpiton:
Addressed. Ready for re-review.

mpiton:
@coderabbit resolve

PR #206, opened 2026-09-18
fix(plugin): unlink socket file on normal quit (#194)

### LINKED ISSUE #194: Quitting the app leaves its tauri-pilot-*.sock file behind
Follow-up to #165 and #193. With #193, `SocketGuard` unlinks the socket file when it drops. It never drops when the app quits, though, so a normal quit still leaves `tauri-pilot-{identifier}.sock` behind.

`setup` moves the guard into the `server::run` task (`crates/tauri-plugin-pilot/src/lib.rs`). That task lives on `tauri::async_runtime`, a static runtime nobody shuts down. On quit, tauri dispatches `RunEvent::Exit`, runs `cleanup_before_exit()` (it clears tray icons and resource tables), then tao's event loop calls `std::process::exit` (tao 0.35.3, `platform_impl/linux/event_loop.rs:998`; macOS and Windows do the same). `process::exit` runs no destructors, so the task and its guard are never dropped. The file stays until the next bind at that path finds it stale and removes it.

I found this reading the exit path during the #193 review and haven't run it against a real app yet.

## Reproduce

```
# debug build of an app with tauri-plugin-pilot, #193 applied
$ ls "$XDG_RUNTIME_DIR"/tauri-pilot-*.sock
# quit the app by closing its last window
$ ls "$XDG_RUNTIME_DIR"/tauri-pilot-*.sock
```

Per the exit path above, the second `ls` still lists the socket after the process has exited.

The CLI trips over it. Without `--socket` or `TAURI_PILOT_SOCKET`, `resolve_socket` takes the newest `tauri-pilot-*.sock` in `$XDG_RUNTIME_DIR` (then `/tmp`) by mtime and never checks that anything listens on it. With no app running you get a connection error instead of "No tauri-pilot socket found". Worse: start app A, start app B, quit B, and the CLI picks B's dead socket while A is still up.

## Expected

Quitting the app removes the socket file.

The plugin can do that from an `on_event` hook, since tauri hands `RunEvent::Exit` to plugins (`on_event_loop_event` in `tauri/src/app.rs`) before tao exits the process. Keep the guard in plugin state instead of moving it into `run` (say `app.manage(Mutex<Option<SocketGuard>>)` in `setup`), then take it out and drop it on `RunEvent::Exit`. `Drop` keeps its inode check, so it still won't touch another instance's socket. Aborting the server task from `on_event` wouldn't be enough: the runtime drops the future on a worker thread, racing `process::exit`.

A signal or a crash skips `RunEvent::Exit`, so `newest_socket_in_dir` should also skip sockets that refuse a connection. The Windows path already does the equivalent with `is_pid_alive`.

Windows leaks the same way: `RegistryGuard` is also created inside `server::run`, so the instance file under `%LOCALAPPDATA%\tauri-pilot\instances\` survives a quit. `is_pid_alive` keeps that harmless for the CLI, but the same hook could drop it.

The `SocketGuard` doc comment promises removal on "normal shutdown or panic", which isn't true today (flagged in the #193 review). Worth fixing in the same change.

mpiton:
## Summary

- Drop `SocketGuard` / `RegistryGuard` from plugin state on `RunEvent::Exit` so a normal quit unlinks the pathname socket (and Windows instance file) before tao calls `process::exit`.
- CLI auto-detect skips `tauri-pilot-*.sock` files that refuse a connection, so a leftover file is not treated as a live app.
- If Windows bind finishes after Exit, a `Released` slot drops the guard immediately instead of storing it.

## Motivation

Follow-up to #165 / #193. The RAII guard never ran on quit because the server task lives on Tauri's static runtime and tao exits with `process::exit` (no destructors). The CLI then picked the newest leftover socket by mtime, including dead ones.

Closes #194

## Changes

- Plugin: keep the cleanup guard in managed state; `on_event` drops it on `RunEvent::Exit`. Do not move it into the accept-loop task.
- CLI: `newest_socket_in_dir` probes with `connect` and prefers a live older socket over a newer dead one. Explicit `--socket` is unchanged.
- `SocketGuard` docs now match shutdown reality (normal quit cleans up; crash may not).

## Test Plan

## Checklist

## Type

fix

---
## Summary by cubic
Fixes leftover socket files after a normal quit. The cleanup guard never ran because the server task lives on Tauri's static runtime and tao exits with `process::exit`, leaving the socket file (or Windows instance file) behind; the CLI auto-detect could then pick a dead socket as a running app.

- The plugin holds the cleanup guard in managed state and drops it on `RunEvent::Exit` before `process::exit`.
- CLI auto-detect probes sockets with a bounded connect attempt, skips dead files and symlinks, and prefers a live older socket over a newer dead one; explicit `--socket` is unchanged.
- On Windows, the guard is also released when the accept loop ends, and a bind finishing after Exit drops immediately instead of being stored.

<sup>Written for commit c9515c811b026ef1adf92711335aecfda7b4edd8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Unix socket detection to select the newest active listener while skipping stale or unreachable sockets.
  - Ensured socket and instance files are cleaned up when the plugin exits normally.
  - Prevented late-starting server processes from leaving behind registry entries after shutdown.
  - Improved resilience when connection probes close before sending request data.

- **Documentation**
  - Clarified socket cleanup behavior during normal exits, crashes, and forced termination.


crates/tauri-plugin-pilot/src/server/windows.rs
```diff
@@ -444,6 +445,13 @@ pub async fn run(
...
+    // Plugin Exit drops the guard; tests pass None and keep it in this task.
+    let _guard = if let Some(cleanup) = cleanup {
+        cleanup.hold_windows(guard);
+        None
+    } else {
+        Some(guard)
+    };
     if let Err(e) = accept_loop(first_server, &identifier, engine, webviews, recorder).await {
         tracing::error!("named pipe server error: {e}");
     }
```
mpiton, line 459: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — Pipe server death leaves the Windows instance file registered

Production `run` moves `RegistryGuard` into `PilotGuard` and keeps the task-local `_guard` as `None`. When `accept_loop` returns (the comment on `connect().await` says the pipe handle is dead), the app PID is still alive and nothing unregisters `%LOCALAPPDATA%/tauri-pilot/instances/*.json`. `resolve_socket_windows` only skips dead PIDs, so auto-detect returns a pipe that will fail to open. Before this change the guard lived in the task and Drop cleaned that path up.

**Fix:** Keep `cleanup` after `hold_windows` and call `cleanup.release()` once `accept_loop` returns. `release` is idempotent with `GuardSlot::Released`, so a later `RunEvent::Exit` is a no-op. Tests that pass `None` still drop `RegistryGuard` at the end of the task.

```suggestion
    // Plugin Exit drops the guard; tests pass None and keep it in this task.
    // Also release when the accept loop ends so a dead pipe is unregistered
    // while the app PID is still alive.
    let _guard = if let Some(cleanup) = &cleanup {
        cleanup.hold_windows(guard);
        None
    } else {
        Some(guard)
    };
    if let Err(e) = accept_loop(first_server, &identifier, engine, webviews, recorder).await {
        tracing::error!("named pipe server error: {e}");
    }
    if let Some(cleanup) = cleanup {
        cleanup.release();
    }
```

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.88</sub>

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -1600,7 +1600,20 @@ fn newest_socket_in_dir(dir: &Path) -> Option<PathBuf> {
...
+}
+
+/// True when a process is accepting on this pathname socket.
+///
+/// `ConnectionRefused` (and any other connect error) means no listener, so
+/// auto-detect must not treat the file as a running app (#194).
+#[cfg(not(windows))]
+fn socket_has_listener(path: &Path) -> bool {
+    std::os::unix::net::UnixStream::connect(path).is_ok()
 }
```
mpiton, line 1666: > [!IMPORTANT]
> **🟡 Medium · 🔒 Security** — Liveness probe follows /tmp symlinks past the UID ownership check

`newest_socket_in_dir` keeps a path when `metadata(p).uid()` matches, then `socket_has_listener` calls `UnixStream::connect(path)`. Both follow symlinks, so a world-writable `/tmp/tauri-pilot-*.sock` link to a live socket the user owns (D-Bus, ssh-agent, another app) passes the UID filter added as a security control, counts as live, and is returned as the auto-detected endpoint. This PR now probes every candidate and, when `$XDG_RUNTIME_DIR` only has dead leftovers, falls through to `/tmp`, so a planted symlink that previously lost to a leftover file can win and receive the later JSON-RPC.

**Fix:** In `socket_has_listener`, `lstat` with `symlink_metadata`, require `file_type().is_socket()` and `uid() == getuid()`, and only then `connect`. Apply the same `lstat`+socket-type check in the candidate filter so attacker-owned names never reach the probe.

```suggestion
fn socket_has_listener(path: &Path) -> bool {
    use std::os::unix::fs::{FileTypeExt, MetadataExt};
    let Ok(meta) = std::fs::symlink_metadata(path) else {
        return false;
    };
    if !meta.file_type().is_socket() {
        return false;
    }
    // SAFETY: getuid() has no preconditions.
    let my_uid = unsafe { libc::getuid() };
    if meta.uid() != my_uid {
        return false;
    }
    std::os::unix::net::UnixStream::connect(path).is_ok()
}
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the new connect() probe is the sink; it runs over the existing UID filter that still uses metadata() and follows links
- **References:** CWE-59

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.78</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -33,6 +33,102 @@ pub(crate) const BRIDGE_JS: &str = concat!(
...
+}
+
+#[cfg(all(any(unix, windows), debug_assertions))]
+fn hold_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>, value: T) {
+    let mut guard = lock_slot(slot);
+    if matches!(&*guard, GuardSlot::Released) {
+        drop(guard);
+        drop(value);
+        return;
+    }
```
mpiton, line 108: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Late-hold GuardSlot::Released path is never exercised

`hold_slot` must drop a guard when the slot is already `GuardSlot::Released`; that is the Windows bind-after-Exit race this PR adds. `normal_quit_unlinks_instance_file` polls until the instance file exists before `on_pilot_event(Exit)`, so `register_instance` and `hold_windows` have already run, and Unix bind is synchronous in `setup`, so Linux CI never takes this branch. Deleting the `Released` arm would still pass the new tests.

**Fix:** Add a `lib.rs` unit test that `release_slot`s a `Mutex<GuardSlot<_>>`, then `hold_slot`s a Drop spy and asserts the value is dropped immediately. Optionally change the Windows test to fire Exit first, wait for bind to finish, then assert the instance file is gone.

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.93</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #206

> [!IMPORTANT]
> 💬 **Review posted** — No blocking issue. 4 findings to consider.

|:---|---:|:---|
| 🔵 Low | 0 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/server/windows.rs` | 0 | 0 | 1 | 0 |
| `crates/tauri-pilot-cli/src/main.rs` | 0 | 0 | 1 | 0 |
| `crates/tauri-plugin-pilot/src/lib.rs` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟡 **Pipe server death leaves the Windows instance file registered** — `crates/tauri-plugin-pilot/src/server/windows.rs:457`
2. 🟡 **Liveness probe follows /tmp symlinks past the UID ownership check** — `crates/tauri-pilot-cli/src/main.rs:1617`
3. 🟡 **Late-hold GuardSlot::Released path is never exercised** — `crates/tauri-plugin-pilot/src/lib.rs:108`

<details>

- 🟡 **Unreleased changelog still says quit leaves the socket** — `CHANGELOG.md:72` — The Unreleased #165 note still tells readers the guard only drops when the server task ends and that a normal quit leaves `tauri-pilot-*.sock` until the next bind ([#194]). This PR closes that issue and changes CLI auto-detect to skip dead files, so the published product history describes the old bug as current behaviour.
  - *Fix:* Rewrite the #165 paragraph so it no longer claims quit leaves the socket, and add an Unreleased Fixed bullet for #194: `RunEvent::Exit` unlinks the pathname socket and the Windows instance file, and `newest_socket_in_dir` skips sockets that refuse a connection.
  - *Link to this PR:* this PR closes #194 but does not touch CHANGELOG.md

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `b4389d0c94` |
| Base | `main` |
| Files changed | 5 (+281/-15) |
| Languages | rust |
| Reviewers | lang-rust, maintainability, product-logic, quality, security, tests |

</details>

crates/tauri-plugin-pilot/src/server/windows.rs
```diff
@@ -444,6 +445,13 @@ pub async fn run(
         }
     };
     let identifier = guard.identifier.clone();
+    // Plugin Exit drops the guard; tests pass None and keep it in this task.
+    let _guard = if let Some(cleanup) = cleanup {
+        cleanup.hold_windows(guard);
+        None
+    } else {
+        Some(guard)
+    };
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Release the registry entry when the server task ends.

Fixed in `37fb023`. `run` keeps `cleanup` after `hold_windows` and calls `cleanup.release()` when `accept_loop` returns. `release` is a no-op if Exit already set the slot to `Released`.

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -1600,7 +1600,20 @@ fn newest_socket_in_dir(dir: &Path) -> Option<PathBuf> {
...
+        .find(|p| socket_has_listener(p))
+}
+
+/// True when a process is accepting on this pathname socket.
+///
+/// `ConnectionRefused` (and any other connect error) means no listener, so
+/// auto-detect must not treat the file as a running app (#194).
+#[cfg(not(windows))]
+fn socket_has_listener(path: &Path) -> bool {
+    std::os::unix::net::UnixStream::connect(path).is_ok()
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When a live socket’s accept backlog is full, this synchronous probe can hang the CLI before any configured connection timeout runs.

Fixed in `37fb023`. The probe `lstat`s first (uid + socket type, no symlink follow), then connects on a helper thread with a 200ms cap. Timeout counts as live: connect did not refuse, so something is listening. Leftovers still fail immediately with `ConnectionRefused`.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -121,10 +222,12 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                     engine,
                     webviews,
                     recorder,
+                    Some(cleanup),
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> passing `Some(cleanup)` leaves the `RegistryGuard` in app state after `run` returns.

Fixed in `37fb023`, in `windows.rs` `run` rather than here. Passing `Some(cleanup)` is still required so Exit can drop the guard; `run` now also `release()`s when `accept_loop` ends.

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -101,9 +200,11 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                     }) else {
                         return Ok(());
                     };
+                    // Keep the guard in plugin state, not the server task (#194).
+                    cleanup.hold_unix(guard);
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Coordinate listener shutdown before releasing the guard.

Not doing that. `RunEvent::Exit` is followed by `process::exit`, so waiting on the accept loop does not run. Unlinking the pathname while the dying listener still holds the inode is the usual Unix restart: a new instance `bind()`s a new inode, the CLI hits that, and the old fd dies with the process. Aborting the server task from `on_event` is the race #194 was written to avoid.

crates/tauri-plugin-pilot/src/server/windows.rs
mpiton: @mpiton

> Pipe server death leaves the Windows instance file registered

Fixed in `37fb023` with that shape: borrow `cleanup` for `hold_windows`, `release()` after `accept_loop`. Tests still pass `None` and drop `RegistryGuard` in the task.

crates/tauri-pilot-cli/src/main.rs
mpiton: @mpiton

> Liveness probe follows /tmp symlinks past the UID ownership check

Fixed in `37fb023`. Candidate filter and `socket_has_listener` both `lstat`, require `is_socket()` and our uid, then connect. `test_resolve_socket_skips_symlink_to_live_socket` covers a planted `tauri-pilot-*.sock` link.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: @mpiton

> Late-hold GuardSlot::Released path is never exercised

Added `hold_after_release_drops_guard_immediately` in `37fb023`: `release_slot` then `hold_slot` a Drop spy, assert it dropped.

mpiton:
@coderabbit resolve

Issue #152, opened 2026-09-10
Second app instance panics at startup when the pilot socket is already bound


mpiton:
Launching a second instance of a Tauri app that embeds the plugin kills the host app before it ever shows a window.

## Reproduce

1. Start an app with `tauri_plugin_pilot::init()` registered (tested with `pilot-test-app`, identifier `com.pilot.test-app`).
2. Start a second instance of the same binary while the first is still running.

The second process aborts:

```
thread 'main' panicked at src/main.rs:47:10:
called `Result::unwrap()` on an `Err` value: PluginInitialization("pilot", "io error: socket already in use: /run/user/1000/tauri-pilot-com.pilot.test-app.sock")
```

## Cause

`crates/tauri-plugin-pilot/src/lib.rs:83` propagates the bind failure out of the `.setup()` hook with `?`:

```rust
let (listener, guard) = server::bind(&address).map_err(|e| {
    tracing::error!(?address, "failed to bind socket: {e}");
    e
})?;
```

An `Err` returned from `setup` becomes a `PluginInitialization` error, which Tauri surfaces as a fatal builder error. `server::bind` returns `AddrInUse` on purpose when a live server answers the probe (`crates/tauri-plugin-pilot/src/server/unix.rs:157-183`) — only a `ConnectionRefused` probe unlinks a stale socket and retries. That probe logic is correct; the problem is what the caller does with the error.

## Platform inconsistency

Windows binds the named pipe inside the spawned task rather than in `setup` (see the comment at `lib.rs:99-102`), so on Windows a second instance starts normally and simply has no pilot server. Unix and Windows disagree on whether a second instance can run at all.

## Expected

The plugin is debug-only tooling. A bind conflict should log a warning and skip starting the server, leaving the host app fully functional. Failing to start a QA server must never prevent the application under test from running.

PR #207, opened 2026-09-18
fix(plugin): serialize pathname bind with a sibling lock (#195)

### LINKED ISSUE #195: Two instances starting over a stale socket can remove each other's socket file
Follow-up to cubic's finding on #193.

`bind_pathname` replaces a stale socket in steps that nothing serializes across processes: `bind` fails with `AddrInUse`, `connect` gets `ConnectionRefused`, then `remove_file(path)` and a second `bind`. Two instances of the same app starting over the same stale file can interleave like this:

1. A and B both fail to bind, and both get `ConnectionRefused` from the stale socket.
2. A removes it and binds its own socket.
3. B's `remove_file` deletes A's new file, then B binds.

A ends up with a listener nobody can reach, since its file is gone. That much predates #193. B can also get `ConnectionRefused` from a socket that is about to go live, if its `connect` lands between A's `bind()` and `listen()` (std runs both inside `UnixListener::bind`), with the same result.

#193 adds a second effect. `SocketGuard::new` reads the inode from the path right after bind. If B's remove and rebind land between A's `bind` returning and that `metadata` call, A's guard records B's inode, and dropping it unlinks B's socket. The window is small, and today the guard never drops on a normal quit, so I don't expect anyone to hit this yet. That changes with #194, which drops the guard on `RunEvent::Exit`, so this fix should land first or together with it.

## Expected

Only one instance wins the path, and a guard never records another instance's inode.

## Fix ideas

Hold an exclusive `flock` for the length of `bind_pathname`, covering the stale probe, the remove, the rebind and the inode read. Not for the server's lifetime: a second instance has to get `AddrInUse` right away instead of blocking (#152). The open question is what to lock. A lock file next to the socket stays on disk forever. Locking the socket directory works for `$XDG_RUNTIME_DIR`, but any local user can lock `/tmp` and stall the bind, so the `/tmp` fallback would need `LOCK_NB` with a bounded retry.

Reading the bound file's inode from the listener instead of the path would also close the second effect, but I don't know a portable way to do it: `fstat` on the fd returns the sockfs inode (#165).

mpiton:
## Summary

- Serialize pathname socket bind with an exclusive flock on a sibling lock file so two instances replacing a stale socket cannot steal the path or record each other's inode.
- Release the lock before bind returns so a live second instance still gets "already in use" immediately (#152).

## Motivation

The stale-replace path (`AddrInUse` → `ConnectionRefused` → unlink → rebind → inode read) was unsynchronized. After #194 the guard drops on `RunEvent::Exit`, so a crossed inode would unlink a live peer on quit.

Closes #195

## Changes

- `bind_pathname` takes `LOCK_EX` on `{socket}.lock` for the whole bind, then drops it.
- The lock file may remain; leftover files are reused; create/lock failure aborts bind.
- Tests cover overlapping stale replace, live `AddrInUse`, leftover lock, lock released after bind, lock create failure, owner-only mode, and sibling lock teardown.

## Test Plan

## Checklist

---
## Summary by cubic
Fixes the stale socket bind race (#195) where two instances could unlink each other's file or record the wrong inode. Pathname bind is now serialized with an exclusive flock on a sibling lock file, released before bind returns so a live second instance still gets "already in use" immediately.

- The lock opens with `O_NOFOLLOW` and is validated as an owner-only regular file with a single hard link; leftover locks are reused and re-permissioned to owner-only.
- Lock creation or acquisition failure aborts bind. Tests cover overlapping stale replace, live in use, leftover lock, lock released, lock create failure, owner-only mode, and teardown.

<sup>Written for commit ab171d0e0fe351cdeeb4d0ed94c6633a33f6d839. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when replacing stale Unix sockets during application startup.
  * Prevented concurrent instances from interfering with each other’s socket files.
  * Preserved immediate errors when a socket is actively in use.
  * Improved cleanup of temporary socket-related files during startup and shutdown.

* **Documentation**
  * Added a changelog entry documenting the socket-handling fix and related issue.


crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+    let rc = unsafe {
+        // SAFETY: `file` is an open fd we own. `flock` does not take
+        // ownership of the fd or invalidate it. `LOCK_EX` is a valid
+        // operation for an open file.
+        libc::flock(file.as_raw_fd(), libc::LOCK_EX)
+    };
+    if rc != 0 {
+        return Err(std::io::Error::last_os_error());
+    }
+    Ok(BindLock { _file: file })
```
mpiton, line 218: > [!WARNING]
> **🟠 High · 🔒 Security** — Lock file open follows /tmp symlinks into victim files

`acquire_bind_lock` creates `{socket}.lock` with a following `open` and then `chmod`s that path. When the socket lives in `/tmp` (macOS default, Linux without a private `XDG_RUNTIME_DIR`), another local user can plant a symlink at the predictable `tauri-pilot-{identifier}.sock.lock` name pointing at any file the victim can write. Plugin `setup` then sets that file to `0o600` and blocks in `LOCK_EX` on the same inode, so the host app hangs in `setup` for as long as the attacker holds a flock on the target (bind errors are only skipped after `bind` returns).

**Fix:** Open with `custom_flags(libc::O_NOFOLLOW)`, `fstat` the fd and reject unless it is a regular file owned by the current uid with `nlink == 1`, then `file.set_permissions(0o600)` (fchmod) and `file.lock()?` instead of path `set_permissions` plus `unsafe` `libc::flock`. Fail the bind if those checks fail.

```suggestion
    use std::fs::OpenOptions;
    use std::os::unix::fs::{MetadataExt, OpenOptionsExt};

    let path = bind_lock_path(socket_path);
    let file = OpenOptions::new()
        .read(true)
        .write(true)
        .create(true)
        // Reuse a leftover lock file; flock is on the inode, not the contents.
        .truncate(false)
        .mode(0o600)
        .custom_flags(libc::O_NOFOLLOW)
        .open(&path)?;
    let meta = file.metadata()?;
    // SAFETY: getuid() has no preconditions.
    let my_uid = unsafe { libc::getuid() };
    if !meta.is_file() || meta.uid() != my_uid || meta.nlink() != 1 {
        return Err(std::io::Error::new(
            std::io::ErrorKind::PermissionDenied,
            "bind lock must be a regular file owned by this user",
        ));
    }
    file.set_permissions(std::fs::Permissions::from_mode(0o600))?;
    file.lock()?;
    Ok(BindLock { _file: file })
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the lock-file open and chmod-by-path are added by this diff
- **References:** CWE-59, CWE-367

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.93</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+        .truncate(false)
+        .mode(0o600)
+        .open(&path)?;
+    std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600))?;
+    let rc = unsafe {
+        // SAFETY: `file` is an open fd we own. `flock` does not take
+        // ownership of the fd or invalidate it. `LOCK_EX` is a valid
+        // operation for an open file.
+        libc::flock(file.as_raw_fd(), libc::LOCK_EX)
+    };
```
> **🟡 Medium · 🧱 Maintainability** — Unsafe libc::flock reimplements std::fs::File::lock

`acquire_bind_lock` takes `LOCK_EX` through `unsafe { libc::flock(...) }`. Workspace MSRV is 1.95; `std::fs::File::lock` has been stable since 1.89 and is flock on Unix, released when the `File` is closed — the same invariant `BindLock` already documents. The SAFETY comment, `AsRawFd` import, and the duplicate `LOCK_EX|LOCK_NB` block in `bind_releases_sibling_lock_before_returning` are an FFI layer the standard library already owns.

**Fix:** After a successful open, call `file.lock()?`. In `bind_releases_sibling_lock_before_returning`, replace `flock(..., LOCK_EX | LOCK_NB)` with `lock.try_lock()`. If `acquire_bind_lock` also gains `O_NOFOLLOW` and fd checks, fold `file.lock()?` into that change rather than applying a flock-only rewrite.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff introduces the unsafe flock wrapper
- **References:** https://doc.rust-lang.org/std/fs/struct.File.html#method.lock

</details>

<sub>🤖 <b>mpiton-review</b> · maintainability · confidence 0.93</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -370,6 +426,158 @@ mod tests {
...
+        let socket = unique_socket_path();
+        std::fs::create_dir(super::bind_lock_path(&socket)).expect("lock path is a directory");
+        let address = SocketAddr::from_pathname(&socket).expect("test socket address");
+        let result = bind(&address);
+        let _ = std::fs::remove_dir(super::bind_lock_path(&socket));
+        cleanup_bind_files(&socket);
+        assert!(
+            result.is_err(),
+            "must not bind if the sibling lock cannot be created"
+        );
```
mpiton, line 574: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Lock-create failure test ignores a leftover bound socket

`bind_fails_when_sibling_lock_cannot_be_created` only checks `result.is_err()`, and `cleanup_bind_files` runs first, so a bind that creates the pathname socket and then fails the sibling lock still passes. That is exactly the unlocked-bind order the #195 contract forbids.

**Fix:** Record `socket.exists()` before teardown, then assert both `result.is_err()` and that no socket file was left behind.

```suggestion
        let result = bind(&address);
        let bound = socket.exists();
        let _ = std::fs::remove_dir(super::bind_lock_path(&socket));
        cleanup_bind_files(&socket);
        assert!(
            result.is_err(),
            "must not bind if the sibling lock cannot be created"
        );
        assert!(!bound, "lock create failure must not leave a bound socket");
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff adds the lock-create failure test

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -370,6 +426,158 @@ mod tests {
...
+    #[test]
+    fn leftover_bind_lock_file_does_not_block_bind() {
+        let socket = unique_socket_path();
+        std::fs::write(super::bind_lock_path(&socket), b"").expect("plant leftover lock");
+        let address = SocketAddr::from_pathname(&socket).expect("test socket address");
+        let (listener, guard) = bind(&address).expect("bind with leftover lock file");
+        drop(guard);
+        drop(listener);
+        cleanup_bind_files(&socket);
+    }
```
mpiton, line 542: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Leftover lock reuse never checks owner-only mode

`OpenOptions::mode(0o600)` applies only when the lock file is created. Tightening a leftover file is `set_permissions(0o600)`, but `sibling_lock_file_is_owner_only` binds a fresh path and `leftover_bind_lock_file_does_not_block_bind` never reads the mode. Deleting `set_permissions` still passes both tests.

**Fix:** In `leftover_bind_lock_file_does_not_block_bind`, chmod the planted lock to `0o666`, bind, then assert the reused file is `0o600`.

```suggestion
    fn leftover_bind_lock_file_does_not_block_bind() {
        use std::os::unix::fs::PermissionsExt;

        let socket = unique_socket_path();
        let lock_path = super::bind_lock_path(&socket);
        std::fs::write(&lock_path, b"").expect("plant leftover lock");
        std::fs::set_permissions(&lock_path, std::fs::Permissions::from_mode(0o666))
            .expect("widen leftover lock");
        let address = SocketAddr::from_pathname(&socket).expect("test socket address");
        let (listener, guard) = bind(&address).expect("bind with leftover lock file");
        let mode = std::fs::metadata(&lock_path)
            .expect("lock metadata")
            .permissions()
            .mode()
            & 0o777;
        drop(guard);
        drop(listener);
        cleanup_bind_files(&socket);
        assert_eq!(
            mode, 0o600,
            "reused leftover lock must be owner-only, got {mode:#o}"
        );
    }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff adds leftover-lock reuse without asserting chmod

</details>

<sub>🤖 <b>mpiton-review</b> · tests · confidence 0.88</sub>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+/// The suffix is appended to the full socket pathname (`foo.sock.lock`) so
+/// it cannot collide with the socket file. The lock must not be the socket
+/// itself: `unlink` + rebind replaces that inode and would drop the flock
+/// mid-section. The socket directory is not used either (`/tmp` is
+/// world-writable).
+pub(crate) fn bind_lock_path(socket_path: &std::path::Path) -> std::path::PathBuf {
+    let mut lock = socket_path.as_os_str().to_owned();
+    lock.push(".lock");
+    std::path::PathBuf::from(lock)
+}
```
> **🔵 Low · 🧱 Maintainability** — bind_lock_path is pub(crate) only so lib.rs tests can rm leftovers

`bind_lock_path` is private to pathname bind except that `lib.rs` plugin tests now construct `{socket}.lock` themselves and `remove_file` it by hand. Nested `unix.rs` tests already have `cleanup_bind_files` and can see a private parent item; exporting the naming rule is a test-only leak of the lock-file contract into plugin setup tests. A later suffix change can leave stale lock files in `/tmp` without failing those tests.

**Fix:** Keep `bind_lock_path` private. Lift `cleanup_bind_files` next to it as `#[cfg(test)] pub(crate) fn cleanup_bind_files` and call that from the `#152`/`#194` tests in `lib.rs` instead of repeating `bind_lock_path` plus two `remove_file`s.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff exports bind_lock_path for plugin tests

</details>

CHANGELOG.md
```diff
@@ -34,6 +34,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Pathname socket bind serializes stale replacement with an exclusive
+  flock on a sibling lock file, so two instances starting over the same
+  leftover socket cannot unlink each other's file or record the wrong
+  inode. The lock is released before bind returns, so a live second
+  instance still gets "already in use" immediately. [#195]
```
> **🔵 Low · 📄 Docs** — CHANGELOG omits leftover `{socket}.lock` files after pathname bind

Pathname bind creates `{socket}.lock` in `acquire_bind_lock` and never unlinks it: `BindLock` only closes the flock fd, and `SocketGuard::drop` removes only the socket. After a normal quit the socket is gone but that regular file stays in `$XDG_RUNTIME_DIR` or `/tmp`. The new #195 bullet says the lock is released before bind returns, so a leftover `.lock` can look like a held lock with no record that the file is expected, unlocked, and reused.

**Fix:** In the #195 bullet, say the sibling `{socket}.lock` file may remain and is reused on the next bind (flock is released; the file is not unlinked). Add the same sentence to the `bind` rustdoc, which currently says only that the lock is released before return.

```suggestion
  inode. The flock is released before bind returns, so a live second
  instance still gets "already in use" immediately. The sibling
  `{socket}.lock` file may remain and is reused on the next bind. [#195]
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff adds the changelog bullet and the leftover lock file

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.86</sub>

mpiton:
## 🧨 mpiton-review — PR #207

> [!CAUTION]
> 🚫 **Changes requested** — 1 blocking issue (0 critical, 1 high) out of 6 findings.

> GitHub refused a `REQUEST_CHANGES` review on this PR (the author cannot request changes on their own pull request). Inline comments were posted individually; this is the ranked summary.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/server/unix.rs` | 0 | 1 | 3 | 1 |

### 🎯 Start here

1. 🟠 **Lock file open follows /tmp symlinks into victim files** — `crates/tauri-plugin-pilot/src/server/unix.rs:209`
2. 🟡 **Unsafe libc::flock reimplements std::fs::File::lock** — `crates/tauri-plugin-pilot/src/server/unix.rs:205`
3. 🟡 **Lock-create failure test ignores a leftover bound socket** — `crates/tauri-plugin-pilot/src/server/unix.rs:560`
4. 🟡 **Leftover lock reuse never checks owner-only mode** — `crates/tauri-plugin-pilot/src/server/unix.rs:524`
5. 🔵 **bind_lock_path is pub(crate) only so lib.rs tests can rm leftovers** — `crates/tauri-plugin-pilot/src/server/unix.rs:183`

<details>

| | |
|:---|:---|
| Commit reviewed | `2a4be2b62c` |
| Base | `main` |
| Files changed | 3 (+234/-10) |
| Languages | rust |

</details>

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+    std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600))?;
+    let rc = unsafe {
+        // SAFETY: `file` is an open fd we own. `flock` does not take
+        // ownership of the fd or invalidate it. `LOCK_EX` is a valid
+        // operation for an open file.
+        libc::flock(file.as_raw_fd(), libc::LOCK_EX)
+    };
+    if rc != 0 {
+        return Err(std::io::Error::last_os_error());
+    }
```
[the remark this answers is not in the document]
mpiton: @coderabbitai

> Use a private per-user lock directory, or enforce no-follow and ownership validation; apply permissions only when this process creates the file. Replace blocking LOCK_EX with LOCK_NB and a bounded retry window.

Symlink / ownership / chmod-by-path: fixed in `ab171d0`. Open uses `O_NOFOLLOW`, `fstat` rejects anything that isn't a regular file owned by this uid with `nlink == 1`, and mode is set on the fd.

Blocking `File::lock()` stays. #195 serializes two starters over a stale socket so the loser gets AddrInUse. `LOCK_NB` would turn that into a lock error, and `overlapping_stale_replaces_leave_one_reachable_winner` relies on the wait. The flock only covers bind; leftover flock dies with the process. A live second instance after that still fails immediately (#152).

Would only take `LOCK_NB` if we dropped the serialize-then-AddrInUse contract.

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+        // Reuse a leftover lock file; flock is on the inode, not the contents.
+        .truncate(false)
+        .mode(0o600)
+        .open(&path)?;
+    std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600))?;
+    let rc = unsafe {
+        // SAFETY: `file` is an open fd we own. `flock` does not take
+        // ownership of the fd or invalidate it. `LOCK_EX` is a valid
+        // operation for an open file.
+        libc::flock(file.as_raw_fd(), libc::LOCK_EX)
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> this blocking `flock` can wait indefinitely, so `bind_pathname` never returns

Not switching to `LOCK_NB`. The exclusive flock is the #195 critical section: two overlapping stale replaces take turns, then the loser gets already-in-use. A leftover lock file with no holder does not block; the kernel drops flock on exit. After bind returns the lock is released, so a live second instance still fails immediately (`live_socket_is_already_in_use_without_waiting`).

A hang would mean another process of this user held that inode for the whole bind, which is the same local DoS as holding the socket itself.

crates/tauri-plugin-pilot/src/server/unix.rs
```diff
@@ -158,15 +162,67 @@ fn bind_abstract(address: &SocketAddr) -> Result<std::os::unix::net::UnixListene
...
+    let path = bind_lock_path(socket_path);
+    let file = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        // Reuse a leftover lock file; flock is on the inode, not the contents.
+        .truncate(false)
+        .mode(0o600)
+        .open(&path)?;
+    std::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600))?;
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> `set_permissions(&path, ...)` follows it and changes the target's mode

Fixed in `ab171d0`. Open uses `O_NOFOLLOW` and `file.set_permissions` (fchmod) on that fd.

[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> another local user can pre-create the sibling lock as a symlink and prevent the pilot server from starting

Fixed in `ab171d0`. Open uses `O_NOFOLLOW`, then `fstat` requires a regular file owned by this uid with `nlink == 1`. Bind fails if those checks fail.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: > Open with `custom_flags(libc::O_NOFOLLOW)`, `fstat` the fd and reject unless it is a regular file owned by the current uid with `nlink == 1`, then `file.set_permissions(0o600)` (fchmod) and `file.lock()?`

Fixed in `ab171d0`. `acquire_bind_lock` now opens with `O_NOFOLLOW`, rejects non-regular / foreign / hardlinked inodes, fchmods the fd, and uses `File::lock`.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: > After a successful open, call `file.lock()?`. In `bind_releases_sibling_lock_before_returning`, replace `flock(..., LOCK_EX | LOCK_NB)` with `lock.try_lock()`.

Fixed in `ab171d0`. `acquire_bind_lock` calls `file.lock()?`; the release test uses `try_lock()`.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: > Record `socket.exists()` before teardown, then assert both `result.is_err()` and that no socket file was left behind.

Fixed in `ab171d0`. `bind_fails_when_sibling_lock_cannot_be_created` now records `socket.exists()` before teardown and asserts the path was not bound.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: > In `leftover_bind_lock_file_does_not_block_bind`, chmod the planted lock to `0o666`, bind, then assert the reused file is `0o600`.

Fixed in `ab171d0`. That test now plants `0o666` and asserts `0o600` after bind.

crates/tauri-plugin-pilot/src/server/unix.rs
mpiton: > Keep `bind_lock_path` private. Lift `cleanup_bind_files` next to it as `#[cfg(test)] pub(crate) fn cleanup_bind_files` and call that from the `#152`/`#194` tests in `lib.rs`.

Fixed in `ab171d0`. `bind_lock_path` is private; `cleanup_bind_files` is `cfg(test) pub(crate)` and used from those plugin tests.

CHANGELOG.md
mpiton: > In the #195 bullet, say the sibling `{socket}.lock` file may remain and is reused on the next bind. Add the same sentence to the `bind` rustdoc.

Fixed in `ab171d0`. Changelog and `bind` rustdoc both say the sibling `{socket}.lock` file may remain and is reused.

mpiton:
@coderabbit resolve

PR #224, opened 2026-09-19
fix(cli): exit quietly when stdout is closed

### LINKED ISSUE #213: CLI panics with 'failed printing to stdout: Broken pipe' when output is piped to head
The CLI panics when stdout is closed early, for example when piped into `head`:

```
$ tauri-pilot snapshot | head -1
  - heading "Pilot Test App" [ref=e1]

thread 'main' (1925226) panicked at .../library/std/src/io/stdio.rs:1166:9:
failed printing to stdout: Broken pipe (os error 32)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
$ echo ${PIPESTATUS[0]}
101
```

Found during end-to-end QA of 0.8.0 (Linux, CLI 0.7.3, commit 61ba190). Any command with long output does this (`snapshot`, `logs`, `network`, `html`...). A script using `set -o pipefail` sees a failure, and the backtrace hint is noise.

## Cause

Rust ignores SIGPIPE by default, so a write to a closed pipe returns `EPIPE`, and `println!` panics on it.

## Expected

Exit quietly (status 0, or 141 like other Unix tools) when stdout is a closed pipe. Two options: write output through a `writeln!` path that treats `ErrorKind::BrokenPipe` as a clean exit, or restore the default SIGPIPE handler at startup. The project forbids `unsafe`, so that rules out `libc::signal`. That leaves the `writeln!` route, or a crate that already wraps it.

mpiton:
Fixes #213

When stdout is a closed pipe (e.g. `tauri-pilot snapshot | head -1`), every write fails with EPIPE. Rust ignores SIGPIPE by default, so `println!` panics instead of exiting cleanly.

**Changes:**
- Add `out!` and `outln!` macros with `write_stdout()` backing
- On BrokenPipe, exit with status 0 (allows `set -o pipefail` scripts to pass)
- Other write errors still panic as expected (preserves safety)
- Add regression test `broken_pipe_exit` against closed pipe
- Update CHANGELOG

**Testing:**
- New integration test closes stdout and verifies exit 0
- `cargo test --workspace` passes (457 tests)
- `cargo clippy --all-targets -D warnings` passes

---
## Summary by cubic
Fixes #213: the CLI panicked with "failed printing to stdout: Broken pipe" when stdout was a closed pipe, like `tauri-pilot snapshot | head -1`. It now exits quietly with status 0, so `set -o pipefail` scripts stay green; other write errors still panic as before.

**Details**
- All CLI stdout writes go through `out!`/`outln!`, and `clippy::print_stdout` is denied so a bare `println!` can't be reintroduced.
- `storage get` on a missing key still exits 1 even when stdout is closed.
- Adds regression tests for closed-pipe exit 0, missing-key exit 1, and non-EPIPE write errors still panicking.
- Updates CHANGELOG.

<sup>Written for commit af9b9bed28627918b3a542b2725e9d218a43c317. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/224?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - CLI commands now exit successfully when standard output is closed, avoiding broken-pipe panics.
  - Missing `storage get` keys continue to exit with status 1, even when output is unavailable.
  - Other standard-output write failures continue to be reported as errors.
  - Output behavior remains consistent across snapshots, logs, network results, records, watch mode, and window details.

- **Documentation**
  - Updated the changelog with the broken-pipe behavior and related issue reference.


crates/tauri-pilot-cli/src/output.rs
```diff
@@ -3,9 +3,44 @@ use std::time::Duration;
...
+/// rather than 141 keeps `set -o pipefail` scripts green, like ripgrep does.
+///
+/// # Panics
+///
+/// Panics on any other write error, as `print!` does, so a full disk or a
+/// revoked terminal still fails loudly instead of losing output silently.
+pub(crate) fn write_stdout(args: std::fmt::Arguments<'_>) {
+    if let Err(e) = std::io::Write::write_fmt(&mut std::io::stdout(), args) {
+        if e.kind() == std::io::ErrorKind::BrokenPipe {
+            std::process::exit(0);
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — Broken-pipe exit(0) overrides the documented exit 1 of `storage get --json` for a missing key

main.rs:145-147 prints the `{"found": false}` payload through `format_json` -> `outln!` and only then calls `std::process::exit(1)`. If the reader has already closed the pipe, `write_stdout` exits 0 first. The pretty JSON goes out in two writes (LineWriter flushes up to the last newline, then `}\n`), so `storage get k --json | head -1` under `pipefail` can hit it. A probe against a mock server returning `{"found": false}` with stdout closed exits 0; before this PR it exited 101, and with stdout open it exits 1 as `storage_get_missing_key_json_exits_1` requires. docs/src/content/docs/reference/cli.md:1237 documents exit 1 for this case, so a script now reads a missing key as present.

**Fix:** Keep the status the command already decided. Simplest: a `static PIPE_EXIT: AtomicI32` (default 0) in output.rs that `write_stdout` passes to `std::process::exit`, set to 1 in main.rs before `format_result` when the kind is `StorageGet` and `found == false`. Alternative, as ripgrep does it: return the `io::Error` up to `main` and map `BrokenPipe` to an exit code in one place. Add a `storage get --json` / `{"found": false}` case to broken_pipe_exit.rs that asserts exit 1.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** The unconditional `std::process::exit(0)` on `BrokenPipe` is added by this diff and now runs before the existing missing-key `exit(1)` in main.rs:146.

</details>

mpiton (commented):
## 🧨 mpiton-review — PR #224

> [!IMPORTANT]

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/src/output.rs` | 0 | 0 | 1 | 2 |
| `crates/tauri-pilot-cli/tests/broken_pipe_exit.rs` | 0 | 0 | 0 | 2 |
| `crates/tauri-pilot-cli/src/main.rs` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟡 **Broken-pipe exit(0) overrides the documented exit 1 of `storage get --json` for a missing key** — `crates/tauri-pilot-cli/src/output.rs:35`
2. 🟡 **Nothing stops a new println! from bringing back the #213 panic** — `crates/tauri-pilot-cli/src/main.rs:26`
3. 🔵 **Unit-test output from `format_*` now bypasses libtest capture** — `crates/tauri-pilot-cli/src/output.rs:33`
4. 🔵 **Sixth hand-copied mock JSON-RPC server across the CLI integration tests** — `crates/tauri-pilot-cli/tests/broken_pipe_exit.rs:48`
5. 🔵 **The panic branch for write errors other than a broken pipe has no test** — `crates/tauri-pilot-cli/src/output.rs:37`

<details>

| | |
|:---|:---|
| Commit reviewed | `f6d1bdc0c8` |
| Base | `main` |
| Files changed | 4 (+174/-50) |
| Languages | rust |

</details>

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -23,6 +23,7 @@ use cli::{
     parse_target,
 };
 use client::Client;
+use output::{out, outln};
```
mpiton, line 30: > [!IMPORTANT]
> **🟡 Medium · 🧱 Maintainability** — Nothing stops a new println! from bringing back the #213 panic

The fix only holds while every stdout write in the crate goes through `out!`/`outln!`, and nothing in the build checks that rule. The next `println!` added to main.rs, scenario.rs or a new formatter will panic again with exit 101 when stdout is a closed pipe. CI stays green because `broken_pipe_exit.rs` only runs `snapshot`.

**Fix:** Put `#![deny(clippy::print_stdout)]` on line 1 of crates/tauri-pilot-cli/src/main.rs. The crate has no `print!`/`println!` left and `write_stdout` calls `Write::write_fmt` directly, so it compiles as-is. CI's `cargo clippy --workspace --all-targets -- -D warnings` then rejects any new `println!`. It also guards `mcp` mode, where stdout carries only JSON-RPC. A `crates/tauri-pilot-cli/clippy.toml` with `disallowed-macros` for `std::print`/`std::println` works too and lets the message point at `out!`/`outln!`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR adds the rule that every stdout write goes through `out!`/`outln!`, so the lint enforcing it belongs in the same change. It goes on main.rs line 1, next to the new `use output::{out, outln};` import.

</details>

crates/tauri-pilot-cli/src/output.rs
```diff
@@ -3,9 +3,44 @@ use std::time::Duration;
...
+/// Rust ignores `SIGPIPE`, so once the reader is gone (`tauri-pilot snapshot
+/// | head -1`) a write fails with `EPIPE` and `print!` panics. Exiting with 0
+/// rather than 141 keeps `set -o pipefail` scripts green, like ripgrep does.
+///
+/// # Panics
+///
+/// Panics on any other write error, as `print!` does, so a full disk or a
+/// revoked terminal still fails loudly instead of losing output silently.
+pub(crate) fn write_stdout(args: std::fmt::Arguments<'_>) {
+    if let Err(e) = std::io::Write::write_fmt(&mut std::io::stdout(), args) {
```
mpiton, line 55: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Unit-test output from `format_*` now bypasses libtest capture

libtest only captures output that goes through `print!`/`println!` (`std::io::_print` checks the capture buffer). A direct `std::io::stdout().write_fmt` skips it. About 40 unit tests in output.rs call `format_text`, `format_snapshot`, `format_diff`, `format_forms`, `format_windows` and so on. Every run now dumps their snapshot trees, form dumps and ANSI-coloured status lines into the `cargo test` output, even for passing tests, and parallel test threads interleave that text. If the test binary's stdout is a closed pipe, one of these calls also runs `process::exit(0)` and ends the whole test binary early with success.

**Fix:** In `write_stdout`, add a `#[cfg(test)]` branch that calls `print!("{args}")` and returns, so the harness captures unit-test output again. The broken-pipe path is still covered by tests/broken_pipe_exit.rs, which runs the real binary.

```suggestion
    if cfg!(test) {
        print!("{args}");
        return;
    }
    if let Err(e) = std::io::Write::write_fmt(&mut std::io::stdout(), args) {
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** routing every `println!` in output.rs through `write_stdout` is what removes the capture

</details>

<sub>🤖 <b>mpiton-review</b> · 3 reviewers agree: product-logic, quality, tests · confidence 1.0</sub>

crates/tauri-pilot-cli/tests/broken_pipe_exit.rs
```diff
@@ -0,0 +1,80 @@
...
+                    {"depth": 0, "role": "heading", "name": "Pilot Test App", "ref": "e1"},
+                    {"depth": 0, "role": "button", "name": "Click", "ref": "e2"}
+                ]
+            }
+        });
+        let mut bytes = serde_json::to_vec(&resp).expect("serialize");
+        bytes.push(b'\n');
+        writer.write_all(&bytes).expect("write");
+        writer.flush().expect("flush");
+    });
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Sixth hand-copied mock JSON-RPC server across the CLI integration tests

This block is almost a line-for-line copy of `spawn_mock_snapshot_server` in snapshot_save_json.rs. storage_get_exit.rs, storage_get_scenario.rs, rpc_error_data.rs and record_stop_exit.rs each carry their own copy too. The copies already differ: this one hardcodes `/tmp/tauri-pilot-it-epipe-{pid}.sock` instead of using `unique_socket_path`, and waits with `join()` after the binary exits, while three other files switched to `recv_timeout`. Any fix to the harness means editing six files.

**Fix:** Move `unique_socket_path` and a `spawn_mock_server(socket, result, gate: Option<mpsc::Receiver<()>>)` helper into crates/tauri-pilot-cli/tests/common/mod.rs, with the `recv_timeout` completion signal. Use it here: the `gate` receiver covers the wait-until-stdout-closes step. Move the other five files onto it in a follow-up.

crates/tauri-pilot-cli/src/output.rs
```diff
@@ -3,9 +3,44 @@ use std::time::Duration;
...
+        if e.kind() == std::io::ErrorKind::BrokenPipe {
+            std::process::exit(0);
+        }
+        panic!("failed printing to stdout: {e}");
```
mpiton, line 59: > [!NOTE]
> **🔵 Low · 🧪 Tests** — The panic branch for write errors other than a broken pipe has no test

The CHANGELOG and the `# Panics` doc both promise that other write errors still fail loudly, but no test covers that. If someone later loosens `write_stdout` to exit 0 on any error, a full disk would lose output silently and CI would stay green. The only new test covers `BrokenPipe`, and nothing in the crate references `/dev/full` or `failed printing`.

**Fix:** Add a `#[cfg(target_os = "linux")]` test next to the existing one in `tests/broken_pipe_exit.rs`. It runs `snapshot` against the same mock server with stdout set to `OpenOptions::new().write(true).open("/dev/full")`, then asserts `status.code() == Some(101)` and that stderr contains `failed printing to stdout`. Writes to `/dev/full` fail with ENOSPC, not EPIPE.

crates/tauri-pilot-cli/tests/broken_pipe_exit.rs
```diff
@@ -0,0 +1,80 @@
...
+    // Fail fast before join() so a binary that exits before connecting cannot
+    // leave the mock server blocked on accept() and hang the test.
+    if !output.status.success() {
+        let _ = std::fs::remove_file(&socket);
+        panic!(
+            "expected exit 0 on a closed stdout pipe, got {:?}: stderr={stderr}",
+            output.status.code()
+        );
+    }
+    server.join().expect("mock server join");
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — server.join() hangs forever if the binary exits 0 without connecting

The fail-fast guard only runs on a non-zero exit. The comment above it claims it covers any binary that exits before connecting. Suppose a future change writes to stdout before `Client::connect`, for example a notice. That write hits the closed pipe, exits 0, and leaves the mock server blocked in `accept()`, so `join()` hangs the CI job instead of failing. The sibling tests (`storage_get_exit.rs`, `record_stop_exit.rs`) use `recv_timeout(SERVER_DONE_TIMEOUT)` for exactly this case.

**Fix:** Have the server thread send on a `done_tx` channel after `flush()`. Replace `server.join()` with `done_rx.recv_timeout(SERVER_DONE_TIMEOUT)` and panic with stderr on `Err`, as the sibling tests do. Reword the comment at lines 65-66 so it no longer says every early exit is covered.

PR #225, opened 2026-09-19
fix(cli): report oversized requests and id:null errors (#214)

### LINKED ISSUE #214: Requests over 1 MiB fail with 'Response ID mismatch' instead of the server's size error
When a request is larger than the plugin's 1 MiB line limit, the server answers with a clear error, but the CLI throws it away. What reaches the user is an id mismatch or a broken pipe:

```
$ python3 -c "print(\"'\" + 'a'*1100000 + \"'.length\")" | tauri-pilot eval -
Error: Response ID mismatch: expected 1, got null

$ # same command, second run
Error: Broken pipe (os error 32)
```

Found during end-to-end QA of 0.8.0 (Linux, plugin and CLI 0.7.3, commit 61ba190). I hit it for real when injecting a React development build (1.19 MB) with `eval -`. The minified build (143 KB) works.

## What the server sends

A raw client gets the real reason:

```
{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Request line exceeds maximum length"}}
```

`MAX_LINE_LENGTH` is `1_048_576` (`crates/tauri-plugin-pilot/src/server/mod.rs:19`). The server replies with `id: null`, as JSON-RPC 2.0 says it should when it can't read the request id. Then it closes the connection. That's why the CLI sometimes fails on its own write first.

## Expected

- The CLI should show an error response with `id: null` as the error itself, not as an id mismatch. The spec uses `id: null` exactly for parse errors.
- Better still, check the serialized request size in the CLI before sending it, and fail with something like `request is 1.1 MB, the plugin accepts at most 1 MiB`. That also covers the broken-pipe case.
- Mention the limit in the `eval` docs.

mpiton:
## Summary

- The CLI refuses a request longer than the plugin's 1 MiB line limit before sending it: `<method> request is N bytes; the plugin accepts at most 1048576 bytes`.
- A JSON-RPC error whose `id` is `null` is reported as that error, not as `Response ID mismatch`.
- `drop` checks its files against the same limit before reading them.

## Motivation

Closes #214

A request over 1 MiB, such as a React development build sent with `eval -`, gets a `-32700` error with `"id": null` from the plugin, which then closes the connection. The CLI threw that error away and failed with `Response ID mismatch: expected 1, got null`, or with `Broken pipe (os error 32)` when the plugin closed the socket before the CLI finished writing.

`drop` hit the same wall: it allowed 50 MB per file and 100 MB of base64 in total, yet anything past about 768 KiB of file data could never fit in one request.

## Changes

- `Client::call` (`crates/tauri-pilot-cli/src/client/mod.rs`)
  - New `MAX_REQUEST_LEN` (1,048,576 bytes, newline included) mirrors the plugin's `MAX_LINE_LENGTH`. A larger request fails before anything is written, so the broken pipe goes away too.
  - A response with `"id": null` and an `error` skips the id check, so the plugin's error reaches the user. A null id without an error is still a mismatch.
- `run_drop_command` (`crates/tauri-pilot-cli/src/main.rs`) sizes the whole request before reading any file: the envelope through the new `Client::request_len`, which shares `call`'s encoder, plus each file's base64 length from metadata. A drop that cannot fit in one request is refused up front, and each read stops at what the room left can encode, so a file larger than its metadata said (still being written, or a `/proc` file) is refused before it is encoded. This replaces `MAX_DROP_FILE_SIZE` and `MAX_TOTAL_DROP_SIZE`. `diff --ref` keeps its 50 MiB check, as a memory guard only, and its error names that threshold.
- Plugin (`crates/tauri-plugin-pilot/src/server/mod.rs`): a comment on `MAX_LINE_LENGTH` points at the CLI copy. No behavior change.
- Tests
  - CLI: oversized request refused; request of exactly 1 MiB still sent; `id: null` error reported as the error; `id: null` without `error` still a mismatch; `drop` refused before reading when the request around the files pushes it over; `drop` request of exactly 1 MiB still sent; `drop` of a `/proc` file that reports 0 bytes refused once read (Linux only); `diff --ref` error names the 50 MiB guard.
  - Plugin: `test_server_reads_lines_up_to_one_mib` pins the limit. A 1,048,576-byte line gets an answer; a 1,048,577-byte line gets `-32700` with `id: null`, then EOF.
- Docs: CLI reference (`eval`, `drop`, and the limit in the JSON-RPC protocol section), `SKILL.md`, CHANGELOG.

## Test Plan

- [x] `cargo test --workspace` passes (468 tests)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] Tested manually with a Tauri app (if applicable): not run. The plugin test drives the real server over a Unix socket; the CLI tests use a mock server that sends the plugin's exact reply.

`cargo fmt --all -- --check` is clean. Mutants that fail the new tests: `>` to `>=` on the size check, `error.is_some()` removed, id check removed, `<=` to `<` on the drop check, envelope left out of the drop check, read-loop room check removed, `+ 1` dropped from the read cap (would truncate silently), plugin limit raised to 2 MiB.

## Checklist

- [x] All new files are under 150 lines (no new files)

---
## Summary by cubic
Fixes #214 by refusing requests over the plugin's 1 MiB line limit before sending them and reporting `id: null` errors as the actual RPC error instead of `Response ID mismatch` or `Broken pipe`.

- Applies the same limit to `drop` before reading files: files travel base64-encoded, so one drop must total a little under 768 KiB, not the 50 MB per file the docs promised.
- Keeps the boundary inclusive: an exactly 1 MiB request is still sent.
- Reads dropped files only up to the space left in the request, so a file that outgrows its metadata is refused instead of sent oversized.
- Adds regression tests for oversized and boundary requests, `id: null` responses, and the `drop` limit, and documents the limit in the CLI reference, eval docs, `SKILL.md`, and CHANGELOG.

<sup>Written for commit 1761aa569ace5ab5798565dcfdcb77b85424ee8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/225?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI requests exceeding the 1 MiB JSON-RPC request limit are rejected before transmission with a clear size-related error.
  * Server requests over the 1 MiB line limit return a parse error and close the connection.
  * File-drop requests enforce the 1 MiB encoded request limit, including base64 overhead.
  * Error responses with null request IDs are handled correctly, while successful null-ID responses remain mismatches.

* **Documentation**
  * Documented request-size limits, oversized-request errors, split evaluation behavior, and recommendations to use minified scripts.


crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -4,6 +4,13 @@ use anyhow::{Result, bail};
 use std::path::Path;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 
+/// Longest request line the plugin reads, trailing newline included.
+///
+/// Mirrors `MAX_LINE_LENGTH` in the plugin's `server/mod.rs`, which answers a
+/// longer line with an error and closes the connection, often while the CLI
+/// is still writing. The crates ship separately, so change both together.
+const MAX_REQUEST_LEN: usize = 1_048_576;
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — `drop` and `diff --ref` still accept 50 MB files that the new 1 MiB cap refuses

`Client::call` now refuses any request over `MAX_REQUEST_LEN`, and that covers every command, not only `eval`. `drop` sends its files base64-encoded in one request, so the files in one drop can total a little under 768 KiB. `run_drop_command` still checks them against `MAX_DROP_FILE_SIZE` (50 MB) and `MAX_TOTAL_DROP_SIZE` (100 MB) at main.rs:977-978, `diff --ref` still allows a 50 MB snapshot (main.rs:585-586), and cli.md:642 still says `50 MB per file, 100 MB total payload`. Against a mock socket, an 800,000-byte file is read and encoded in full, then fails with `drop request is 1066814 bytes; the plugin accepts at most 1048576 bytes (1 MiB)`, while a 700,000-byte file goes through. Before this PR the same drop died on the plugin's limit with a Broken pipe or an ID mismatch, so the new error is an improvement, but the docs and guards now contradict it.

**Fix:** Make `MAX_REQUEST_LEN` `pub(crate)` and check the drop files against it from their metadata, before any file is read: base64 turns 3 bytes into 4, so the raw total must stay under about three quarters of the limit. That replaces `MAX_DROP_FILE_SIZE`, `MAX_TOTAL_DROP_SIZE` and their `File too large (>50 MB)` / `Total drop payload exceeds 100 MB limit` messages. Change cli.md:642 to the real limit, e.g. `**Limits:** one request is at most 1 MiB and files are sent base64-encoded, so the files in one drop can total a little under 768 KiB.` The `diff --ref` stat check can stay as a memory guard, but its `(>50 MB)` message should not suggest such a snapshot can be sent. The 1 MiB paragraph added under `eval` also belongs in `## JSON-RPC Protocol` (cli.md:1424), since the limit applies to every command.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR makes `Client::call` refuse every request over 1 MiB, `drop` and `diff` included, but documents the limit only under `eval`; the `drop` limits on the same page now contradict it.

</details>

<sub>🤖 <b>mpiton-review</b> · 4 reviewers agree: documentation, maintainability, product-logic, quality · confidence 1.0</sub>

crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -253,6 +270,81 @@ mod tests {
...
+        let mut client = connect_with_retry(&socket).await;
+        let err = client.call("eval", None).await.expect_err("server error");
+        assert_eq!(
+            err.to_string(),
+            "RPC error (-32700): Request line exceeds maximum length"
+        );
+
+        handle.abort();
+        let _ = std::fs::remove_file(&socket);
+    }
```
mpiton, line 296: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The id-mismatch check has no test, so the null-id exemption can widen unnoticed

`test_client_null_id_error_is_reported_as_the_error` covers only the case the exemption exists for, and no test in the crate asserts `Response ID mismatch`. Dropping `&& response.error.is_some()` on line 78, or the whole check on line 79, keeps all seven client tests green (checked by mutation): a null-id `result`, or an error carrying another request's id, would then be returned as this call's answer.

**Fix:** Add a negative case next to the null-id test: the mock answers the first `call` with `{"jsonrpc":"2.0","id":null,"result":42}` and the second with a -32601 error for id 7, and the test asserts `Response ID mismatch: expected 1, got null` and `Response ID mismatch: expected 2, got 7`. The suggestion passes on this branch, fails when either condition on lines 78-79 is loosened, and is clean under `clippy --all-targets -D warnings` and rustfmt.

```suggestion
    }

    #[tokio::test]
    async fn test_client_null_id_exemption_needs_an_error() {
        // Only an error may come back with `"id": null` (#214): a null-id
        // result, or an error for another id, is still a mismatch.
        let socket = unique_socket_path("t05h");
        let _ = std::fs::remove_file(&socket);
        let listener = UnixListener::bind(&socket).expect("bind mock socket");
        let handle = tokio::spawn(async move {
            let (stream, _) = listener.accept().await.expect("accept");
            let (reader, mut writer) = stream.into_split();
            let mut reader = BufReader::new(reader);
            for raw in [
                r#"{"jsonrpc":"2.0","id":null,"result":42}"#,
                r#"{"jsonrpc":"2.0","id":7,"error":{"code":-32601,"message":"Method not found"}}"#,
            ] {
                let mut line = String::new();
                reader.read_line(&mut line).await.expect("read line");
                writer.write_all(raw.as_bytes()).await.expect("write raw");
                writer.write_all(b"\n").await.expect("write newline");
                writer.flush().await.expect("flush");
            }
        });

        let mut client = connect_with_retry(&socket).await;
        let err = client.call("eval", None).await.expect_err("null-id result");
        assert_eq!(
            err.to_string(),
            "Response ID mismatch: expected 1, got null"
        );
        let err = client.call("eval", None).await.expect_err("other id");
        assert_eq!(err.to_string(), "Response ID mismatch: expected 2, got 7");

        handle.abort();
        let _ = std::fs::remove_file(&socket);
    }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** Line 78 adds the exemption and line 79 changes the id check; the diff tests only the null-id error case.
- **References:** https://www.jsonrpc.org/specification#response_object

</details>

crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -4,6 +4,13 @@ use anyhow::{Result, bail};
 use std::path::Path;
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
 
+/// Longest request line the plugin reads, trailing newline included.
+///
+/// Mirrors `MAX_LINE_LENGTH` in the plugin's `server/mod.rs`, which answers a
+/// longer line with an error and closes the connection, often while the CLI
+/// is still writing. The crates ship separately, so change both together.
```
mpiton, line 11: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Plugin's `MAX_LINE_LENGTH` has no pointer back to this copy, and no test pins it

Only this side says to change both constants together. `MAX_LINE_LENGTH` at `crates/tauri-plugin-pilot/src/server/mod.rs:19` is a bare local const, and no plugin test sends a line of 1 MiB or more. If someone raises the plugin limit, for example to fit development builds, CI stays green while the CLI keeps refusing requests between the old and new limits, with an error that names the old one. The `(1 MiB)` literal in the `bail!` message on line 60 goes stale the same way if only `MAX_REQUEST_LEN` changes.

**Fix:** Add the mirror note on `MAX_LINE_LENGTH`, naming `MAX_REQUEST_LEN` in `crates/tauri-pilot-cli/src/client/mod.rs`, and drop the `(1 MiB)` literal from the `bail!` message or derive it from the constant. Optionally pin the plugin side in `server/unix.rs` next to `test_server_handles_invalid_json`, using `start_test_server`: a 1_048_576-byte line (newline included) is answered with its own id, and a 1_048_577-byte line gets `id: null`, -32700, then EOF.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR adds `MAX_REQUEST_LEN` as a copy of the plugin's limit and writes the change-both rule only on the copy.

</details>

<sub>🤖 <b>mpiton-review</b> · 3 reviewers agree: documentation, maintainability, tests · confidence 1.0</sub>

crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -253,6 +270,81 @@ mod tests {
...
+            let (stream, _) = listener.accept().await.expect("accept");
+            let (reader, mut writer) = stream.into_split();
+            let mut reader = BufReader::new(reader);
+            let mut line = String::new();
+            reader.read_line(&mut line).await.expect("read line");
+            let raw = r#"{"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Request line exceeds maximum length"}}"#;
+            writer.write_all(raw.as_bytes()).await.expect("write raw");
+            writer.write_all(b"\n").await.expect("write newline");
+            writer.flush().await.expect("flush");
+        });
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Null-id test pastes the reply-once mock server a third time

These lines repeat the remove_file/bind/accept/into_split/read_line/write/flush block from `test_client_null_result_is_success` and `test_client_missing_result_is_success` almost word for word. Only the reply line differs. With three copies, any change to the mock, such as an accept timeout, has to be made three times, and the next protocol regression test will likely add a fourth.

**Fix:** Add a helper next to `mock_server`: `fn reply_once(path: &Path, reply: impl FnOnce(&str) -> String + Send + 'static) -> tokio::task::JoinHandle<()>`. It binds, accepts one connection, reads one line and writes `reply(&line)` followed by a newline. The three tests then pass only their raw response; this one ignores the request line.

crates/tauri-pilot-cli/src/client/mod.rs
```diff
@@ -253,6 +270,81 @@ mod tests {
...
+        );
+
+        drop(client);
+        let received = handle.await.expect("mock server task");
+        assert_eq!(
+            received, 0,
+            "an oversized request must not reach the plugin"
+        );
+        let _ = std::fs::remove_file(&socket);
+    }
```
mpiton, line 367: > [!NOTE]
> **🔵 Low · 🧪 Tests** — No test sends a request of exactly 1 MiB, so an off-by-one in the size check passes

The plugin accepts a line of exactly 1_048_576 bytes, newline included (`take(MAX_LINE_LENGTH as u64 + 1)` then `line.len() > MAX_LINE_LENGTH`), but the only size test sends 1_048_577. Turning `>` into `>=` on line 58 keeps all seven client tests green (checked by mutation), and the CLI would then refuse a request the plugin takes.

**Fix:** Add the boundary case after `test_client_refuses_request_over_plugin_limit_before_sending`: a script of `1_048_576 - EVAL_ENVELOPE` bytes sent to `mock_server` must reach it and come back as its -32601. The suggestion passes on this branch, fails under `>=`, and is clean under `clippy --all-targets -D warnings` and rustfmt.

```suggestion
    }

    #[tokio::test]
    async fn test_client_sends_request_at_exactly_plugin_limit() {
        // The plugin still reads a line of exactly 1 MiB, newline included,
        // so that request must be sent: the mock answers `eval` with -32601.
        let socket = unique_socket_path("t05g");
        let handle = mock_server(&socket);

        let mut client = connect_with_retry(&socket).await;
        let script = "a".repeat(1_048_576 - EVAL_ENVELOPE);
        let err = client
            .call("eval", Some(serde_json::json!({ "script": script })))
            .await
            .expect_err("mock answers eval with Method not found");
        assert!(
            err.to_string().contains("-32601"),
            "a 1048576-byte request must reach the plugin, got: {err}"
        );

        handle.abort();
        let _ = std::fs::remove_file(&socket);
    }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** Line 58 adds the size check; the diff tests it only one byte past the limit.

</details>

docs/src/content/docs/reference/cli.md
```diff
@@ -822,6 +822,14 @@ Prefer `<<'EOF'` with quotes around the heredoc delimiter. It disables shell
 variable and command expansion, so `$` and backticks inside the script do not
 need escaping.
 
+The plugin reads at most 1 MiB (1,048,576 bytes) per request. The limit counts
+the JSON-RPC request the CLI sends, where quotes, backslashes and newlines in
+the script are escaped, so a script just under 1 MiB can still be too large. A
+larger request fails before it is sent, with
+`eval request is N bytes; the plugin accepts at most 1048576 bytes (1 MiB)`.
+To inject a library, use its minified build rather than the development one;
+otherwise split the script across several calls.
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — Advice to split the script across calls omits that `let`/`const` do not carry over

The bridge compiles each `eval` call on its own (`new Function` wrappers, then indirect `eval`, in `evalScript` at bridge.js:1414-1468). A chunk cut mid-function is a SyntaxError, and a top-level `let`/`const`/`class` from one call is undefined in the next. A development bundle, the case this sentence is about, is one UMD expression that cannot be split at all, and the next paragraph says `const` and `let` work, which leads readers to the exact pattern that breaks.

**Fix:** State that each call must be a complete script and that state passes between calls only through `window`, because `let`, `const` and `class` bindings stay local to one call.

```suggestion
To inject a library, use its minified build rather than the development one.
Otherwise split the script into calls that each parse on their own, and pass
state between them through `window`: `let`, `const` and `class` bindings do
not carry over to the next call.
```

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.75</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #225

> [!IMPORTANT]

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/src/client/mod.rs` | 0 | 0 | 2 | 3 |
| `docs/src/content/docs/reference/cli.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟡 **`drop` and `diff --ref` still accept 50 MB files that the new 1 MiB cap refuses** — `crates/tauri-pilot-cli/src/client/mod.rs:12`
2. 🟡 **The id-mismatch check has no test, so the null-id exemption can widen unnoticed** — `crates/tauri-pilot-cli/src/client/mod.rs:302`
3. 🔵 **Plugin's `MAX_LINE_LENGTH` has no pointer back to this copy, and no test pins it** — `crates/tauri-pilot-cli/src/client/mod.rs:11`
4. 🔵 **Null-id test pastes the reply-once mock server a third time** — `crates/tauri-pilot-cli/src/client/mod.rs:291`
5. 🔵 **No test sends a request of exactly 1 MiB, so an off-by-one in the size check passes** — `crates/tauri-pilot-cli/src/client/mod.rs:346`

<details>

| | |
|:---|:---|
| Commit reviewed | `4f558d5240` |
| Base | `main` |
| Files changed | 4 (+115/-2) |
| Languages | rust |

</details>

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -974,35 +976,32 @@ async fn run_storage_command(
...
         anyhow::ensure!(meta.is_file(), "Not a regular file: {}", path.display());
-        anyhow::ensure!(
-            meta.len() <= MAX_DROP_FILE_SIZE,
-            "File too large (>50 MB): {}",
-            path.display()
-        );
+        encoded_len = encoded_len.saturating_add(meta.len().div_ceil(3) * 4);
+    }
+    anyhow::ensure!(
+        encoded_len < MAX_REQUEST_LEN as u64,
```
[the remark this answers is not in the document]
mpiton: Valid that the pre-check leaves out the envelope, but that is on purpose: it only exists to avoid reading and encoding files that can never fit. The exact check is `Client::call` (`crates/tauri-pilot-cli/src/client/mod.rs:58`), which refuses the request before writing anything, with `drop request is N bytes; the plugin accepts at most 1048576 bytes`. A drop inside the envelope margin still never reaches the socket, it just gets that message instead. Budgeting exactly would mean copying the request encoding into `run_drop_command`, so leaving it as is.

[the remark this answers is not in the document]
mpiton: Changed course after CodeRabbit flagged the same case: 9f0ac1e sizes the whole request before reading any file. The envelope comes from the new `Client::request_len`, which uses the same encoder as `call`, and each file adds its base64 length from metadata. New tests cover a drop the envelope alone pushes over the limit and a drop request of exactly 1,048,576 bytes.

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -974,43 +976,43 @@ async fn run_storage_command(
...
+    let mut p = target_params(target);
     p["files"] = json!(files);
+    let envelope = client.request_len("drop", with_window(Some(p.clone()), window))?;
+    let request_len = encoded_len.saturating_add(envelope as u64);
+    anyhow::ensure!(
+        request_len <= MAX_REQUEST_LEN as u64,
+        "drop request would be {request_len} bytes once the files are base64-encoded; the plugin accepts at most {MAX_REQUEST_LEN} bytes"
+    );
+    for (entry, path) in files.iter_mut().zip(&file) {
+        let data = std::fs::read(path)
```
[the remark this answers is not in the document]
mpiton: Valid. Fixed in 1761aa5: the read loop now tracks the room left in the request and reads at most what that room can encode, plus one byte. A file that holds more than its metadata reported is refused before it is base64-encoded or sent. `test_drop_refuses_file_larger_than_reported` covers this with `/proc/self/status`, which reports 0 bytes.

PR #226, opened 2026-09-20
fix(cli): report screenshot path and reason in run results (#215)

### LINKED ISSUE #215: run and pilot.run save failure screenshots to the process cwd and never report the path
`run` and MCP `pilot.run` save failure screenshots to `./tauri-pilot-failures/`, relative to the process working directory. Over MCP that directory is whatever the client launched the server from. The JSON report doesn't include the path, and if the directory isn't writable the screenshot is silently dropped.

Found during end-to-end QA of 0.8.0 (Linux, plugin and CLI 0.7.3, commit 61ba190).

## Reproduce

Start the MCP server from a scratch directory:

```
# `tauri-pilot mcp` started with cwd /tmp/x, then tools/call:
pilot.run {"content":"[[step]]\naction = \"click\"\ntarget = \"#does-not-exist\"\n"}
{"result":{"duration_ms":229,"failed":1,"name":"unnamed scenario","ok":false,"passed":0,"skipped":0,
 "steps":[{"duration_ms":1,"message":"RPC error (-32603): Eval error: JavaScript error: No element matches selector: #does-not-exist","name":"step-1","status":"failed"}],
 "summary":"0 passed · 1 failed · 0 skipped  (0.229s)"}}

$ ls -R /tmp/x
tauri-pilot-failures/
step-1-1789827768941.png
```

The report doesn't mention the PNG. Only a stderr line (`failure screenshot → ...`) prints it, and an MCP client doesn't show stderr to the model.

Run the same thing with the server's cwd at `/`: nothing is written, and nothing says so. `take_failure_screenshot` fails on `create_dir_all` and `scenario.rs:221` discards the error with `let _ =`.

With the CLI, `tauri-pilot run` from a project root drops an untracked `tauri-pilot-failures/` into the repo. That happened to me in this repo during QA.

## Expected

- Put the screenshot path into the step result (for example `"screenshot": "/abs/path/step-1-....png"`), in both the MCP JSON report and `run --json` / `--junit` output.
- Let the caller choose the directory (`--screenshots-dir` on `run`, a `screenshots_dir` argument on `pilot.run`). For MCP, default to a temp or cache directory instead of the cwd.
- If the screenshot can't be saved, put the reason in the step result instead of dropping it.

mpiton:
## Summary

• Screenshots from failed steps now reported in results with absolute path or error reason
• `run --screenshots-dir <DIR>` picks directory (default: ./tauri-pilot-failures)
• `run --json` now prints JSON report (flag was silently ignored before)
• MCP `pilot.run` takes `screenshots_dir` argument (default: temp/tauri-pilot-failures)
• JUnit XML includes screenshot info in `<system-out>`
• stderr shows why screenshot save failed instead of silently dropping it

## Changes

- StepResult carries screenshot outcome as Option<Result<PathBuf, String>>
- Failed steps report `screenshot` (absolute path) or `screenshot_error` (why not saved)
- take_failure_screenshot now returns absolute path and takes configurable directory
- report_to_json and JUnit output carry screenshot data
- New integration test: CLI run with --screenshots-dir --json --junit

## Type

fix

---
## Summary by cubic
Fixes #215 by making failed `run` steps report their screenshot outcome and making `run --json` print the JSON report instead of silently ignoring the flag.

**Bug Fixes**
- Failed steps now carry the screenshot's absolute path in `screenshot`, or the reason it couldn't be saved in `screenshot_error`, in JSON, MCP, and JUnit XML (`<system-out>`).
- `--screenshots-dir` and MCP `pilot.run`'s `screenshots_dir` choose the directory; an empty value falls back to the default, and `run` keeps its `./tauri-pilot-failures` default.
- MCP defaults to an owner-only per-user `tauri-pilot-failures-<uid>` directory under `$XDG_RUNTIME_DIR` or the system temp dir, replacing the shared `/tmp` path and the server's working directory.
- `--screenshots-dir` accepts non-UTF-8 Unix paths.
- The JUnit XML is written before the JSON report, so `run --json | head -1` still exits 1 with the report on disk.
- The text summary always goes to stderr, and stderr explains why a screenshot save failed.

<sup>Written for commit a70571ac9c393dcb1d6486b6a46b22ad2bf2de6b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/226?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Failed scenario steps now capture screenshots and report their absolute paths in CLI, JSON, MCP, and JUnit results.
  * Screenshot storage can be customized; configurable defaults are provided for CLI and MCP runs.
  * Reports identify screenshot save failures when capture files cannot be written.

* **Documentation**
  * Updated CLI, scenario, contribution, and changelog documentation with screenshot reporting and configuration details.


crates/tauri-pilot-cli/src/main.rs
```diff
@@ -1471,10 +1476,23 @@ async fn run_scenario_command(
     };
 
     let fail_fast_override = if no_fail_fast { Some(false) } else { None };
-    let report = scenario::run_scenario(&mut client, &loaded, window, fail_fast_override).await?;
+    let screenshots_dir =
+        screenshots_dir.unwrap_or_else(|| Path::new(scenario::DEFAULT_SCREENSHOT_DIR));
```
mpiton: > [!WARNING]
> **🟠 High · 🐛 Correctness** — `Path` is imported under `#[cfg(unix)]`, so the new default breaks the Windows build

`use std::path::Path;` at main.rs:20-21 is gated by `#[cfg(unix)]` — the only other bare `Path::new` in this file (line 1522) sits inside a `#[cfg(unix)]` block. `run_scenario_command` is compiled on every target, so this line does not resolve on Windows. Reproduced locally: `cargo check --target x86_64-pc-windows-gnu -p tauri-pilot-cli` fails with `error[E0433]: cannot find type `Path` in this scope --> crates/tauri-pilot-cli/src/main.rs:1480:43`. CI runs the same crate on `windows-latest` (.github/workflows/ci.yml:33), so this leg will not compile.

**Fix:** Fully qualify it, like the other path types in this signature, or drop the `#[cfg(unix)]` gate on the import.

```suggestion
    let screenshots_dir = screenshots_dir
        .unwrap_or_else(|| std::path::Path::new(scenario::DEFAULT_SCREENSHOT_DIR));
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** both lines are added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.99</sub>

crates/tauri-pilot-cli/src/main.rs
```diff
@@ -1471,10 +1476,23 @@ async fn run_scenario_command(
...
+        fail_fast_override,
+        screenshots_dir,
+    )
+    .await?;
 
     scenario::print_report(&report);
 
+    if json {
+        output::format_json(&scenario::report_to_json(&report))?;
+    }
```
mpiton: > [!WARNING]
> **🟠 High · 🐛 Correctness** — `run --json` exits 0 on a failed scenario when stdout closes early

`output::format_json` goes through `write_stdout`, which on `EPIPE` calls `std::process::exit(BROKEN_PIPE_EXIT)` — 0 unless `set_broken_pipe_exit` ran first (output.rs:24-60). `run_scenario_command` never calls it, so `tauri-pilot run ci.toml --json | head -1` on a failing scenario exits 0: the `std::process::exit(1)` at line 1502 is never reached and a `pipefail` CI gate reads green on a red run. The same early exit also skips the `--junit` write two lines below, so the XML artifact disappears. Before this PR `run` wrote nothing to stdout, so the path was unreachable; `print_result` (main.rs:157-166) already guards it for `storage get` (#213).

**Fix:** Set the exit status before printing, as `print_result` does, and move the JUnit write above the JSON print so a closed stdout does not also cost the XML.

```suggestion
    if json {
        // Status first, so a closed stdout pipe cannot turn the failure into a 0 (#213).
        if !report.all_passed() {
            output::set_broken_pipe_exit(1);
        }
        output::format_json(&scenario::report_to_json(&report))?;
    }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff adds the stdout write that makes the closed-pipe exit reachable for `run`

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.9</sub>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -1757,6 +1766,15 @@ fn replay_schema() -> Arc<JsonObject> {
     )
 }
 
+/// Failure screenshot directory used when the caller passes none.
+///
+/// The MCP server inherits its working directory from the client that spawned
+/// it, so the CLI default (`./tauri-pilot-failures`) would land anywhere. The
+/// temp directory is writable and predictable instead.
+fn default_screenshots_dir() -> PathBuf {
+    std::env::temp_dir().join(scenario::DEFAULT_SCREENSHOT_DIR)
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🔒 Security** — MCP default screenshot directory is a fixed, shared path in the system temp dir

`default_screenshots_dir()` resolves to the constant `/tmp/tauri-pilot-failures` on Linux — no uid, no pid, no randomness — and `save_screenshot_result` reaches it via `create_dir_all` + `fs::write` with default permissions (scenario.rs:604-607), so the directory lands 0755 and each PNG 0644. Two consequences on a shared host: the first user to hit a failing run owns the directory and everyone else's writes fail with EACCES, so the feature reports `screenshot_error: Permission denied` and silently stops working; and since the name is predictable in a world-writable `/tmp`, a local user can pre-create it as a symlink — `create_dir_all` succeeds on a symlink to a directory and `fs::write` follows it — and collect screenshots that routinely show login screens and tokens.

**Fix:** Make the default owner-private and non-guessable. The repo already has the pattern for this in `crates/tauri-plugin-pilot/src/server/unix.rs:60-95`: prefer `$XDG_RUNTIME_DIR` when `is_private_dir` accepts it, otherwise a per-user name under the temp dir created with `std::fs::DirBuilder::new().recursive(true).mode(0o700)`. The doc comment calls the temp dir "predictable", which is exactly the problem — reword it once the path is hardened.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff adds `default_screenshots_dir()` and makes it the default sink for `pilot.run` failure screenshots
- **References:** CWE-377, CWE-59, CWE-200

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.8</sub>

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
```diff
@@ -0,0 +1,150 @@
...
+        panic!(
+            "mock server did not finish: {err}\n--- stderr ---\n{}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+    }
+    output
+}
+
+#[test]
+fn run_json_reports_screenshot_from_requested_directory() {
```
mpiton, line 136: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — Nothing exercises the CLI default directory or the relative→absolute conversion

The only end-to-end case passes an absolute `--screenshots-dir`, and both MCP cases pass a tempdir too. So `screenshots_dir.unwrap_or_else(|| Path::new(DEFAULT_SCREENSHOT_DIR))` (main.rs:1479) and `std::path::absolute(&path).unwrap_or(path)` (scenario.rs:582) are never run on a relative path — which is the half of #215 about the default location and the absolute path being reported. Deleting the `absolute()` call keeps the whole suite green.

**Fix:** Add a second `#[test]` here calling `run_failing_scenario(tmpdir.path(), &["--json"])`, then assert the reported `screenshot` is absolute, starts with `tmpdir.path().join("tauri-pilot-failures")`, and exists on disk. `run_failing_scenario` already takes `extra_args` for exactly this.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the default branch and the relative→absolute conversion are both added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.85</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #226

> [!IMPORTANT]
> 💬 **Review posted** — 2 blocking issues (0 critical, 2 high) out of 14 findings.

**Actionable comments posted: 13**

|:---|---:|:---|
| 🟠 High | 2 | must fix before merge |
| 🟡 Medium | 5 | fix now or file a follow-up |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/src/mcp.rs` | 0 | 0 | 2 | 2 |
| `crates/tauri-pilot-cli/src/main.rs` | 0 | 2 | 0 | 0 |
| `crates/tauri-pilot-cli/tests/run_failure_screenshot.rs` | 0 | 0 | 1 | 1 |
| `crates/tauri-pilot-cli/src/scenario.rs` | 0 | 0 | 0 | 2 |
| `crates/tauri-pilot-cli/src/cli.rs` | 0 | 0 | 0 | 1 |
| `README.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟠 **`Path` is imported under `#[cfg(unix)]`, so the new default breaks the Windows build** — `crates/tauri-pilot-cli/src/main.rs:1480`
2. 🟠 **`run --json` exits 0 on a failed scenario when stdout closes early** — `crates/tauri-pilot-cli/src/main.rs:1494`
3. 🟡 **MCP default screenshot directory is a fixed, shared path in the system temp dir** — `crates/tauri-pilot-cli/src/mcp.rs:1775`
4. 🟡 **Nothing exercises the CLI default directory or the relative→absolute conversion** — `crates/tauri-pilot-cli/tests/run_failure_screenshot.rs:110`
5. 🟡 **The mock now succeeds on `screenshot`, leaving the RPC-failure branch of `screenshot_error` untested** — `crates/tauri-pilot-cli/src/mcp.rs:3196`

<details>

- 🔵 **README and the ai-agents guide still describe the old `pilot.run` report** — `README.md:211` — `README.md:211-215` and `docs/src/content/docs/guides/ai-agents.md:168-171` carry the same `pilot.run` blurb cli.md has, listing the report as `ok`, counts, `summary`, `steps`. Neither mentions the new per-step `screenshot` / `screenshot_error` fields or `screenshots_dir`, so the three copies of the paragraph now disagree — and these two are the agent-facing ones.
  - *Fix:* One sentence in both: a failed step also carries `screenshot` (absolute path) or `screenshot_error`, and `screenshots_dir` overrides the default. The README `run` row (line 174) can gain `--screenshots-dir` next to `--junit`.
  - *Link to this PR:* the PR changed the pilot.run report shape and updated only the cli.md copy of this paragraph

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `bdccd64aaa` |
| Base | `main` |
| Files changed | 9 (+431/-24) |
| Languages | rust |
| Reviewers | curated |

</details>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -3111,10 +3187,12 @@ target = "#btn"
...
-                        serde_json::Value::Number(request.id.into()),
-                        -32_000,
-                        "screenshot failed",
+                    // 1x1 transparent PNG, enough for the save path to run.
+                    "screenshot" => Response::success(
+                        request.id,
+                        json!(
+                            "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="
+                        ),
                     ),
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The mock now succeeds on `screenshot`, leaving the RPC-failure branch of `screenshot_error` untested

`spawn_failing_click_server` used to answer `screenshot` with a JSON-RPC error; it now always returns a PNG, and no other fixture in the crate fails that call. `screenshot_error` is user-visible and has two distinct shapes — a write failure wrapped as `failed to save screenshot to <path>`, and a bare RPC error from `client.call("screenshot", …)` (scenario.rs:584-588). Only the write failure is asserted, in `run_tool_reports_why_screenshot_was_not_saved`.

**Fix:** Keep a fixture that errors on `screenshot` — a second mock, or make this one fail the call when the scenario name asks for it — and assert `screenshot_error` carries the RPC message with `screenshot` absent.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff flips the mock from error to success and at the same time makes that error text user-visible

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.8</sub>

CHANGELOG.md
```diff
@@ -34,6 +34,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Failure screenshots from `run` and MCP `pilot.run` are now reported and
+  no longer land wherever the process happened to start. A failed step
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — CHANGELOG claims `run` screenshots no longer follow the working directory

The entry says screenshots from `run` *and* `pilot.run` "no longer land wherever the process happened to start", but only the MCP default moved. `main.rs:1479` still falls back to the relative `DEFAULT_SCREENSHOT_DIR`, so `tauri-pilot run` from a repo root keeps creating `./tauri-pilot-failures/` there unless `--screenshots-dir` is passed — the exact symptom #215 reports.

**Fix:** Scope the claim to `pilot.run` and let the next sentence carry the CLI story (`--screenshots-dir` picks the directory; the default stays `./tauri-pilot-failures`).

```suggestion
- Failure screenshots from `run` and MCP `pilot.run` are now reported, and
  `pilot.run` no longer leaves them wherever it was spawned. A failed step
```

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.9</sub>

docs/src/content/docs/reference/cli.md
```diff
@@ -112,7 +112,12 @@ The MCP server exposes tools for the CLI's app-facing commands, including
 
 `pilot.run` executes a declarative TOML scenario. Pass `path` to a `.toml` file
 or inline `content` (not both), optionally set `fail_fast` to override the file,
-and read the JSON report (`ok`, counts, `summary`, `steps`). A finished run
+and read the JSON report (`ok`, counts, `summary`, `steps`). A failed step also
```
mpiton, line 115: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — `run --screenshots-dir` and the now-working `run --json` are missing from the CLI reference

This PR documents `screenshots_dir` for MCP in this very file, but the CLI flag and the fact that `run --json` now prints the report appear nowhere on the docs site: `grep -rn 'screenshots-dir' docs/` returns nothing, and cli.md has no `### run` section at all (the `###` headings stop at `screenshot`). A user following the published reference cannot learn where failure screenshots go or how to move them.

**Fix:** Add a short `### run` subsection listing `--junit <FILE>`, `--no-fail-fast`, `--screenshots-dir <DIR>` (default `./tauri-pilot-failures`, relative to the cwd) and `--json` (report on stdout, human summary on stderr), mirroring the SKILL.md table.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the PR adds the flag and makes `--json` functional, and updates this file for the MCP side only

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.85</sub>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -471,6 +471,8 @@ impl PilotMcpServer {
         let path = optional_string(&args, "path")?;
         let content = optional_string(&args, "content")?;
         let fail_fast = optional_bool(&args, "fail_fast")?;
+        let screenshots_dir = optional_string(&args, "screenshots_dir")?
+            .map_or_else(default_screenshots_dir, PathBuf::from);
```
mpiton, line 478: > [!NOTE]
> **🔵 Low · 🐛 Correctness** — An empty `screenshots_dir` bypasses the temp-dir default and writes to the server's cwd

`optional_string` returns `Some("")` for `"screenshots_dir": ""` (mcp.rs:1258-1264), so `map_or_else` takes the `PathBuf::from` branch and `default_screenshots_dir` is skipped. `Path::new("").join(filename)` is a bare filename, which `std::path::absolute` then resolves against the server's working directory — the placement this PR removes. `--screenshots-dir ''` on the CLI behaves the same way.

**Fix:** Treat an empty or whitespace value as absent, here and in `run_scenario_command`.

```suggestion
        let screenshots_dir = optional_string(&args, "screenshots_dir")?
            .filter(|dir| !dir.trim().is_empty())
            .map_or_else(default_screenshots_dir, PathBuf::from);
```

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.85</sub>

crates/tauri-pilot-cli/src/scenario.rs
```diff
@@ -546,14 +579,16 @@ async fn take_failure_screenshot(
         .collect();
     let filename = format!("{safe_name}-{ts}.png");
     let path = dir.join(filename.as_str());
+    let path = std::path::absolute(&path).unwrap_or(path);
```
mpiton: > [!NOTE]
> **🔵 Low · 🐛 Correctness** — `absolute()` failure silently degrades the path the report promises is absolute

`std::path::absolute(&path).unwrap_or(path)` keeps the relative join when the call fails (deleted or unreadable cwd), with no log line. The function's own doc comment, the CHANGELOG, SKILL.md and cli.md all state the reported path is absolute, so `screenshot` in the JSON report and the JUnit `<system-out>` can hand a consumer a path it cannot open — re-entering #215 through a silent fallback.

**Fix:** Propagate instead of swallowing, so the step reports `screenshot_error` rather than an unusable path: `let path = std::path::absolute(&path).with_context(|| format!("failed to resolve {}", path.display()))?;`

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the fallback is added by this diff alongside the absolute-path change

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.8</sub>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -3066,6 +3138,10 @@ target = "#btn"
         if let Some(fail_fast) = fail_fast {
             args.insert("fail_fast".to_owned(), json!(fail_fast));
         }
+        args.insert(
+            "screenshots_dir".to_owned(),
+            json!(screenshots_dir.display().to_string()),
+        );
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — The `pilot.run` temp-dir default has no test caller

`call_run_two_clicks` now always inserts `screenshots_dir`, so `default_screenshots_dir()` (mcp.rs:1774) is never reached from a test; `run_schema_properties_include_run_options_and_window` only checks the property is advertised. Replacing the default with `PathBuf::from(DEFAULT_SCREENSHOT_DIR)` — the client-cwd behaviour this PR fixes — would pass the suite.

**Fix:** Take `screenshots_dir: Option<&Path>` in the helper and insert the argument only when `Some`, then add a case that omits it and asserts the reported path starts with `std::env::temp_dir().join(scenario::DEFAULT_SCREENSHOT_DIR)`; clean up the PNG afterwards.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the temp-dir default is introduced by this diff and the test helper it modifies now hides it

</details>

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.9</sub>

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// How long to wait for the mock server once the binary has exited.
+const SERVER_DONE_TIMEOUT: Duration = Duration::from_secs(10);
+
+fn unique_socket_path() -> PathBuf {
+    PathBuf::from(format!(
+        "/tmp/tauri-pilot-it-run-shot-{}.sock",
+        std::process::id()
+    ))
+}
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — The new integration test re-implements `tests/common` helpers and drops the socket counter

`SERVER_DONE_TIMEOUT` and `unique_socket_path` already exist in `tests/common/mod.rs`, and the looping mock below is a near copy of `storage_get_scenario.rs:52-84`. The local `unique_socket_path` drops the atomic counter the shared one has, so the moment a second `#[test]` calls `run_failing_scenario` — which is parameterised by `extra_args` for exactly that, see the missing default-directory case above — both bind `/tmp/tauri-pilot-it-run-shot-<pid>.sock` and one fails on `bind`.

**Fix:** `mod common;` then `common::SERVER_DONE_TIMEOUT` and `common::unique_socket_path("run-shot")`. The multi-request mock now has three copies under `tests/`; lifting it into `tests/common/mod.rs` beside `spawn_mock_server`, as a scripted variant taking a `method -> result` closure, would let this file and `storage_get_scenario.rs` share it.

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.85</sub>

crates/tauri-pilot-cli/src/cli.rs
```diff
@@ -242,6 +242,9 @@ pub(crate) enum Command {
         /// Override `fail_fast` setting from the scenario file.
         #[arg(long)]
         no_fail_fast: bool,
+        /// Directory for failure screenshots (default: ./tauri-pilot-failures).
+        #[arg(long, value_name = "DIR")]
+        screenshots_dir: Option<PathBuf>,
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — The screenshots-dir default is spelled three times

Once in the clap help text here, once in `main.rs:1479` as `unwrap_or_else(|| Path::new(scenario::DEFAULT_SCREENSHOT_DIR))`, once in `mcp.rs::default_screenshots_dir`. The help string goes stale the first time the constant changes, and the `Option` also gives `run_scenario_command` two adjacent `Option<&Path>` parameters (`junit`, `screenshots_dir`) that the compiler cannot tell apart if the call site ever swaps them.

**Fix:** Let clap own the default — `#[arg(long, value_name = "DIR", default_value = crate::scenario::DEFAULT_SCREENSHOT_DIR)] screenshots_dir: PathBuf` — so `--help` renders the real constant, `main.rs` drops the fallback (and with it the Windows break above), and the parameter becomes a plain `&Path`.

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.8</sub>

crates/tauri-pilot-cli/src/scenario.rs
```diff
@@ -95,10 +95,15 @@ pub(crate) enum StepOutcome {
...
 
+/// Where the failure screenshot landed, or why it could not be saved.
+pub(crate) type ScreenshotOutcome = Result<PathBuf, String>;
+
 #[derive(Debug)]
 pub(crate) struct StepResult {
     pub(crate) name: String,
     pub(crate) outcome: StepOutcome,
+    /// `None` unless the step failed and a screenshot was attempted.
+    pub(crate) screenshot: Option<ScreenshotOutcome>,
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — `StepResult.screenshot` can express two states the runner never produces

A screenshot is only attempted on the `Err` arm of `run_scenario_steps`, so `Some(..)` on a passed or skipped step is unreachable — yet every construction site (`:222`, `:259`, the test helper at `:796`) must spell `screenshot: None`, and both serializers re-open the `Option`.

**Fix:** Put the field in the variant that owns it: `StepOutcome::Failed { duration, message, screenshot: ScreenshotOutcome }`. That removes the `Option`, the three `screenshot: None` literals, the `step_result_to_json`/`step_outcome_to_json` split (the `Failed` arm builds the object in one `json!`), and the `as_ref()?` in `screenshot_note`.

<sub>🤖 <b>mpiton-review</b> · curated · confidence 0.75</sub>

CHANGELOG.md
```diff
@@ -34,6 +34,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  carries the absolute file in its result (`screenshot`), or the reason it
+  could not be written (`screenshot_error`), in the MCP report, in
+  `run --json` and as `<system-out>` in the JUnit XML. `run
+  --screenshots-dir <DIR>` and the `screenshots_dir` argument on `pilot.run`
+  pick the directory; `pilot.run` defaults to `tauri-pilot-failures` under
+  the system temp directory rather than the client's working directory.
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> The CLI default remains relative to the process working directory: `./tauri-pilot-failures`. Only MCP uses the system temporary directory.

Right, the entry over-claimed. Fixed in `4768452`: the "no longer drops them wherever the process happened to start" clause is scoped to `pilot.run` now, and the next sentence says `run` keeps `./tauri-pilot-failures` relative to the working directory while `pilot.run` defaults to a per-user directory under `$XDG_RUNTIME_DIR` or the temp dir.

CONTRIBUTING.md
```diff
@@ -52,7 +52,8 @@ tauri-pilot offers two complementary automation modes:
 
 **`run` (TOML scenario)** — define steps declaratively with action types, assertions, and
 timeouts. Exits 0 on success, 1 on any failure. Supports JUnit XML output (`--junit`).
-Automatically captures failure screenshots to `./tauri-pilot-failures/`.
+Automatically captures failure screenshots to `./tauri-pilot-failures/`, or to
+`--screenshots-dir <DIR>`; the path lands in the failed step's result.
```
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> When saving fails, the failed step contains `screenshot_error` instead of a screenshot path.

Fixed in `4768452`. CONTRIBUTING.md:55-57 now reads "carries the absolute path as `screenshot`, or the reason the shot could not be written as `screenshot_error`".

crates/tauri-pilot-cli/src/mcp.rs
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Use an OS-created, unique owner-only temporary directory (or otherwise secure the directory before saving) so screenshots cannot be read or redirected by another local user.

Valid, fixed in `4768452`. `default_screenshots_dir` creates `tauri-pilot-failures-<uid>` with `DirBuilder::mode(0o700)`, preferring `$XDG_RUNTIME_DIR` when that is itself a private directory we own, and re-checks the result with `symlink_metadata` + `uid()` + `mode().trailing_zeros() >= 6`, so a pre-created symlink is rejected instead of followed. A squatted name falls back to a pid-suffixed one, and a path that cannot be secured is still returned so the write failure surfaces as `screenshot_error` rather than landing somewhere world-readable.

Not a fresh `mkdtemp` per call on purpose: the directory has to stay stable across runs so a CI job can collect the PNGs afterwards.

docs/src/content/docs/reference/cli.md
```diff
@@ -112,7 +112,12 @@ The MCP server exposes tools for the CLI's app-facing commands, including
...
+carries the absolute failure screenshot as `screenshot`, or the reason it could
+not be written as `screenshot_error`. Screenshots go to `tauri-pilot-failures`
+under the system temp directory unless `screenshots_dir` says otherwise — the
```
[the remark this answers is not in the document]

> The PR adds the CLI flag `--screenshots-dir` (default `./tauri-pilot-failures`), but the CLI reference only documents the MCP `screenshots_dir` variant.

Fixed in `4768452`: cli.md has a `### run` section listing `--junit <FILE>`, `--no-fail-fast`, `--screenshots-dir <DIR>` (default `./tauri-pilot-failures`, resolved against the cwd) and `--json`, plus a note that a failed step reports the absolute path as `screenshot` or the reason as `screenshot_error`.

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
```diff
@@ -0,0 +1,150 @@
...
+use assert_cmd::Command;
+
+/// How long to wait for the mock server once the binary has exited.
+const SERVER_DONE_TIMEOUT: Duration = Duration::from_secs(10);
+
+fn unique_socket_path() -> PathBuf {
+    PathBuf::from(format!(
+        "/tmp/tauri-pilot-it-run-shot-{}.sock",
+        std::process::id()
+    ))
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Tests of one binary run in parallel threads under the same PID, so the moment a second test is added here, both will remove/bind the same path and their connections will cross-wire or one bind will fail

Correct, and that second test exists now. `4768452` switches the file to `mod common;` and `common::unique_socket_path("run-shot")`, so the name carries the atomic `SOCK_COUNTER`. Three tests share the binary and none collide.

CHANGELOG.md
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> The CLI default remains relative to the process working directory, so this wording is inaccurate.

Fixed in `4768452`. The claim is scoped to `pilot.run`, and `run` is spelled out as keeping `./tauri-pilot-failures` relative to the working directory.

CONTRIBUTING.md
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> Document that a failed step reports either the absolute screenshot path or `screenshot_error` when saving fails.

Done in `4768452`, CONTRIBUTING.md:55-57.

crates/tauri-pilot-cli/src/main.rs
mpiton: @mpiton

> `use std::path::Path;` at main.rs:20-21 is gated by `#[cfg(unix)]` [...] `run_scenario_command` is compiled on every target, so this line does not resolve on Windows.

Fixed in `4768452`, by the route the cli.rs finding suggested rather than a qualified path: clap owns the default now (`default_value = crate::scenario::DEFAULT_SCREENSHOT_DIR`), the parameter is `screenshots_dir: &std::path::Path`, and the `unwrap_or_else` line is gone. `cargo check --target x86_64-pc-windows-gnu -p tauri-pilot-cli` exits 0.

crates/tauri-pilot-cli/src/main.rs
mpiton: @mpiton

> `tauri-pilot run ci.toml --json | head -1` on a failing scenario exits 0 [...] The same early exit also skips the `--junit` write two lines below, so the XML artifact disappears.

Fixed in `4768452`. The JUnit write moved above the JSON print, and `output::set_broken_pipe_exit(1)` runs before it when the report failed, same order as `print_result`. `run_json_into_closed_pipe_keeps_exit_1_and_writes_junit` in tests/run_failure_screenshot.rs wires stdout to a pipe whose reader is already dropped; with the `set_broken_pipe_exit` call removed it fails.

crates/tauri-pilot-cli/src/mcp.rs
mpiton: @mpiton

> since the name is predictable in a world-writable `/tmp`, a local user can pre-create it as a symlink [...] and collect screenshots that routinely show login screens and tokens

Fixed in `4768452`. `default_screenshots_dir` builds `tauri-pilot-failures-<uid>` with `DirBuilder::mode(0o700)` under `$XDG_RUNTIME_DIR` when that is private, else the temp dir, and validates the result with `symlink_metadata` so a symlink is rejected rather than followed, same shape as `server/unix.rs:58-95`. The uid in the name closes the EACCES half, a squatted name falls back to a pid suffix, and a path that cannot be secured is still returned so the caller gets `screenshot_error`. Windows keeps the bare name, `temp_dir()` is already per-user there. `run_tool_without_screenshots_dir_uses_the_private_default` asserts `is_private_dir` on what the tool reports.

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
mpiton: @mpiton

> Deleting the `absolute()` call keeps the whole suite green.

Fixed in `4768452`: `run_json_reports_absolute_path_under_the_default_directory` runs with `--json` only and asserts the reported path is absolute, starts with `<cwd>/tauri-pilot-failures` and exists on disk. The tempdir is canonicalized before it becomes the cwd, otherwise macOS compares `/var/...` against the `/private/var/...` that `absolute()` returns. The `absolute()` call is no longer a silent fallback either, see the scenario.rs thread.

crates/tauri-pilot-cli/src/mcp.rs
mpiton: @mpiton

> Keep a fixture that errors on `screenshot` [...] and assert `screenshot_error` carries the RPC message with `screenshot` absent.

Fixed in `4768452`. `spawn_failing_click_server_with(socket, ScreenshotMock)` takes a `Png` / `RpcError` selector and the old helper delegates with `Png`. `run_tool_reports_the_rpc_error_that_blocked_the_screenshot` uses `RpcError` (`-32000 window is gone`) and asserts exactly that.

CHANGELOG.md
mpiton: @mpiton

> The entry says screenshots from `run` *and* `pilot.run` "no longer land wherever the process happened to start", but only the MCP default moved.

Fixed in `4768452`, close to the suggestion. The claim is scoped to `pilot.run`, and the directory sentence names both defaults explicitly plus the reason the old `/tmp` path had to go.

docs/src/content/docs/reference/cli.md
mpiton: @mpiton

> Add a short `### run` subsection listing `--junit <FILE>`, `--no-fail-fast`, `--screenshots-dir <DIR>` (default `./tauri-pilot-failures`, relative to the cwd) and `--json` (report on stdout, human summary on stderr)

Added in `4768452`, above the JSON-RPC section, with those four flags, the default, an example block and a note on `screenshot` / `screenshot_error`. `9757690` fixes the `--json` row to say the summary goes to stderr either way, which is what `print_report` (scenario.rs:638-645) does.

crates/tauri-pilot-cli/src/mcp.rs
mpiton: @mpiton

> `optional_string` returns `Some("")` for `"screenshots_dir": ""` [...] `--screenshots-dir ''` on the CLI behaves the same way.

Fixed in `4768452`, suggestion taken verbatim for mcp.rs. The CLI half goes through clap: `value_parser = screenshots_dir_value` maps an empty or whitespace value back to `DEFAULT_SCREENSHOT_DIR`, so `--screenshots-dir ''` lands in the default directory instead of the cwd.

crates/tauri-pilot-cli/src/scenario.rs
mpiton: @mpiton

> `std::path::absolute(&path).unwrap_or(path)` keeps the relative join when the call fails (deleted or unreadable cwd), with no log line.

Fixed in `4768452` with the suggested `with_context(...)?`, so an unresolvable cwd shows up as `screenshot_error` instead of a path the consumer cannot open.

crates/tauri-pilot-cli/src/mcp.rs
mpiton: @mpiton

> Take `screenshots_dir: Option<&Path>` in the helper and insert the argument only when `Some`, then add a case that omits it

Done in `4768452`. `call_run_two_clicks` takes `Option<&Path>`, the four existing call sites pass `Some(...)`, and `run_tool_without_screenshots_dir_uses_the_private_default` omits it. It asserts against the hardened default rather than `temp_dir().join(DEFAULT_SCREENSHOT_DIR)`, since that fixed path is what the security thread moved away from: `tauri-pilot-failures-<uid>` under `$XDG_RUNTIME_DIR` or the temp dir, `is_private_dir` on it, PNG removed afterwards.

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
mpiton: @mpiton

> The local `unique_socket_path` drops the atomic counter the shared one has, so the moment a second `#[test]` calls `run_failing_scenario` [...] both bind `/tmp/tauri-pilot-it-run-shot-<pid>.sock` and one fails on `bind`.

Fixed in `4768452`: `mod common;` with both helpers imported. `tests/common/mod.rs` also gained a module-level `#![allow(dead_code)]`, because each test binary compiles the module separately and `run_failure_screenshot` does not use `spawn_mock_server`; `#[expect]` would fire an unfulfilled expectation in `broken_pipe_exit.rs`, which uses every item.

Skipped the scripted-mock lift. `storage_get_scenario.rs` collects methods and params over channels while this file only maps method to response, so the shared helper needs both a script and a recorder to replace ~25 lines of copy. Worth doing when a fourth caller shows up.

crates/tauri-pilot-cli/src/cli.rs
mpiton: @mpiton

> Let clap own the default [...] so `--help` renders the real constant, `main.rs` drops the fallback (and with it the Windows break above), and the parameter becomes a plain `&Path`.

Done in `4768452`, with one addition: `value_parser = screenshots_dir_value` next to `default_value`, so `--screenshots-dir ''` maps back to the constant instead of reaching `Path::new("").join(name)`. The parser is infallible and carries `#[expect(clippy::unnecessary_wraps)]` since clap's `value_parser` only accepts `Fn(&str) -> Result<T, E>`.

crates/tauri-pilot-cli/src/scenario.rs
mpiton: @mpiton

> Put the field in the variant that owns it: `StepOutcome::Failed { duration, message, screenshot: ScreenshotOutcome }`

Done in `4768452`. The `Option`, the three `screenshot: None` literals and the `step_result_to_json` / `step_outcome_to_json` split are gone; the `Failed` arm builds the object and attaches `screenshot` / `screenshot_error` in one place, and `screenshot_note` matches the variant directly. Test fixtures got `failed(ms, message)` / `failed_with(ms, message, screenshot)`, which also removed the two `report.results[n].screenshot = ...` mutations.

crates/tauri-pilot-cli/src/cli.rs
```diff
@@ -242,9 +242,38 @@ pub(crate) enum Command {
         /// Override `fail_fast` setting from the scenario file.
         #[arg(long)]
         no_fail_fast: bool,
+        /// Directory for failure screenshots.
+        #[arg(
+            long,
+            value_name = "DIR",
+            default_value = crate::scenario::DEFAULT_SCREENSHOT_DIR,
+            value_parser = screenshots_dir_value
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When `--screenshots-dir` names a Unix directory containing non-UTF-8 bytes, this custom `Fn(&str)` parser rejects the otherwise valid path.

Valid, fixed in `a70571a`. The parser is `OsStringValueParser::new().map(screenshots_dir_value)` now and takes an `OsString`, so the bytes go straight into the `PathBuf`. Blank still means "use the default", decided on `as_encoded_bytes().iter().all(u8::is_ascii_whitespace)` — ASCII bytes never appear inside a multi-byte sequence, so that check cannot chop a non-UTF-8 path. `non_utf8_screenshots_dir_is_kept_verbatim` in cli.rs parses `/tmp/shots-\xff` and asserts it round-trips.

CHANGELOG.md
```diff
@@ -34,6 +34,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  (`screenshot`), or the reason it could not be written
+  (`screenshot_error`), in the MCP report, in `run --json` and as
+  `<system-out>` in the JUnit XML. `run --screenshots-dir <DIR>` and the
+  `screenshots_dir` argument on `pilot.run` pick the directory. `run` keeps
+  its `./tauri-pilot-failures` default, relative to the working directory;
+  `pilot.run` now defaults to an owner-only per-user
+  `tauri-pilot-failures-<uid>` directory under `$XDG_RUNTIME_DIR` or the
+  system temp directory, because the server's working directory belongs to
+  whichever client spawned it, and the old fixed `/tmp` path was writable
+  and readable by every user on the host. `run --json` prints the report as
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> A directory first created there with default permissions (0755 under a typical umask) is world-readable but only owner-writable; the real exposure was that any local user could pre-create the predictable name

Right, reworded in `a70571a`: "was world-readable, and since the name was predictable in a world-writable `/tmp`, any local user could pre-create it as a directory or a symlink and collect or redirect the screenshots".

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
```diff
@@ -0,0 +1,245 @@
...
+        "screenshot {} is not under the default directory {}",
+        saved.display(),
+        expected.display()
+    );
+    assert!(saved.is_file(), "screenshot {} is missing", saved.display());
+}
+
+/// Returns a stdout whose reader is already gone, so the first write fails
+/// with `EPIPE`.
+fn closed_pipe() -> Stdio {
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> `closed_pipe()` duplicates the identical helper in `crates/tauri-pilot-cli/tests/broken_pipe_exit.rs`.

Fixed in `a70571a`. `closed_pipe` lives in `tests/common/mod.rs` and both binaries call it.

crates/tauri-pilot-cli/tests/run_failure_screenshot.rs
```diff
@@ -0,0 +1,245 @@
...
+    ];
+    args.extend_from_slice(extra_args);
+    let output = Command::new(env!("CARGO_BIN_EXE_tauri-pilot"))
+        .current_dir(cwd)
+        .args(args)
+        .stdout(stdout)
+        .stderr(Stdio::piped())
+        .spawn()
+        .expect("spawn tauri-pilot")
+        .wait_with_output()
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> If a regression makes the child hang on the closed stdout — the exact failure mode this test exists to catch — the suite hangs until the CI runner kills it instead of failing here.

Valid, fixed in `a70571a`. The poll loop from `broken_pipe_exit.rs::run` moved into `common::wait_bounded`, which kills the child and panics at `SERVER_DONE_TIMEOUT`; both files call it, so neither keeps its own copy. One behaviour change on the timeout path: the socket file is no longer unlinked, since the panic now happens before `run` gets to it. The name carries the pid and the counter, so a leftover cannot collide with a later run.

PR #227, opened 2026-09-20
fix(cli): bare snapshot ids parse as refs for all commands (#216)

### LINKED ISSUE #216: Bare snapshot ref e1 only works for scroll; click, value and others treat it as a CSS selector
Since #157/#180, `scroll --ref e1` accepts a bare snapshot ref, but every other target-taking command reads `e1` as a CSS selector and fails. The MCP tool schemas disagree with each other on this too.

Found during end-to-end QA of 0.8.0 (Linux, plugin and CLI 0.7.3, commit 61ba190).

## Reproduce

```
$ tauri-pilot snapshot -i | head -1
            - textbox [ref=e1] value="user@example.com"

$ tauri-pilot scroll down 10 --ref e1
✓ ok
$ tauri-pilot value @e1
user@example.com
$ tauri-pilot value e1
Error: RPC error (-32603): Eval error: JavaScript error: No element matches selector: e1
$ tauri-pilot click e1
Error: RPC error (-32603): Eval error: JavaScript error: No element matches selector: e1
```

Over MCP:

```
pilot.click {"target":"e1"}  -> isError: "No element matches selector: e1"
pilot.value {"target":"e1"}  -> isError: "No element matches selector: e1"
pilot.value {"ref":"e1"}     -> -32602 "'target' is required"
```

`pilot.scroll` documents `ref` as "Snapshot refs accept e12 or @e12". `pilot.click` and `pilot.value` say "Element ref, CSS selector, or x,y coordinates" and don't mention the `@`. An agent that copies `[ref=e1]` from a snapshot reasonably sends `e1` and gets a selector error.

## Cause

The normalization lives only in the scroll path (`normalize_scroll_target` / `is_snapshot_ref_id` in `crates/tauri-pilot-cli/src/main.rs`). `parse_target` (`cli.rs:319`) treats only `@`-prefixed strings as refs.

## Expected

Move the `e<digits>` normalization into `parse_target`, or the shared MCP target parsing, so every command accepts it. A bare `e12` isn't a useful CSS selector: custom element names need a hyphen, so no real element is called `e12`. Also accept `ref` as an alias of `target` everywhere, or nowhere, and word the schema descriptions the same across tools.

mpiton:
Bare snapshot ids (`e12`) were only accepted as refs by `scroll`; `click`, `value`, `fill`, and others treated them as CSS selectors and failed with 'No element matches selector: e12'. Snapshot prints `[ref=e12]`, so agents copy that id verbatim.

• Move normalization from scroll's `normalize_scroll_target` into `parse_target`, the shared entry point for CLI, MCP, and TOML scenario steps
• Remove `optional_scroll_target` wrapper; use `optional_string` on 'target' only
• MCP breaking: tools take element through 'target', not 'ref'; `scroll --ref` flag unchanged
• Unify every target/source property description with shared constant `TARGET_SHAPES`
• Add test cases for bare ref recognition and mutual exclusion

Fixes #216.

---
## Summary by cubic
Fixes bare snapshot ids (`e12`) so they parse as refs for every target-taking command; previously only `scroll` accepted them, while `click`, `value`, and `fill` read them as CSS selectors and failed with "No element matches selector: e12". A bare `e<digits>` is now always a ref, so a page with an unknown `<e12>` tag must use `:is(e12)` or `wait --selector e12`.

- MCP tools no longer accept a `ref` property; pass the element through `target` (`{"ref": "e12"}` becomes `{"target": "e12"}`). The MCP `scroll` tool now rejects `ref` instead of silently scrolling the page, and TOML steps reject `ref` at parse time (`ref = "e1"` becomes `target = "e1"`). The CLI `scroll --ref` flag is unchanged.
- Unified every MCP `target`/`source` property description; `wait` and `html` no longer advertise coordinate support.
- Added tests for bare ref recognition, near-miss selectors, and the removed `ref` alias.

<sup>Written for commit b61c8de571d0d5018f7d8931bc13513db9e8992c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Bare snapshot references such as `e3` are now accepted alongside `@e3` across supported targeting commands and interfaces.
  - Scroll targeting supports snapshot references, CSS selectors, and coordinates through a unified target input.
  - MCP targeting descriptions now document supported reference, selector, and coordinate formats.

- **Documentation**
  - Updated CLI, skill, and changelog documentation to reflect the expanded target syntax.
  - MCP scroll targeting now uses the `target` input consistently.


mpiton (commented):
## 🧨 mpiton-review — PR #227

> [!IMPORTANT]
> 💬 **Review posted** — 1 blocking issue (0 critical, 1 high) out of 8 findings.

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-pilot-cli/src/mcp.rs` | 0 | 1 | 1 | 2 |
| `crates/tauri-pilot-cli/src/cli.rs` | 0 | 0 | 0 | 1 |
| `crates/tauri-pilot-cli/src/scenario.rs` | 0 | 0 | 1 | 0 |
| `docs/src/content/docs/guides/architecture.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟠 **MCP scroll silently drops the legacy `ref` property instead of erroring** — `crates/tauri-pilot-cli/src/mcp.rs:188`
2. 🟡 **Shared target wording promises x,y coordinates to `wait` and `html`, which ignore them** — `crates/tauri-pilot-cli/src/mcp.rs:1600`
3. 🔵 **Target-description test asserts against the same constant that builds the descriptions** — `crates/tauri-pilot-cli/src/mcp.rs:2085`
4. 🔵 **Rationale for claiming `e12` says no real element can carry that name** — `crates/tauri-pilot-cli/src/cli.rs:369`

<details>
<summary>🧭 Findings outside the diff or over the inline cap (4)</summary>

- 🟡 **The `ref`/`target` alias deleted from MCP survives as two hand-rolled copies in scenario.rs** — `crates/tauri-pilot-cli/src/scenario.rs:547` — `scroll_step_target` is the deleted `optional_scroll_target` re-implemented for TOML, and the `wait` arm at scenario.rs:368-374 is a third copy of the same `target`/`ref` conflict check. `pilot.run` parses those steps (mcp.rs:488-492), so an MCP caller can still address an element through `ref` via a scenario; the new `no_tool_advertises_a_ref_alias` test cannot see it because step shapes never re
  - *Fix:* Apply the PR's own argument to the TOML surface: drop `step_ref` from `Step` (scenario.rs:74), delete `scroll_step_target`, delete the `wait` conflict block plus the `format!("@{r}")` re-encode at scenario.rs:373 — its comment's reason (keep `parse_target` single-input) no longer holds now that bare
  - *Link to this PR:* the PR deletes the MCP copy of this helper on the grounds that `ref` is now a redundant second spelling, and leaves the two TOML copies of the same helper behind
- 🟡 **`wait` reference still documents any non-`@` target as a CSS selector** — `docs/src/content/docs/reference/cli.md:1066` — The `wait` section enumerates the parse rules as `@e3` — snapshot ref, then `anything else — CSS selector`. After this PR `build_wait_params` routes a bare `e3` through `parse_target`, which returns `Target::Ref`, so `tauri-pilot wait e3` now does an `idMap` lookup and fails with `Unknown ref: e3` instead of the documented `querySelector` wait. This is the only place in the docs that still spells 
  - *Fix:* Update the two bullets at lines 1065-1066 to match the table at line 44: `@e3` or `e3` — snapshot ref (the `@` is optional), and anything else that is not `e<digits>` — CSS selector.
  - *Link to this PR:* the PR moves the `e<digits>` normalization into `parse_target`, the exact parser this section documents for `wait`.
- 🔵 **Two tool descriptions keep their own wording of the shapes TARGET_SHAPES was meant to unify** — `crates/tauri-pilot-cli/src/mcp.rs:837` — `click`'s description still reads "by ref, selector, or coordinates" and `scroll`'s (mcp.rs:1005) reads "(ref, CSS selector, or coordinates)". Neither mentions the bare `e12` form this PR exists to advertise, and `target_properties_share_one_wording` only walks `properties`, so the new test reads as proof of one wording while two divergent copies remain in the same file.
  - *Fix:* Drop the shape list from both tool descriptions — the `target` property immediately below already carries `TARGET_SHAPES` — so the constant is the single source. Same edit at mcp.rs:1005: `description: "Scroll the page or an element."`.
- 🔵 **Architecture guide still states CLI refs require the `@` prefix** — `docs/src/content/docs/guides/architecture.md:74` — The "Element Reference System" section says "In CLI commands, refs use the `@` prefix: `@e1`, `@e2`, etc.", which now contradicts SKILL.md:54 and reference/cli.md:44, both updated by this PR to say the `@` is optional. A reader landing on the guide will not learn that the ids `snapshot` prints are usable verbatim — the very confusion #216 reports.
  - *Fix:* Reword line 74 to "In CLI commands, refs may carry the `@` prefix (`@e1`) or be used bare (`e1`), exactly as `snapshot` prints them".
  - *Link to this PR:* this PR updated the other two target-format descriptions for the optional `@` and left this third one behind.

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `447e127356` |
| Base | `main` |
| Files changed | 6 (+139/-142) |
| Languages | rust |
| Reviewers | documentation, maintainability, product-logic, quality, rust, security, tests |

</details>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -185,7 +185,7 @@ impl PilotMcpServer {
             }
             "check" => self.target_call("check", &args, window).await,
             "scroll" => {
-                let target = optional_scroll_target(&args)?;
+                let target = optional_string(&args, "target")?;
```
mpiton, line 197: > [!WARNING]
> **🟠 High · 🐛 Correctness** — MCP scroll silently drops the legacy `ref` property instead of erroring

`PilotMcpServer::call_tool` forwards `request.arguments` straight into `call_tool_by_name` with no schema validation; `additionalProperties: false` in `scroll_schema()` (mcp.rs:1841) is only advertised in `tools/list`, never enforced at runtime (confirmed: no validation call exists anywhere in mcp.rs, and rmcp 3.4.0's `ServerHandler::call_tool` is fully overridden here, bypassing whatever validation the crate might otherwise do). Replacing `optional_scroll_target(&args)?` with `optional_string(&args, "target")?` means a client still sending the pre-PR shape `{"ref": "e12"}` (no `target` key) is not rejected: `target` silently becomes `None`, and `build_scroll_params` scrolls the whole page instead of the intended element, with no error surfaced anywhere. Before this diff, `optional_scroll_target` read both keys and used whichever was set, so the same call worked correctly.

**Fix:** Either reject a leftover `ref` key explicitly (e.g. `if args.contains_key("ref") { return Err(invalid_params("scroll no longer accepts 'ref'; use 'target'")); }`) so old callers fail loudly and learn to migrate, or keep `ref` as a deprecated-but-working alias into `target`. Right now the breaking change is silent instead of loud, which is worse than either option. Also worth a regression test calling `call_tool_by_name("scroll", {"ref": ...})` directly, since none of the tests added in this PR exercise that path.

```suggestion
                if args.contains_key("ref") {
                    return Err(invalid_params(
                        "scroll no longer takes 'ref'; pass the element through 'target' (a bare 'e12' is accepted)",
                    ));
                }
                let target = optional_string(&args, "target")?;
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this line replaces `optional_scroll_target`, which was the only code path that looked at the `ref` argument

</details>

<sub>🤖 <b>mpiton-review</b> · 4 reviewers agree: product-logic, quality, rust, tests · confidence 1.0</sub>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -1641,10 +1597,7 @@ fn navigate_schema() -> Arc<JsonObject> {
 fn wait_schema() -> Arc<JsonObject> {
     object_schema(
         props([
-            (
-                "target",
-                string_prop("Element ref, CSS selector, or x,y coordinates."),
-            ),
+            ("target", target_prop("Element to act on")),
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — Shared target wording promises x,y coordinates to `wait` and `html`, which ignore them

`wait_schema` and `optional_target_schema` (pilot.html) now build their `target` description from `TARGET_SHAPES`, which ends in `or x,y coordinates`. Neither tool honours that: `build_wait_params` sends coordinates through as `{"selector": "100,200"}` (main.rs:1087) and the bridge's `waitFor` calls `querySelector("100,200")`, which throws a `SyntaxError`; `html(params)` branches on `params.ref || params.selector` only (bridge.js:1323), so a coordinate target silently returns `document.documentElement.innerHTML` — the whole page. The CLI reference already states the opposite for `wait` ("coordinate targets (`x,y`) are not supported here"). The wording predates this PR, but `target_properties_share_one_wording` now makes the wrong claim mandatory for every tool.

**Fix:** Split the constant: keep `TARGET_SHAPES` for the tools that resolve coordinates and add a no-coords variant for `wait` and `html` ("snapshot ref (e12 or @e12) or CSS selector"), then let the new wording test accept either. Alternatively make `html()` fall through to `resolveTarget` when `x`/`y` are present.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the line is added by this diff and the new wording test makes the incorrect claim mandatory.

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.85</sub>

crates/tauri-pilot-cli/src/mcp.rs
```diff
@@ -2099,8 +2064,54 @@ mod tests {
...
+            for name in ["target", "source"] {
+                let Some(prop) = properties.get(name) else {
+                    continue;
+                };
+                let description = prop
+                    .get("description")
+                    .and_then(Value::as_str)
+                    .expect("property has a description");
+                assert!(
+                    description.contains(TARGET_SHAPES),
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Target-description test asserts against the same constant that builds the descriptions

`target_prop` formats every description from `TARGET_SHAPES` and the assertion checks `description.contains(TARGET_SHAPES)`, so the wording itself is never pinned: setting `TARGET_SHAPES` to `""` or to a wrong list of shapes keeps the test green. It only catches a future property that hand-rolls its description instead of calling `target_prop`, not a regression in what the shapes say.

**Fix:** Keep the loop, and add one literal assertion next to it so the text is pinned, e.g. assert that `target_schema()`'s `target` description equals `"Element to act on: snapshot ref (e12 or @e12), CSS selector, or x,y coordinates."`.

crates/tauri-pilot-cli/src/cli.rs
```diff
@@ -352,9 +356,24 @@ pub(crate) fn parse_target(s: &str) -> Target {
...
+        return Target::Ref(s.to_owned());
+    }
+
     Target::Selector(s.to_owned())
 }
 
+/// Returns `true` for a bare snapshot id: `e` followed by at least one digit.
+///
+/// Claiming this shape costs nothing as a selector: a custom element name must
+/// contain a hyphen, so no real element is named `e12`.
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — Rationale for claiming `e12` says no real element can carry that name

The comment justifies the new parse rule with "a custom element name must contain a hyphen, so no real element is named `e12`". The hyphen rule only constrains `customElements.define`; an unknown tag `<e12>` still lands in the DOM as an `HTMLUnknownElement` and `document.querySelector("e12")` matches it. The stated rationale therefore claims a guarantee the platform does not give, and it is repeated verbatim in CHANGELOG.md:37 where users read it.

**Fix:** State the actual limit in both places: no *registered* custom element can be named `e12`, so the shape is safe to claim in practice; a bare type selector for an unknown tag is no longer reachable and must be written another way (e.g. `:is(e12)`, or `--selector` on `wait`).

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the comment and the CHANGELOG paragraph are both added by this diff.

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.7</sub>

PR #228, opened 2026-09-20
fix(plugin): unlink socket on SIGINT/SIGTERM (#217)

### LINKED ISSUE #217: SIGINT/SIGTERM leave the pilot socket file behind
Stopping the app with SIGINT or SIGTERM leaves `tauri-pilot-{identifier}.sock` behind. #194 fixed the normal quit path (`RunEvent::Exit`), but Ctrl+C in `tauri dev` and `kill` still skip it.

Found during end-to-end QA of 0.8.0 (Linux, plugin 0.7.3, commit 61ba190).

## Reproduce

```
$ ./target/debug/pilot-test-app &
$ kill <pid>                     # or Ctrl+C in `tauri dev`
$ ls /run/user/1000 | grep test-app
tauri-pilot-com.pilot.test-app.sock
tauri-pilot-com.pilot.test-app.sock.lock
```

## Impact

Low. The CLI's auto-detection skips sockets with no listener, and the next bind replaces the stale file under the #195 lock. The file still piles up, though, and `ls` or other tooling may take it for a running app.

## Possible fix

Listen for SIGINT/SIGTERM in the plugin with `tokio::signal::unix`, and unlink the socket before the process dies. Caveat: that means installing a handler inside the host app, and it can change how the app reacts to Ctrl+C. Chain to the default behaviour or call `app.exit(0)` so the host still quits. If that's too invasive for a debug-only plugin, documenting the limitation is enough.

mpiton:
Quitting with Ctrl+C or `kill` now removes tauri-pilot socket files.

SIGINT/SIGTERM end process before Tauri emits RunEvent::Exit, so plugin guard never drops. Plugin now watches both signals on Unix (debug builds only), unlinks socket via the same SocketGuard logic as normal quit, then re-raises with default handler.

- Add spawn_signal_cleanup task on Tauri's runtime watching unix::signal
- Add re_raise to restore SIG_DFL and re-raise so app dies with 128+n status
- Skip re-raise in tests (mock apps per process, would kill test runner)
- Add Unix-only tokio signal feature
- Regression test sigterm_unlinks_socket_file retries raise until socket gone
- CHANGELOG update

---
## Summary by cubic
Fixes #217: on Unix debug builds (Android excepted), Ctrl+C, `kill`, or closing the terminal running `cargo tauri dev` previously left the `tauri-plugin-pilot` socket behind because the signal killed the process before Tauri emitted `RunEvent::Exit`. The plugin now watches SIGINT, SIGTERM, and SIGHUP, unlinks the socket, and re-raises so the app still exits with 128+n status.

- Watchers are armed during `setup`, so a signal before the task's first poll no longer leaks the socket.
- Cleanup uses the same guard logic as normal quit, restores the default handler, and re-raises.
- Apps with their own signal handlers still run them, but the re-raise can cut a long graceful shutdown short.
- A crash or `SIGKILL` still leaves the socket; the next bind treats it as stale.
- Windows is unchanged and still leaves `instances/{identifier}.json` behind on Ctrl+C (#229).
- Adds the Unix-only `tokio` signal feature and regression tests for all three signals; re-raise is skipped in tests so the test runner survives (#230).

<sup>Written for commit 516d8173630b7e70b874833d8c532c1ead4814d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/228?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Unix debug builds now remove the Pilot socket when terminated with SIGINT or SIGTERM.
  - Signal handling is restored after cleanup, allowing normal termination behavior.
  - Test environments clean up the socket without terminating the test process.
  - Crashes and SIGKILL may still leave stale socket files.


crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -35,6 +35,8 @@ enigo = { version = "0.6.1", optional = true }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.189"
+# Signal watching is Unix-only (#217); Windows has no socket file to unlink.
+tokio = { workspace = true, features = ["signal"] }
```
mpiton, line 41: > [!CAUTION]
> **🔴 Critical · 🐛 Correctness** — tokio's `signal` feature is added without `macros`, so the crate no longer compiles on its own

`spawn_signal_cleanup` uses `tokio::select!` (lib.rs:141), which tokio gates behind its `macros` feature, but the new `cfg(unix)` entry enables only `signal`. The workspace's own `cargo test` / `cargo clippy --all-targets` pass because the dev-dependency (`features = ["test-util", "macros"]`) and `tauri-pilot-cli` unify the feature into the graph; a build of this package alone does not. `cargo check -p tauri-plugin-pilot --lib` fails locally with `error[E0433]: cannot find select in tokio`, and this PR's `Check & Test (macos-latest)` job is red on exactly that: step `Check iOS plugin` (`cargo check -p tauri-plugin-pilot --target aarch64-apple-ios-sim`) ends with `error: could not compile tauri-plugin-pilot (lib)`. `cargo package -p tauri-plugin-pilot` (release.yml:90) and every downstream Tauri app that pulls the plugin as a normal dependency hit the same wall, in debug on Unix — the only configuration this plugin runs in.

**Fix:** Declare the feature the code uses.

```suggestion
tokio = { workspace = true, features = ["signal", "macros"] }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** both the `tokio::select!` call and the unix tokio edge are added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.99</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -119,6 +119,52 @@ fn release_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>) {
...
+///
+/// The host app then dies from the signal it was sent, with the usual 128+n
+/// status, instead of ignoring it because the plugin installed a handler.
+#[cfg(all(unix, not(target_os = "android"), debug_assertions))]
+fn re_raise(signum: i32) {
+    // `cargo test` builds every mock app in one process, each with its own
+    // watcher: re-raising there kills the whole test run.
+    if cfg!(test) {
+        return;
+    }
```
mpiton: > [!WARNING]
> **🟠 High · 🧪 Tests** — `cfg!(test)` skips the re-raise, leaving it untested and the test binary deaf to SIGINT/SIGTERM

All plugin tests are in-crate unit tests, so `cfg!(test)` is true for every one of them: `sigterm_unlinks_socket_file` exercises `guard.release()` and nothing else, and `libc::signal(SIG_DFL)` + `libc::raise` never run under test. The riskier half of the fix is unpinned — if the re-raise regresses, a dev app silently swallows Ctrl+C and only SIGKILL stops it, and no test notices. The early return also skips `SIG_DFL`, while tokio's handler stays installed for the life of the process (tokio never unregisters it), so once any test builds a mock app the test binary itself ignores both signals: Ctrl+C on `cargo test -p tauri-plugin-pilot` kills cargo and orphans the test process.

**Fix:** Restore the default disposition unconditionally and skip only the `raise` under `cfg!(test)`, so the next signal still reaches the binary. To pin the exit contract, cover the real path from a child process: an integration test under `crates/tauri-plugin-pilot/tests/` (where `cfg(test)` is off for the plugin) that re-execs its own binary, builds the mock app there, sends SIGTERM from the parent and asserts both that the socket is gone and that `ExitStatus::signal()` is `SIGTERM`. With that in place the `cfg!(test)` escape hatch can go.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the PR adds both the signal handler and the `cfg!(test)` skip that makes its termination half untestable

</details>

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -119,6 +119,52 @@ fn release_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>) {
...
+    // SAFETY: `signal` and `raise` are libc entry points with no preconditions.
+    unsafe {
+        libc::signal(signum, libc::SIG_DFL);
+        libc::raise(signum);
```
> **🟡 Medium · 🎯 Product** — The plugin now decides how the host process dies, and no doc says so

tokio registers through `signal_hook_registry`, which chains to a handler the host app installed earlier, so the app's own SIGTERM path does start — but `re_raise` then sets `SIG_DFL` and raises immediately, killing the process before a graceful shutdown (window state save, flush, sidecar stop) can finish, and discarding that handler for any later signal. Before this PR an app with its own handler decided when to exit; in debug builds it no longer does, so dev-mode termination diverges from release. Nothing documents it: the `init()` rustdoc enumerates every other host-visible effect (socket path, bridge injection, bind failure never blocking the app) and says nothing about signals, and neither does `docs/src/content/docs/guides/plugin-setup.md`. This is the first process-global state the plugin touches.

**Fix:** Read the existing disposition with `sigaction(signum, NULL, &old)` before registering and skip the re-raise when it is not `SIG_DFL` — the app already owns the signal, only the unlink is the plugin's business. Either way, state on `init()` and in `plugin-setup.md` (§ Debug-only compilation) that debug Unix builds watch SIGINT/SIGTERM, unlink the socket and terminate the process themselves with 128+n.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR installs the watcher and the re-raise; the rustdoc and guide that describe host-visible behaviour were not updated

</details>

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.8</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -706,6 +755,58 @@ mod tests {
         assert!(!left, "RunEvent::Exit must unlink the socket (#194)");
     }
 
+    #[cfg(all(unix, not(target_os = "android"), debug_assertions))]
+    #[test]
+    fn sigterm_unlinks_socket_file() {
```
> **🟡 Medium · 🧪 Tests** — Only SIGTERM is exercised; the SIGINT arm the issue leads with has no test

The regression test raises only SIGTERM, so `signal(SignalKind::interrupt())` and the `_ = interrupt.recv() => libc::SIGINT` arm never execute — yet Ctrl+C is the case #217 and the CHANGELOG entry open with. `signum` is consumed only by `re_raise`, which returns early under `cfg!(test)`, so swapping the two select arms or watching the wrong `SignalKind` keeps this test green while Ctrl+C still leaves the socket behind.

**Fix:** Extract the body into a helper taking the signal number (`fn signal_unlinks_socket(signum: i32, tag: &str)`) and call it from two tests, `sigint_unlinks_socket_file` and `sigterm_unlinks_socket_file`, each with its own identifier so the socket paths do not collide.

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -706,6 +755,58 @@ mod tests {
...
+        // The plugin watcher registers from a task on Tauri's runtime, so the
+        // first raise can land before it is listening. Retry until it does.
+        let deadline = Instant::now() + Duration::from_secs(5);
+        while Instant::now() < deadline && path.exists() {
+            // SAFETY: `raise` is a libc entry point with no preconditions.
+            unsafe {
+                libc::raise(libc::SIGTERM);
+            }
+            std::thread::sleep(Duration::from_millis(20));
+        }
```
mpiton, line 828: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The SIGTERM raise loop is process-wide and releases the socket guard of every other mock-app test

`libc::raise(libc::SIGTERM)` goes to tokio's process-global registry, which broadcasts to every `Signal` stream in the binary — and `init()` now spawns one watcher per mock app. `normal_quit_unlinks_socket_file` (lib.rs:730) builds its own app whose guard is live between `build(context)` and `assert!(path.exists(), "plugin must bind a pathname socket")`; a raise from this loop (up to 250 over 5s) releases that guard and unlinks its socket, so the #194 test fails at random, or passes without `RunEvent::Exit` doing anything. libtest runs these on parallel threads and no test in `lib.rs` is serialized.

**Fix:** Put `sigterm_unlinks_socket_file` and the other mock-app tests that bind a socket in the same serial group with `#[serial]` — `serial_test` is already a dev-dependency and `server/windows.rs:570` uses it exactly this way.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the raise loop and the per-app watcher that make the existing #194 test vulnerable are both added here

</details>

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.85</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -119,6 +119,52 @@ fn release_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>) {
...
+
+        let (Ok(mut interrupt), Ok(mut terminate)) = (
+            signal(SignalKind::interrupt()),
+            signal(SignalKind::terminate()),
+        ) else {
+            tracing::warn!(
+                "tauri-pilot cannot watch SIGINT/SIGTERM, a signal leaves the socket behind"
+            );
+            return;
+        };
```
> **🔵 Low · 🐛 Correctness** — A half-failed registration leaves SIGINT hooked with no receiver, so Ctrl+C is ignored

Both `signal()` calls are evaluated before the `let ... else` match, so if `interrupt()` succeeds and `terminate()` fails, the `else` arm returns and drops the interrupt stream. Tokio never restores the default disposition once a signal is registered, so SIGINT stays handled by a handler nobody reads and the dev app ignores Ctrl+C for its whole lifetime — worse than the leak being fixed. The warning text ("cannot watch SIGINT/SIGTERM") also misreports the state, since SIGINT was in fact hooked.

**Fix:** Register each kind separately and keep watching whichever succeeded, or call `libc::signal(libc::SIGINT, libc::SIG_DFL)` before returning when the second registration fails, so Ctrl+C keeps its default kill.

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.8</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #228

> [!IMPORTANT]
> 💬 **Review posted** — 2 blocking issues (1 critical, 1 high) out of 9 findings.

**Actionable comments posted: 9**

|:---|---:|:---|
| 🔴 Critical | 1 | ship-stopper: security, data loss, broken behaviour |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/lib.rs` | 0 | 1 | 3 | 3 |
| `crates/tauri-plugin-pilot/Cargo.toml` | 1 | 0 | 0 | 1 |

### 🎯 Start here

1. 🔴 **tokio's `signal` feature is added without `macros`, so the crate no longer compiles on its own** — `crates/tauri-plugin-pilot/Cargo.toml:39`
2. 🟠 **`cfg!(test)` skips the re-raise, leaving it untested and the test binary deaf to SIGINT/SIGTERM** — `crates/tauri-plugin-pilot/src/lib.rs:160`
3. 🟡 **The plugin now decides how the host process dies, and no doc says so** — `crates/tauri-plugin-pilot/src/lib.rs:164`
4. 🟡 **Only SIGTERM is exercised; the SIGINT arm the issue leads with has no test** — `crates/tauri-plugin-pilot/src/lib.rs:760`
5. 🟡 **The SIGTERM raise loop is process-wide and releases the socket guard of every other mock-app test** — `crates/tauri-plugin-pilot/src/lib.rs:803`

<details>

| | |
|:---|:---|
| Commit reviewed | `276b1c5e48` |
| Base | `main` |
| Files changed | 3 (+113/-0) |
| Languages | rust |
| Reviewers | merged |

</details>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -119,6 +119,52 @@ fn release_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>) {
...
+        let signum = tokio::select! {
+            _ = interrupt.recv() => libc::SIGINT,
+            _ = terminate.recv() => libc::SIGTERM,
+        };
```
> **🔵 Low · 🎯 Product** — SIGHUP still leaves the socket behind when the terminal running `tauri dev` closes

The `select!` watches only `interrupt()` and `terminate()`. Closing the terminal that runs `cargo tauri dev`, or dropping an SSH/tmux session, sends SIGHUP, whose default disposition terminates the process just as silently and just as far ahead of `RunEvent::Exit`. That path still piles `tauri-pilot-{identifier}.sock` into `$XDG_RUNTIME_DIR` — the exact symptom #217 reports, reached by a route developers hit daily.

**Fix:** Add `signal(SignalKind::hangup())` to the let-else tuple and a `_ = hangup.recv() => libc::SIGHUP` arm to the `select!`. `release()` and `re_raise` need no change: SIGHUP's default disposition is also terminate, so the 128+n status stays correct.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR chooses the set of watched signals, and SIGHUP produces exactly the leak it fixes

</details>

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.85</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -119,6 +119,52 @@ fn release_slot<T>(slot: &std::sync::Mutex<GuardSlot<T>>) {
...
+/// Unlinks the socket when the app is killed with Ctrl+C or `kill` (#217).
+///
+/// SIGINT and SIGTERM end the process before tao emits `RunEvent::Exit`, so the
+/// guard in plugin state never drops and the socket file stays behind. Watching
+/// the signals suppresses their default kill, hence the re-raise.
+#[cfg(all(unix, not(target_os = "android"), debug_assertions))]
+fn spawn_signal_cleanup(guard: PilotGuard) {
+    tauri::async_runtime::spawn(async move {
+        use tokio::signal::unix::{SignalKind, signal};
+
```
> **🔵 Low · 🐛 Correctness** — The watcher is armed inside the spawned task, so a signal at startup still leaks the socket

`signal(SignalKind::interrupt())` and `signal(SignalKind::terminate())` run inside the spawned task, so between `setup` returning and the task's first poll the default disposition is still in force: a signal in that window kills the app and leaves the socket behind — #217 again, just narrower. The new test has to raise SIGTERM in a retry loop for exactly this reason, and that loop is what sprays the signal over the sibling tests.

**Fix:** Create both `Signal`s in `setup` (`tauri::async_runtime::block_on`, as the Windows bind already does because setup is not in a runtime context) and move them into the task. The watcher is then armed before `setup` returns, and the test can raise once and poll for the unlink.

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.8</sub>

crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -35,6 +35,8 @@ enigo = { version = "0.6.1", optional = true }
 
 [target.'cfg(unix)'.dependencies]
 libc = "0.2.189"
+# Signal watching is Unix-only (#217); Windows has no socket file to unlink.
```
> **🔵 Low · 📄 Docs** — The Cargo.toml comment says Windows has nothing to unlink, but it leaks an instance file the same way

"Windows has no socket file to unlink" is true of the pipe itself, but `RegistryGuard` (server/windows.rs:93) removes `%LOCALAPPDATA%/.../instances/{identifier}.json` on drop, and that drop also hangs off `RunEvent::Exit`. Ctrl+C on Windows therefore leaves a stale instance entry behind for the same reason the socket used to stay on Unix, so the comment reads as "Windows is fine" when the platform has the same gap by a different name.

**Fix:** Reword to what is actually true — signal watching is Unix-only for now; Windows leaks `instances/{identifier}.json` on Ctrl+C and is tracked separately — or open a follow-up issue and reference it here.

<sub>🤖 <b>mpiton-review</b> · merged · confidence 0.85</sub>

crates/tauri-plugin-pilot/Cargo.toml
mpiton: > tokio's `signal` feature is added without `macros`, so the crate no longer compiles on its own

Right, `cargo check -p tauri-plugin-pilot --lib` reproduced it here and the macos `Check iOS plugin` job died on the same line.

Fixed in `320c9c7`, but by dropping `select!` rather than adding `macros`: the watcher is now one stream and one spawned task per signal, so the only tokio feature the lib needs is still `signal`. Standalone `cargo check --lib` is green.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > `cfg!(test)` skips the re-raise, leaving it untested and the test binary deaf to SIGINT/SIGTERM

Fixed the disposition half in `320c9c7`: `re_raise` now runs `libc::signal(signum, SIG_DFL)` unconditionally and only guards `libc::raise` behind `cfg!(test)`, so the next signal reaches the binary instead of being eaten by a handler nobody reads.

Left the re-exec integration test out for now. The three in-crate tests do pin the watcher end (helper at `lib.rs`, driving SIGINT/SIGTERM/SIGHUP) — dropping a `SignalKind` from the array makes the run die of the raised signal instead of failing an assert, so the arms are not vacuous. What stays uncovered is the 128+n exit status, and a child-process harness is a bigger piece than this PR. Opening a follow-up for it.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > The plugin now decides how the host process dies, and no doc says so

Documented in `320c9c7`: `init()` has a `# Signals` section, and `plugin-setup.md` § Debug-only compilation says debug Unix builds watch SIGINT/SIGTERM/SIGHUP, unlink the socket and exit 128+n, including the note that a long graceful shutdown gets cut short.

Skipped the `sigaction` probe. It only answers "did someone install a handler *before* plugin `setup` ran", and `setup` runs during `Builder::build()`, i.e. before most app code that would register one — so an app installing its handler after `run()` reads as SIG_DFL and gets re-raised anyway, while an app that happens to register early silently loses the unlink. Wrong either way without an ordering contract we do not have. If a host needs to own the signal, the escape hatch is not to register the plugin, which the debug guard already makes easy.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > Only SIGTERM is exercised; the SIGINT arm the issue leads with has no test

Fixed in `320c9c7`. `signal_unlinks_socket_file(signum, tag)` holds the body, and three tests call it: `sigint_unlinks_socket_file`, `sigterm_unlinks_socket_file`, `sighup_unlinks_socket_file`, each with its own identifier suffix so the socket paths do not collide.

Checked they bite: deleting the `SignalKind::interrupt()` and `SignalKind::hangup()` entries from the watcher array makes the run end with `process didn't exit successfully: ... (signal: 1, SIGHUP: hangup)` instead of passing.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > The SIGTERM raise loop is process-wide and releases the socket guard of every other mock-app test

Fixed in `320c9c7`. `#[serial]` now covers `normal_quit_unlinks_socket_file` and the three signal tests, with a comment on the #194 one saying why.

The retry loop is gone too: the watchers are created in `setup` instead of inside the spawned task, so one `raise` is enough and the test just polls for the unlink.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > A half-failed registration leaves SIGINT hooked with no receiver, so Ctrl+C is ignored

Fixed in `320c9c7`. The tuple let-else is gone: each `SignalKind` is registered in its own loop iteration and gets its own task, so a failure on one kind logs and `continue`s while the streams already opened keep their receivers. No path drops a stream tokio has hooked. The warning now names the `signum` it applies to instead of claiming both are unwatched.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > SIGHUP still leaves the socket behind when the terminal running `tauri dev` closes

Added in `320c9c7`. `SignalKind::hangup()` / `libc::SIGHUP` is a third entry in the watcher array, and `sighup_unlinks_socket_file` covers it. `re_raise` needed no change, as you said — SIGHUP's default disposition terminates, so 128+n still holds.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > The watcher is armed inside the spawned task, so a signal at startup still leaks the socket

Fixed in `320c9c7`. `signal(kind)` now runs in `spawn_signal_cleanup` itself, under `tauri::async_runtime::block_on` since `setup` has no runtime context, and only the `recv().await` moves into the task. The watcher is registered before `setup` returns.

That also let the test's retry loop go — one `raise`, then poll for the unlink.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: > The Cargo.toml comment says Windows has nothing to unlink, but it leaks an instance file the same way

Reworded in `320c9c7`. The comment now says signal watching is Unix-only for now and that Windows has the same gap under another name — `instances/{identifier}.json`, since `RegistryGuard`'s drop also hangs off `RunEvent::Exit`. The CHANGELOG entry says the same. Follow-up issue for the Windows side coming.

crates/tauri-plugin-pilot/src/lib.rs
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> When `cfg!(test)` is true, `re_raise` returns before restoring `SIG_DFL`. Tokio keeps its Unix signal handler process-global after listeners are dropped.

Fixed in `320c9c7`, essentially your diff:

```rust
unsafe {
    libc::signal(signum, libc::SIG_DFL);
    if !cfg!(test) {
        libc::raise(signum);
    }
}
```

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -202,6 +248,9 @@ pub fn init<R: tauri::Runtime>() -> tauri::plugin::TauriPlugin<R> {
                     };
                     // Keep the guard in plugin state, not the server task (#194).
                     cleanup.hold_unix(guard);
+                    // A signal never reaches RunEvent::Exit (#217).
+                    #[cfg(not(target_os = "android"))]
+                    spawn_signal_cleanup(cleanup.clone());
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> When SIGINT/SIGTERM arrives after `bind` but before this spawned task first polls, the default action terminates the app and leaves the socket file.

Valid, fixed in `320c9c7`. `tokio::signal::unix::signal()` now runs inside `spawn_signal_cleanup` under `tauri::async_runtime::block_on` (`setup` is not in a runtime context), and only `recv().await` goes into the spawned task. The handler is installed before `setup` returns, and tokio's `Signal` reports a signal delivered between creation and the first `recv`, so the window is closed rather than narrowed.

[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> a Ctrl+C or a CI-timeout SIGTERM sent to the `cargo test` process for this crate is silently consumed instead of terminating the run; a hung test can no longer be stopped with SIGTERM and needs SIGKILL

Fixed in `320c9c7` by the second option you list. `re_raise` restores `SIG_DFL` even under `cfg!(test)` and only skips `libc::raise`, so a watcher consumes at most the signal it was built for and the next one gets the default kill.

Did not skip `spawn_signal_cleanup` for mock apps: the three signal tests drive the real watcher and would then be testing nothing.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: Follow-up for the child-process exit-status test is #230.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: Follow-up for the Windows instance file is #229, and the Cargo.toml comment now points at it (`516d817`).

Issue #232, opened 2026-09-20
bridge: raw JSC log sources, select: prefix on fill, response_size 0 for fetch


mpiton:
Split out of #218 (JS bridge part). Three independent nits in `crates/tauri-plugin-pilot/js/bridge.js`; they share a file, not a cause.

### 1. Log `source` keeps the raw JavaScriptCore frame

```
[14:23:07.351] error Unhandled rejection: Error: rej (@tauri://localhost:1:143)
[14:23:07.351] error Error: boom (undefined:1:91)
```

`firstAppFrame` (`bridge.js:287`) returns the matched frame verbatim. JSC writes anonymous frames as `@url:line:col`, so the leading `@` survives into the output. Strip it when the function name is empty.

The second line is a different path: the uncaught-error handler builds `where` from `event.filename` (`bridge.js:331`), and WebKitGTK reports the *string* `"undefined"` for code that came from `eval`. The `event.filename ? ... : ...` guard doesn't catch it because a non-empty string is truthy. Drop the file part when it's missing, empty, or the literal `"undefined"`, and keep the line:col.

### 2. `fill` on a `<select>` reports errors as `select:`

```
$ tauri-pilot fill 'select[name=role]' nope
Error: RPC error (-32603): Eval error: JavaScript error: select: no option matches "nope"
```

`bridge.js:959` hardcodes the `select:` prefix, but `fill` delegates to the same code path, so the prefix names a command the user never typed. Carry the calling command's name into the message.

### 3. `network` reports `response_size: 0` for `fetch`

```
$ tauri-pilot eval 'fetch("settings.html").then(r=>r.text()).then(t=>t.length)'
1380
$ tauri-pilot network --json      # fetch entry
... "response_size": 0 ...
```

The fetch wrapper reads `Content-Length` (`bridge.js:436`), which `tauri://` responses don't expose, and `|| 0` turns the miss into a confident zero. The XHR wrapper (`bridge.js:513`) already falls back to the actual response — `responseText.length`, `ArrayBuffer.byteLength`, `Blob.size` — which is why the same request reports `1380` through XHR.

`response.clone().blob().then(b => b.size)` gives the real size, at the cost of buffering the body and resolving the entry asynchronously. Returning `null` when the size is unknown is cheaper and at least stops the output from lying. Pick one; `null` needs the CLI to render a missing size without printing `0`.

Done when each of the three has a regression test, and `fill` on a `<select>` with no matching option names `fill` in the error.

PR #235, opened 2026-09-20
fix(bridge): raw JSC log sources, select: prefix on fill, response_size 0 for fetch (#232)

mpiton:
### 1. Log source keeps raw JavaScriptCore frame
- Strip leading `@` from anonymous JSC frames in logs
- Report `line:col` only when filename is missing or the string `undefined`
- Fixes output like `(@tauri://localhost:1:143)` and `undefined:1:91`

### 2. fill on <select> reports errors as select:
- Carry calling command name into error message
- `fill` now names `fill:`, `select` names `select:`

### 3. network reports response_size 0 for fetch
- Report `null` instead of `0` when Content-Length absent
- tauri:// responses don't expose Content-Length
- Avoids confident zero when size is actually unknown

## Testing
- All 138 JS bridge tests pass
- 4 new regression tests (one per issue, #1 has two)
- 241 Rust tests pass
- Clippy clean, fmt check clean

---
## Summary by cubic
Fixes bridge reporting bugs from #232, makes the JS bridge’s `node:test` suite run in CI, and adds regression tests for each fix.

**Bug Fixes**
- Anonymous JavaScriptCore frames no longer keep the dangling `@`, and eval’d errors drop `"undefined"` filenames, so sources report `tauri://localhost:1:143` and `1:91`.
- The option matcher now names the command actually run, so `fill` on a `<select>` reports `fill:` instead of `select:`.
- `fetch` and XHR recorders now report `response_size: null` whenever the size is unknown instead of `0`: missing or malformed `Content-Length`, unmeasurable XHR bodies, and error/timeout/abort with no response. An explicit `Content-Length: 0` still reports `0`.
- Breaking: `response_size` is now `number | null` in `--json`, the NDJSON `--follow` stream, and the MCP tool; the changelog and CLI reference document the new format.

<sup>Written for commit 6e85fffbf6e7fb7698f34b181df2abe2cdb07ace. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JavaScript error log source formatting, including anonymous stack frames.
  * Removed invalid `undefined` filename references from evaluation errors.
  * Corrected command names in unmatched select-option errors.
  * Reported unknown network response sizes as `null` instead of `0`.
  * Preserved accurate response sizes when valid content-length information is available.


mpiton (commented):
## 🧨 mpiton-review — PR #235

> [!IMPORTANT]
> 💬 **Review posted** — No blocking issue. 5 findings to consider.

**Actionable comments posted: 5**

|:---|---:|:---|
| 🔵 Low | 3 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs` | 0 | 0 | 1 | 1 |
| `crates/tauri-plugin-pilot/js/bridge.js` | 0 | 0 | 1 | 0 |
| `crates/tauri-plugin-pilot/js/bridge.select.test.mjs` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟡 **The fetch path reports an unknown size as null, the XHR path still reports 0** — `crates/tauri-plugin-pilot/js/bridge.js:450`
2. 🟡 **The new regression tests for #232 are never run by CI** — `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs:143`
3. 🔵 **eventSource's two new branches for a missing filename are untested** — `crates/tauri-plugin-pilot/js/bridge.errors.test.mjs:186`
4. 🔵 **Third test for the unmatched-option throw instead of tightening the two that exist** — `crates/tauri-plugin-pilot/js/bridge.select.test.mjs:218`
5. 🔵 **response_size becomes number|null in --json and MCP output with no breaking-change note** — `CHANGELOG.md:77`

<details>

| | |
|:---|:---|
| Commit reviewed | `114b7d8bcb` |
| Base | `main` |
| Files changed | 5 (+146/-10) |
| Languages | javascript |
| Reviewers | documentation, javascript, maintainability, product-logic, quality, security, tests |

</details>

crates/tauri-plugin-pilot/js/bridge.js
```diff
@@ -433,7 +443,11 @@
     return _originalFetch(input, init).then(function(response) {
       const duration_ms = Date.now() - timestamp;
       const status = response.status;
-      const responseSize = parseInt(response.headers.get("Content-Length") || "0", 10) || 0;
+      // tauri:// responses carry no Content-Length. Reading the real size
+      // would mean cloning and buffering every body, so report null — an
+      // unknown size, not a confident 0 (#232).
+      const contentLength = parseInt(response.headers.get("Content-Length"), 10);
+      const responseSize = Number.isFinite(contentLength) ? contentLength : null;
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — The fetch path reports an unknown size as null, the XHR path still reports 0

`response_size` now means two different things depending on which API the page used. The XHR recorder in the same file keeps the old shape: `onLoad` computes `const cl = parseInt(this.getResponseHeader("Content-Length") || "0", 10) || 0` (line 528) and falls back to `cl` whenever the body is not measurable — with `responseType` set to `"json"` or `"document"` the response is a plain object, so neither the `ArrayBuffer` nor the `Blob` branch applies. A `tauri://` XHR with `responseType = "json"` and no `Content-Length` is exactly the case #232 reports, and it still records a confident `0`. The fetch reject path (line 476) and the XHR error/timeout/abort pushes (lines 535-537) also write `0` for a size nobody measured. A consumer that learns to read `null` as unknown is still lied to by every other path.

**Fix:** Give every recorder one unknown-size convention. Hoist the parse next to `bodySize`: `function headerSize(raw) { const n = parseInt(raw, 10); return Number.isFinite(n) ? n : null; }`, use it in the fetch wrapper and as the XHR `onLoad` fallback (`r instanceof ArrayBuffer ? r.byteLength : r instanceof Blob ? r.size : headerSize(this.getResponseHeader("Content-Length"))`), and report `null` rather than `0` on the paths where no response arrived at all. The text branch can keep `0`, since an empty `responseText` really is zero bytes. Then add the XHR counterpart test — `XMLHttpRequestStub.prototype.getResponseHeader` in bridge.network.test.mjs:52 hard-codes `"0"`, the same fiction the fetch stub used to carry, so it must return `null` for the case to be reachable.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** This PR establishes null-for-unknown on `response_size`, but only on one of the four code paths that write the field, so it owns the split meaning.

</details>

<sub>🤖 <b>mpiton-review</b> · 4 reviewers agree: maintainability, product-logic, quality, tests · confidence 1.0</sub>

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -140,6 +140,52 @@ test("a JavaScriptCore stack names the throwing frame, not the caller", () => {
   assert.doesNotMatch(entry.source, /main\.js/, "the throwing frame beats the caller and the event location");
 });
 
+test("an anonymous JavaScriptCore frame drops its leading @ in source", () => {
```
mpiton, line 143: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The new regression tests for #232 are never run by CI

`.github/workflows/ci.yml` has no step that runs `node --test` — grep finds the command nowhere in the repo — and both `paths:` filters only fire on Rust sources, `Cargo.toml`, `Cargo.lock` and `ci.yml`. This PR touches only `bridge.js`, three `.test.mjs` files and `CHANGELOG.md`, so CI does not start at all. The cases that genuinely fail against the pre-PR bridge are enforced by nothing, and the PR description's "All 138 JS bridge tests pass" is a local claim no automation can repeat.

**Fix:** Add a `node --test 'crates/tauri-plugin-pilot/js/*.test.mjs'` step to the `check` job (ubuntu alone is enough) and add `crates/tauri-plugin-pilot/js/**` to both `paths:` filters. Use the glob, not the directory: `node --test crates/tauri-plugin-pilot/js/` fails on Node 24.18 with MODULE_NOT_FOUND. Worth adding the same command to CONTRIBUTING, which currently documents only `cargo test --workspace`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the whole regression safety net for #232 is the tests this PR adds, and no automation runs them, so the three fixed bugs can regress with a green PR.

</details>

crates/tauri-plugin-pilot/js/bridge.errors.test.mjs
```diff
@@ -140,6 +140,52 @@ test("a JavaScriptCore stack names the throwing frame, not the caller", () => {
...
+  const [entry] = pilot.consoleLogs({ level: "error" });
+  assert.equal(entry.source, "1:91");
+});
+
+test("an uncaught error with no location at all has a null source", () => {
+  const { pilot, win } = loadBridge();
+  win.dispatch("error", { message: "Error: boom" });
+
+  const [entry] = pilot.consoleLogs({ level: "error" });
+  assert.equal(entry.source, null);
```
mpiton, line 186: > [!NOTE]
> **🔵 Low · 🧪 Tests** — eventSource's two new branches for a missing filename are untested

`eventSource` flips two cases the added tests never touch: `{ message, lineno: 12, colno: 3 }` with no `filename` goes from `null` (old) to `"12:3"` (new), and `{ message, filename: "undefined" }` with no line or column goes from `"undefined:0:0"` to `null`. The suite pins only `filename: "undefined"` *with* a location and the fully empty event, so either branch can flip back without a failure. The first case is also a behaviour change the CHANGELOG does not mention — an event with a location but no filename used to have no source and now has one.

**Fix:** Add two cases beside this one: `win.dispatch("error", { message: "Error: boom", lineno: 12, colno: 3 })` asserting `"12:3"`, and `win.dispatch("error", { message: "Error: boom", filename: "undefined" })` asserting `null`.

crates/tauri-plugin-pilot/js/bridge.select.test.mjs
```diff
@@ -193,3 +193,26 @@ test("type rejects a select target", () => {
...
+    /^Error: fill: no option matches "nope"$/,
+  );
+
+  const selected = makeSelect(options);
+  const selectPilot = loadBridge({ queryResult: selected });
+  assert.throws(
+    () => selectPilot.select({ selector: "select[name=role]", value: "nope" }),
+    /^Error: select: no option matches "nope"$/,
+  );
+});
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Third test for the unmatched-option throw instead of tightening the two that exist

`select throws when no option matches value or label` (line 134) and `fill on select uses option matching and throws when nothing matches` (line 160) already drive both commands through the unmatched path with the same `makeSelect` fixture; they just assert the loose `/no option/i`. The new test rebuilds both fixtures and both pilots to assert the prefix those two leave unpinned, so the file ends up with three tests for one throw — and the new one drops the `selectedIndex === -1` assertion the other two carry.

**Fix:** Delete the new test and tighten the existing assertions instead: `/^Error: select: no option matches "zzz"$/` at line 144 and `/^Error: fill: no option matches "zzz"$/` at line 178. The regression stays pinned with about 20 fewer lines and keeps the `selectedIndex` check.

CHANGELOG.md
```diff
@@ -58,6 +58,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `undefined:1:91`. WebKitGTK sets the error event's `filename` to the string
+  `"undefined"` there, which the old truthiness check let through; the file
+  part is now dropped and only `line:col` is reported. An event with no
+  location at all still has no source. [#232]
+
+- `fill` on a `<select>` with no matching option now names `fill` in the
+  error instead of `select:`. Both commands share the option matcher, which
+  hardcoded the prefix of a command the user may never have typed. [#232]
+
+- `network` reports `response_size: null` instead of `0` when the size is
```
> **🔵 Low · 📄 Docs** — response_size becomes number|null in --json and MCP output with no breaking-change note

`network --json`, the NDJSON follow output and the MCP `network` tool hand the bridge entry through verbatim, so any script or agent that read `response_size` as an integer now gets `null`. The entry sits under `Fixed` with no `**Breaking:**` marker, unlike the #216 entry which labels a smaller output change, and `response_size` appears nowhere under `docs/` — the `network` section of `docs/src/content/docs/reference/cli.md` has no **JSON output format** block at all, while the `logs` section right above it (cli.md:1139) does. Nothing outside this changelog line tells a consumer the field is now tri-state.

**Fix:** Mark the entry `**Breaking:**` the way the #216 entry does, and give the `network` section of `docs/src/content/docs/reference/cli.md` the same **JSON output format** block `logs` has, stating that `response_size` is `null` when the size is unknown.

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.75</sub>

Issue #233, opened 2026-09-20
plugin: answer url without the bridge, list valid labels on unknown --window


mpiton:
Split out of #218 (plugin part). Both nits hit when you are already lost and the tool refuses to tell you where you are.

### 1. `url` fails on a page without the bridge

After `navigate` to a foreign origin, `url` fails like every other bridge command, even though the plugin knows the answer:

```
$ tauri-pilot url
Error: RPC error (-32603): no pilot bridge on the current page (http://127.0.0.1:8765/). ...
$ tauri-pilot windows
main       http://127.0.0.1:8765/            Pilot DnD Test
```

`url` is routed to the bridge with the rest of the eval-backed commands (`crates/tauri-plugin-pilot/src/handler.rs:217`), but `WebviewWindow::url()` answers it without touching the page — that is how `windows` already gets it. Answering `url` natively makes it work on foreign origins, which is exactly when you need it.

Related: #153 (foreign-origin pages, ACL denies `__callback`).

### 2. `--window` with an unknown label doesn't list the valid ones

```
$ tauri-pilot --window nope title
Error: RPC error (-32603): Window 'nope' not found
```

`TauriWebviews::target` (`crates/tauri-plugin-pilot/src/webview.rs:99`, same at `:321`) knows the full label set — it builds a `BTreeMap` of them a few lines up — and throws it away. `screenshot_native` already returns `available_windows` in the error data for the same class of mistake (#149); do the same here so a typo doesn't cost a `windows` round-trip.

Done when `url` returns the current URL on a page with no bridge, and an unknown `--window` label surfaces the available labels the way #149 does for `screenshot_native`.

PR #236, opened 2026-09-20
fix(plugin): answer url without bridge, list valid labels on unknown --window

mpiton:
• url answered from runtime (WebviewWindow::url), not bridge — survives foreign origin ACL deny (#153/#233)

• Unknown --window label lists available_windows in error.data (#233) 

• handle_press uses shared target() helper, surfaces label list

Tests: url answer without page touch; unknown label surfaces window list; press error message updated

---
## Summary by cubic
Answers `url` from the runtime instead of the injected bridge, so it still works on foreign origins the ACL denies; unknown `--window` labels now get the same `WINDOW_NOT_FOUND` envelope `screenshot_native` uses instead of a generic internal error.

- `url` never touches the page, and errors if the runtime cannot report a URL.
- Unknown-label errors are code `-32602` and carry `data.error`, `data.message`, and `available_windows` rows (`{label, url, title}`), even when empty.
- `press` uses the same lookup, so it reports `Window 'nope' not found` like the other commands.
- `RPC_INTERNAL_ERROR` is imported outside the `press` feature gate, so Android/iOS `--no-default-features` builds compile.

<sup>Written for commit 52df0636b60d034dad1b7b178c3d114e78f86820. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - The `url` command now works on foreign origins by reading the current WebView URL directly.
  - Invalid window labels now include a list of available window choices.

- **Bug Fixes**
  - Window-targeted `press` errors now use the standardized “Window '<label>' not found” message.

- **Documentation**
  - Updated CLI documentation to explain URL behavior and available-window guidance for invalid labels.


docs/src/content/docs/reference/cli.md
```diff
@@ -1008,6 +1008,10 @@ Get the current page URL.
 tauri-pilot url
 ```
 
+The plugin answers this one from the webview itself, not from the injected
+bridge, so it still works on a page the bridge cannot drive — a foreign origin
+missing from `remote.urls`, where every other command fails.
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — cli.md claims every command but `url` fails on a foreign origin, contradicting its own `navigate` section

`dispatch` routes `ping`, `windows.list`, `press`, `navigate` and `screenshot_native` past the bridge as well (handler.rs:191-266; `handle_press` only focuses and injects OS keys, `handle_navigate` explicitly runs the script when the page has no bridge, handler.rs:665-669), so "where every other command fails" is wrong. The same file, 13 lines above, tells the reader "Run `tauri-pilot navigate` with an app URL to get the session back" (cli.md:998-999) — an agent that lands on a denied origin and reads the `url` section concludes it is stranded and never tries the documented way out.

**Fix:** Narrow the claim to bridge-driven commands and name the survivors, or point at the `navigate` paragraph instead of generalising.

```suggestion
The plugin answers this one from the webview itself, not from the injected
bridge, so it still works on a page the bridge cannot drive — a foreign origin
missing from `remote.urls`, where the bridge commands fail. `windows`,
`navigate` and `press` do not go through the bridge either and keep working
there; `navigate` back to an app URL is the way out.
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the sentence is added by this diff

</details>

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -826,15 +826,36 @@ fn origin_moved_error(checked: &tauri::Url, now: &tauri::Url) -> RpcError {
...
 ) -> Result<Box<dyn TargetWindow + 'a>, RpcError> {
-    webviews.target(window).map_err(|e| RpcError {
+    webviews.target(window).map_err(|e| {
+        let available = webviews.list();
+        RpcError {
+            code: -32603,
+            message: e,
+            data: (!available.is_empty())
+                .then(|| serde_json::json!({"available_windows": available})),
+        }
```
> **🟡 Medium · 🧱 Maintainability** — The window-not-found error claims parity with `screenshot_native` but keeps -32603 and no domain code

`screenshot_native` answers the same class of failure through one builder — `rpc_error(RPC_INVALID_PARAMS, codes::WINDOW_NOT_FOUND, …)` (screenshot/ipc.rs:43-52, 292-300) — so its payload carries `-32602`, `data.error = "WINDOW_NOT_FOUND"` and `data.message` next to `available_windows`. `target()` copies the `available_windows` key and nothing else: still `-32603`, still no domain code. A consumer other than this CLI (the MCP path, a script keying on the code) therefore cannot tell "you typed a label that does not exist" from "the plugin failed internally", and the only discriminator left is the message text, which the new assertion at handler.rs:1120 (`assert_eq!(err.message, "Window 'settings' not found")`) now pins as the contract.

**Fix:** Answer the `Some(label)` case with the screenshot envelope: `RPC_INVALID_PARAMS` plus `data.error = WINDOW_NOT_FOUND` alongside `available_windows`, keeping `-32603` for `No webview available`, which really is internal. `test_dispatch_press_with_unknown_window_errors` (handler.rs:1119) and the new label test assert the code, so both move with it. Same sweep: the CHANGELOG says this matches the `screenshot_native` payload, but the rows there are `{window_id, owner, title, layer}` and here `{label, url, title}` — name the fields instead of claiming parity, and update the `WindowInfo` doc comment (webview.rs:10), which still calls the struct the `windows.list` shape although it is now the error payload too.

crates/tauri-plugin-pilot/src/handler.rs
```diff
@@ -826,15 +826,36 @@ fn origin_moved_error(checked: &tauri::Url, now: &tauri::Url) -> RpcError {
...
+    webviews: &dyn Webviews,
+    window: Option<&str>,
+) -> Result<serde_json::Value, RpcError> {
+    let url = target(webviews, window)?.url().ok_or_else(|| RpcError {
         code: -32603,
-        message: e,
+        message: "the runtime cannot report the URL of the current page".to_owned(),
         data: None,
-    })
+    })?;
```
mpiton, line 875: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Both new branches go untested: `url` with no runtime URL, and the empty `available_windows` guard

`handle_url` introduces an error the bridge route never produced: `TargetWindow::url()` is `None` whenever `current_url`'s `catch_unwind` trips or the runtime call returns `Err` (webview.rs:124-128), and the caller then gets -32603 "the runtime cannot report the URL of the current page". No test dispatches `url` against a window with no URL, so swapping that `ok_or_else` for `unwrap_or_default()` would answer an empty string with the suite green. Same for the `(!available.is_empty())` guard at line 842: the only zero-window path is `test_dispatch_press_with_unknown_window_errors` (handler.rs:1105), which asserts the code and message only.

**Fix:** Two short tests next to `test_dispatch_url_answers_without_the_bridge`: dispatch `url` against `FakeWebviews::window("main", None)` and assert -32603 with that message; dispatch `url` with `{"window": "nope"}` against `FakeWebviews::default()` and assert `err.data.is_none()`. Putting the second one there rather than in the press test keeps it running without the `press` feature.

mpiton (commented):
## 🧨 mpiton-review — PR #236

> [!IMPORTANT]

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/handler.rs` | 0 | 0 | 1 | 1 |
| `crates/tauri-pilot-cli/src/mcp.rs` | 0 | 0 | 0 | 1 |
| `crates/tauri-plugin-pilot/js/bridge.js` | 0 | 0 | 0 | 1 |
| `docs/src/content/docs/guides/architecture.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟡 **cli.md claims every command but `url` fails on a foreign origin, contradicting its own `navigate` section** — `docs/src/content/docs/reference/cli.md:1013`
2. 🟡 **The window-not-found error claims parity with `screenshot_native` but keeps -32603 and no domain code** — `crates/tauri-plugin-pilot/src/handler.rs:844`
3. 🔵 **Both new branches go untested: `url` with no runtime URL, and the empty `available_windows` guard** — `crates/tauri-plugin-pilot/src/handler.rs:857`

<details>
<summary>🧭 Findings outside the diff or over the inline cap (3)</summary>

- 🔵 **MCP `url` tool description does not mention the bridge-free read the change adds** — `crates/tauri-pilot-cli/src/mcp.rs:1102` — The value of this change is discoverability: when every command fails on a foreign origin, the caller has to know `url` still answers. The PR adds that paragraph to the CLI reference but leaves the MCP `ToolSpec` description at "Get the current page URL." — and the MCP tool descriptions are the only documentation an agent sees. Other specs in this same list do carry that kind of nuance (`type`: "E
  - *Fix:* Extend the ToolSpec description to something like "Get the current page URL. Read from the webview itself, so it answers even on a page the bridge cannot drive."
  - *Link to this PR:* this PR gives `url` a capability no other read has and leaves the model-facing description unchanged
- 🔵 **`__PILOT__.url` is dead in the injected bridge now that dispatch answers `url` itself** — `crates/tauri-plugin-pilot/js/bridge.js:1425` — `url` left the `handle_eval_method` arm (handler.rs:216-217), and `build_bridge_call` is only reached for the methods still listed there, so nothing in the repo builds `window.__PILOT__.url(...)` any more. `state()` reads `window.location.href` inline rather than calling it, and no JS test covers it, so the function plus its export at bridge.js:1972 ship into every page with no caller.
  - *Fix:* Delete `url()` and its `__PILOT__` entry, or keep it with a one-line comment saying it stays as a hand-written `eval` helper, so the next reader does not re-wire `url` back through the bridge.
  - *Link to this PR:* this PR routes `url` to `handle_url` instead of the bridge, which is what leaves the JS entry point without a caller
- 🔵 **`Origins without a callback` still lists only `navigate` as surviving a denied origin** — `docs/src/content/docs/guides/architecture.md:110` — This section is the canonical explanation of the ACL limitation and it already enumerates the exception — "`navigate` is not origin-pinned: it is the way back from a foreign page" (line 114). After this PR `url` is a second exception, answered from the runtime with no script sent, but the section still reads as if every command goes through the read-URL-then-eval path described here.
  - *Fix:* Add one sentence to the section: `url` is answered from the target window's runtime URL and never sends a script, so it works on an origin that cannot call back.
  - *Link to this PR:* this PR moves `url` off the bridge and this guide is the page that documents which commands need one

</details>

<details>

| | |
|:---|:---|
| Commit reviewed | `bc61a9fc50` |
| Base | `main` |
| Files changed | 3 (+99/-12) |
| Languages | rust |

</details>

docs/src/content/docs/reference/cli.md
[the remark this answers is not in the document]
mpiton: @coderabbitai resolve

> Limit this statement to bridge-dependent commands. `windows.list` and `url` use runtime data without the page bridge.

Right, the sentence overstated it. Fixed in `c13e356`. `ping`, `windows.list`, `press` and `screenshot_native` never reach `handle_eval_method` (handler.rs:193-266), and `handle_navigate` resolves the destination through the runtime and falls back to the destination hello when the departing page has no bridge (handler.rs:663-670). The paragraph now says the bridge commands fail there, names `windows`/`navigate`/`press` as the survivors, and points at `navigate` to an app URL as the way back — the same escape hatch the `navigate` section already documents 13 lines above.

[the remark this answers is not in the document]

> The claim "where every other command fails" is broader than the code: `ping` and `windows.list` also work without the bridge

Correct, fixed in `c13e356`. Took the narrowing but went one step further than the suggestion: the paragraph now names the survivors (`windows`, `navigate`, `press`) and points at `navigate` to an app URL, because an agent reading only the `url` section otherwise concludes it is stranded. `navigate` is the case worth spelling out — from a bridgeless page it resolves the destination off the runtime URL and waits on the destination hello (handler.rs:663-670), so it is the way out of a denied origin.

mpiton: > cli.md claims every command but `url` fails on a foreign origin, contradicting its own `navigate` section

Fixed in `c13e356`, took the suggestion almost verbatim. The contradiction was the real cost here — an agent that lands on a denied origin and reads only the `url` section walks away thinking it is stranded, with the way out sitting 13 lines up.

Wording differs in one place: `navigate` to an app URL is the way back, rather than "back to an app URL", since from a foreign origin it is not going back anywhere it has been.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > The window-not-found error claims parity with `screenshot_native` but keeps -32603 and no domain code

Valid, fixed in `c13e356`. `target()` now splits on `window`: a label that names none gets `RPC_INVALID_PARAMS` with `data.error = codes::WINDOW_NOT_FOUND` (reusing the constant from `screenshot/ipc.rs`, not a fresh literal), `data.message`, and `available_windows`; `No webview available` keeps `-32603`, since that is the app having no window rather than a bad request. `test_dispatch_press_with_unknown_window_errors` moves to -32602 and the label test now asserts the code and the domain code too; `test_dispatch_press_without_any_window_errors` stays on -32603 and still passes, which is the split doing its job.

One deviation: the empty-list guard is gone, so the key is always present. An absent `available_windows` reads as "the plugin did not look", and `screenshot_native` emits the list unconditionally too (ipc.rs:296-300). That also removes the untested branch flagged in the sibling thread.

CHANGELOG now names both row shapes instead of claiming parity, and the `WindowInfo` doc comment says it is the error payload as well.

crates/tauri-plugin-pilot/src/handler.rs
mpiton: > Both new branches go untested: `url` with no runtime URL, and the empty `available_windows` guard

Both covered in `c13e356`.

`test_dispatch_url_errors_when_the_runtime_reports_no_url` dispatches against `FakeWebviews::window("main", None)` and asserts -32603 with the exact message, so swapping the `ok_or_else` for `unwrap_or_default()` now fails instead of answering `""`.

The second one changed shape, because the guard it was meant to pin is gone — the envelope work in the sibling thread made `available_windows` unconditional. `test_dispatch_unknown_window_lists_nothing_when_the_app_has_none` dispatches `url` with `{"window": "nope"}` against `FakeWebviews::default()` and asserts -32602, `data.error = WINDOW_NOT_FOUND` and `available_windows == []`. Kept out of the press tests as suggested, so it runs without the feature.

Issue #221, opened 2026-09-19
CI: build against the dependency floors declared in Cargo.toml


mpiton:
#210 shipped because the CLI manifest allowed `rmcp` 3.2.0 while the code needed 3.4.0. The same drift was fixed once before for rmcp 1.7.0 (see "Pin the `rmcp` manifest floor at `1.7.0`" in `CHANGELOG.md`). CI can't catch it: `Cargo.lock` is gitignored, so every run resolves the newest compatible versions. Nothing ever builds against the floors a consumer can actually end up with.

## Proposal

Add a job that resolves direct dependencies to the lowest versions the manifests allow, then checks the workspace:

```
cargo +nightly update -Z direct-minimal-versions
cargo check --workspace --all-targets
```

`-Z direct-minimal-versions` lowers direct dependencies only, so a transitive crate with a bad floor of its own can't turn the job red. Nightly is only needed for the `update` step; the check can stay on 1.95.0.

## Where it stands today

The resolve step fails before anything compiles:

```
error: failed to select a version for `serde_json`.
    ... required by package `tauri v2.0.0`
package `tauri` depends on `serde_json` with feature `raw_value` but `serde_json` does not have that feature.
  previously selected package `serde_json v1.0.0`
    ... which satisfies dependency `serde_json = "^1"` of package `tauri-pilot-cli v0.7.3`
```

A lot of floors are major-only: `serde_json = "1"`, `tokio = "1"` and `tracing = "0.1"` in `[workspace.dependencies]`, `tauri = "2"`, `thiserror = "2"`, `clap = "4"`, `tracing-subscriber = "0.3"`, `indicatif = "0.18"`, `toml = "1"`, among others. Each one needs raising to a version the code really builds against. Expect a few rounds: raise the floor the resolver or compiler names, rerun.

## Scope

- Raise floors until `cargo check --workspace --all-targets` passes at direct minimal versions. Linux only is enough to start, like the Android check step.
- Add the job to `ci.yml` and add it to the `needs` of `CI Success`.
- CHANGELOG entry for the raised floors, since consumers see them.

Issue #220, opened 2026-09-19
CI clippy skips test targets, and 5 lints already fail with --all-targets


mpiton:
Both Clippy steps in `.github/workflows/ci.yml` run `cargo clippy --workspace -- -D warnings`. That lints lib and bin targets only, so `#[cfg(test)]` code is never linted. Two PRs merged on 2026-09-18 landed test code that fails the workspace lint set, and CI stayed green.

Found while checking #210 (Linux, commit 61ba190).

## Reproduce

Same result on 1.95.0 (the CI toolchain) and 1.98.0:

```
$ cargo clippy --workspace --all-targets --keep-going -- -D warnings
error: used underscore-prefixed binding
    --> crates/tauri-pilot-cli/src/main.rs:1783:14
     |
1783 |         drop(_listener);
     |              ^^^^^^^^^
...
error: unused import: `TargetWindow`
   --> crates/tauri-plugin-pilot/src/webview.rs:418:47
```

## What fails

- `crates/tauri-pilot-cli/src/main.rs:1783`, `1809`, `1810`, `1832`: `clippy::used_underscore_binding`, which the workspace `pedantic` level enables. The socket tests bind `_listener`, `_xdg_listener` and `_tmp_listener`, then `drop()` them explicitly. The bindings are used, so the underscore prefix goes. Came in with #206.
- `crates/tauri-plugin-pilot/src/webview.rs:418`: the `TargetWindow` trait import in the `tests` module is unused, with or without the `press` feature. Came in with #203.

## Fix

- Rename the three bindings and drop `TargetWindow` from the `use super::{...}` list.
- Add `--all-targets` to both Clippy steps in `ci.yml` and to the two clippy lines in `.github/pull_request_template.md`, so test code stays linted from here on.

PR #238, opened 2026-09-20
ci: build against dependency floors at direct minimal versions (#221)

### LINKED ISSUE #210: CLI rmcp floor 3.2.0 is below what the code needs (ServerConfig requires rmcp 3.4.0)
The CLI manifest says `rmcp = "3.2.0"`, but the code has needed rmcp 3.4.0 since #197 renamed `ServerInfo` to `ServerConfig`. Any build that resolves rmcp to 3.2.x or 3.3.x fails to compile.

Found during end-to-end QA of 0.8.0 (Linux, commit 61ba190). CI doesn't see it because `Cargo.lock` is gitignored, so every CI run resolves the newest 3.x.

## Reproduce

```
$ cargo update -p rmcp --precise 3.3.0
 Downgrading rmcp v3.4.0 -> v3.3.0
$ cargo build -p tauri-pilot-cli
error[E0432]: unresolved import `rmcp::model::ServerConfig`
  --> crates/tauri-pilot-cli/src/mcp.rs:14:82
   |
14 |         JsonObject, ListToolsResult, PaginatedRequestParams, ServerCapabilities, ServerConfig,
```

It hits anyone with an older local lockfile, and any consumer that pins rmcp below 3.4 elsewhere in its dependency tree.

## Fix

- `cargo add rmcp@3.4.0 -p tauri-pilot-cli` (`crates/tauri-pilot-cli/Cargo.toml:33`).
- The `[Unreleased]` "Changed" entry in `CHANGELOG.md:280` says `rmcp` 3.1.4 -> 3.2.0. Update it to 3.4.0.

This should land before tagging 0.8.0.

mpiton:
Raise dependency floors to versions consumers can actually build against.
Floors were major-only, allowing outdated lockfiles to resolve versions
the code does not compile with:

- tauri 2 → 2.11.3: paired with tauri-plugin 2.6.3 floor
- tokio 1 → 1.37: Receiver::is_closed (used by rmcp 3.4.0)
- tracing 0.1 → 0.1.23: #[instrument] expansion (rmcp issue)
- serde 1 → 1.0.225, serde_json 1 → 1.0.127: toml 1.0.0 + schemars
- clap 4 → 4.3.4: Error::exit_code

Add minimal-versions CI job: resolves direct dependencies with
-Z direct-minimal-versions (nightly) and checks workspace on Linux.
Transitive crates stay at newest, so bad floors elsewhere cannot break it.

Fixes: #210, #221

## Summary by CodeRabbit

* **Compatibility**
  * Updated minimum supported versions for core Rust dependencies and the command-line interface tooling.
  * Preserved existing dependency features and behavior while establishing clearer dependency floors.

* **Quality Improvements**
  * Added validation against minimum dependency versions to help ensure reliable builds across supported environments.

* **Documentation**
  * Updated the unreleased changelog with dependency requirements, validation coverage, and the missing issue reference.

---
## Summary by cubic
Raises dependency floors from broad major-only ranges to versions the workspace actually supports, including Tauri 2.11.3+ for consumers. Adds required cross-platform CI that resolves direct dependencies at their minimums, catching incompatibilities before consumers do (#210, #221).

**Dependencies**
- Sets floors for `tokio` 1.37, `tracing` 0.1.41, `serde` 1.0.225, `serde_json` 1.0.127, and `clap` 4.3.4.
- Raises macOS `image` to 0.25.2 for `image::ImageReader` and `tracing-subscriber` to 0.3.20 for RUSTSEC-2025-0055.
- Updates the README and getting-started guide, and adds the minimal-versions check to both contributing guides.

**CI**
- Runs the check on Linux, macOS, and Windows with `nightly-2026-09-01` for resolution and Rust 1.95.0 for compilation.
- Keeps transitive dependencies at their newest versions and uses a separate cache.
- Makes `CI Success` fail when any required job fails, is cancelled, or is skipped.

<sup>Written for commit 3cfd5e27aeb9abd103a50ea667617f08433f0572. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -62,7 +62,7 @@ image = { version = "0.25", default-features = false, features = ["png"] }
```
mpiton, line 64: > [!WARNING]
> **🟠 High · 🐛 Correctness** — image floor 0.25 resolves to 0.25.0, which has no `image::ImageReader`

`image = "0.25"` resolves to 0.25.0, which exports only `image::io::Reader`; the crate-root `ImageReader` alias first ships in 0.25.2 (`CHANGES.md`: "Added `image::ImageReader` as an alias"). `screenshot/ipc.rs:496` calls `image::ImageReader::open`, so a macOS consumer whose lockfile holds `image` 0.25.0 or 0.25.1 fails to build the plugin, the #210 failure mode this PR set out to close. The Linux-only CI job cannot see it because `image` sits under `[target.'cfg(target_os = "macos")'.dependencies]`.

**Fix:** Raise the floor to `image = { version = "0.25.2", default-features = false, features = ["png"] }` and add it to the CHANGELOG list of raised floors.

<details>
<summary>🔎 Details</summary>

- **Actual location:** line 61 (comment anchored to the nearest line inside the diff)
- **Why it belongs to this PR:** The PR's stated job is to raise every floor that was major/minor-only so a consumer's lockfile cannot hold a version the code does not compile against; `image = "0.25"` is one of those floors and it was left behind while `tauri`, `tokio`, `tracing`, `serde`, `serde_json` and `clap` were raised.

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: quality, tests · confidence 1.0</sub>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
...
+  # and never sees the floors `Cargo.toml` declares. This job does: it resolves
+  # direct dependencies to the lowest versions the manifests allow, which is
+  # what a consumer with an old lockfile can end up with (#210, #221).
+  # Transitive crates stay at their newest, so a bad floor in someone else's
+  # manifest cannot turn it red. When it fails, raise the floor cargo or rustc
+  # names. Linux only, like the Android check. No cache: the lockfile here must
+  # never leak into `check`, and this job is not the slow one.
+  minimal-versions:
+    name: Direct minimal versions
+    runs-on: ubuntu-latest
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🐛 Correctness** — Minimal-versions job is Linux-only, so no platform-gated floor is ever compiled

`cargo check` on Linux never builds the `cfg(windows)` and `cfg(target_os = "macos")` dependency blocks, so the floors `windows = "0.62"`, `core-foundation = "0.10"`, `core-graphics = "0.25"` and `image = "0.25"` are resolved into the lockfile but never compiled. The `image` floor is the proof that the gap is not theoretical: the job is green today while the plugin does not build on macOS at its own declared floor.

**Fix:** Run the job on a matrix (`ubuntu-latest`, `macos-latest`, `windows-latest`) like `check` does, or at minimum add macOS, which owns three platform-only floors. The `cargo update -Z direct-minimal-versions` + `cargo +1.95.0 check --all-targets --locked` pair works unchanged per OS; only the `apt-get` step needs the `if: runner.os == 'Linux'` guard `check` already uses. If the job stays Linux-only, say so in the job comment and reword the CHANGELOG claim that a floor "can no longer drift below the code".

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** The job is added by this diff and is the only thing that exercises the floors; its platform scope decides which floors the gate can actually protect.

</details>

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.95</sub>

mpiton (commented):
## 🧨 mpiton-review — PR #238

> [!IMPORTANT]
> 💬 **Review posted** — 1 blocking issue (0 critical, 1 high) out of 10 findings.

**Actionable comments posted: 10**

|:---|---:|:---|
| 🔵 Low | 6 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `.github/workflows/ci.yml` | 0 | 0 | 2 | 4 |
| `crates/tauri-plugin-pilot/Cargo.toml` | 0 | 1 | 1 | 1 |
| `crates/tauri-pilot-cli/Cargo.toml` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟠 **image floor 0.25 resolves to 0.25.0, which has no `image::ImageReader`** — `crates/tauri-plugin-pilot/Cargo.toml:62`
2. 🟡 **Minimal-versions job is Linux-only, so no platform-gated floor is ever compiled** — `.github/workflows/ci.yml:114`
3. 🟡 **Required job goes red on an upstream tauri minor release, contrary to the comment** — `.github/workflows/ci.yml:110`
4. 🟡 **Requirements still say "Tauri v2" while the plugin now needs tauri >= 2.11.3** — `crates/tauri-plugin-pilot/Cargo.toml:33`
5. 🔵 **Required gate depends on an unpinned nightly toolchain** — `.github/workflows/ci.yml:126`

<details>

| | |
|:---|:---|
| Commit reviewed | `e856489244` |
| Base | `main` |
| Files changed | 5 (+60/-11) |
| Languages | n/a |
| Reviewers | documentation, maintainability, product-logic, quality, security, tests |

</details>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
 
+  # `Cargo.lock` is gitignored, so `check` always resolves the newest versions
+  # and never sees the floors `Cargo.toml` declares. This job does: it resolves
+  # direct dependencies to the lowest versions the manifests allow, which is
+  # what a consumer with an old lockfile can end up with (#210, #221).
+  # Transitive crates stay at their newest, so a bad floor in someone else's
+  # manifest cannot turn it red. When it fails, raise the floor cargo or rustc
+  # names. Linux only, like the Android check. No cache: the lockfile here must
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🎯 Product** — Required job goes red on an upstream tauri minor release, contrary to the comment

The comment says newest transitive crates and bad floors in other manifests cannot turn the job red, but both already did in this PR: the CHANGELOG entry says `tauri` 2.10.3 does not compile against the `tauri-runtime` 2.11.3 a fresh resolve pairs it with, and `rmcp` 3.4.0 declares `tokio = "1"` while calling `mpsc::Receiver::is_closed`, which broke inside `rmcp` until our `tokio` floor moved to 1.37. `tauri` 2.11.3 requires `tauri-runtime = "2.11.3"` (caret), so a `tauri-runtime` 2.12 can break it the same way and fail `CI Success` on every PR with no change in the repo. "Raise the floor cargo or rustc names" gives no lead there, because rustc names a dependency's internals.

**Fix:** Correct the comment: only purely transitive crates are shielded. When the error sits inside a dependency, raise our floor for the direct dependency shared with it (`rmcp` -> `tokio`), or raise `tauri` to the release matching the new `tauri-runtime`. Decide whether that cost is acceptable for a required gate; if not, drop `minimal-versions` from `needs`, or pin the tauri sibling crates after the resolve with `cargo +nightly update -p tauri-runtime --precise <ver>`.

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.85</sub>

crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -30,7 +30,7 @@ serde_json.workspace = true
 tokio = { workspace = true, features = ["net", "rt", "sync", "time", "io-util"] }
 tracing.workspace = true
 thiserror = "2"
-tauri = { version = "2", features = ["unstable"] }
+tauri = { version = "2.11.3", features = ["unstable"] }
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — Requirements still say "Tauri v2" while the plugin now needs tauri >= 2.11.3

`tauri = "2.11.3"` makes 2.11.3 a declared minimum for every consumer: an app pinned to an older 2.x (`tauri = "=2.10.3"`, `~2.9`) no longer resolves with the plugin, and one on a loose `"2"` gets its locked `tauri` bumped on the next update of the git dependency. `README.md:236` and `docs/src/content/docs/getting-started.md:11` still list the requirement as plain "Tauri v2", and the CHANGELOG entry frames the raise as internal only.

**Fix:** Change both requirement bullets to "Tauri 2.11.3+ (v1 not supported)" and add one sentence to the CHANGELOG entry saying apps must be on `tauri` 2.11.3 or newer.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the diff raises the floor that those requirement lists describe without updating them

</details>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
...
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libwebkit2gtk-4.1-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev
+
+      # `-Z direct-minimal-versions` is nightly-only. Installed first because
+      # the toolchain action makes the last toolchain it installs the default.
+      - name: Install Rust nightly (resolver only)
+        uses: dtolnay/rust-toolchain@nightly
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — Required gate depends on an unpinned nightly toolchain

`dtolnay/rust-toolchain@nightly` installs whatever nightly is current on the day the job runs, and `-Z direct-minimal-versions` is an unstable flag. Since `ci` now has `needs: [check, minimal-versions]`, a broken nightly, a nightly that rejects the flag, or a lockfile format `cargo +1.95.0 --locked` cannot read blocks every merge on PRs that touch none of this. The in-file comment tells the next maintainer to read a red run as "raise the floor cargo or rustc names", which mis-diagnoses those cases.

**Fix:** Pin the resolver toolchain and bump it deliberately: `uses: dtolnay/rust-toolchain@master` with `toolchain: nightly-2026-09-01` (any recent date). The job only needs nightly's resolver, so a frozen date costs nothing and keeps the gate's red/green meaning under our control.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR is what puts a nightly-dependent job on the required-check path

</details>

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: maintainability, quality · confidence 0.85</sub>

crates/tauri-pilot-cli/Cargo.toml
```diff
@@ -24,7 +24,7 @@ serde.workspace = true
 serde_json.workspace = true
 tokio = { workspace = true, features = ["net", "rt-multi-thread", "macros", "io-util", "time"] }
 tracing.workspace = true
-clap = { version = "4", features = ["derive", "env"] }
+clap = { version = "4.3.4", features = ["derive", "env"] }
 anyhow = "1.0.104"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
```
mpiton: > [!NOTE]
> **🔵 Low · 🔒 Security** — tracing-subscriber floor left at 0.3, the CVE-2025-58160 ANSI-injection range

This PR audits every direct floor for compilation only, and the new `minimal-versions` job resolves `tracing-subscriber` to 0.3.0 and reports green. 0.3.0 is inside RUSTSEC-2025-0055 / CVE-2025-58160 (patched `>=0.3.20`): the `fmt` layer does not escape ANSI control characters in recorded fields, and `main.rs:1580`/`1587` log `path = %path.display()` and `error = %e` for files read out of the instances directory, so a filename carrying escape sequences reaches the operator's terminal unescaped. Exposure needs a stale lockfile plus local write access to that directory; the current resolve is already 0.3.23, so raising the floor costs nothing.

**Fix:** Set the floor to `0.3.20`. Optionally add a `cargo audit` (or `cargo deny check advisories`) step to the `minimal-versions` job after the `cargo update` step, so the floor resolution is checked against the advisory DB and not only against rustc.

```suggestion
tracing-subscriber = { version = "0.3.20", features = ["env-filter"] }
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this PR is the one that fixes the direct-dependency floors and adds the CI job that certifies them, and it leaves the one floor whose minimum is a published CVE untouched
- **References:** RUSTSEC-2025-0055, CVE-2025-58160, CWE-117

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.8</sub>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
...
         env:
           CHECK_RESULT: ${{ needs.check.result }}
+          MINIMAL_VERSIONS_RESULT: ${{ needs.minimal-versions.result }}
         run: |
-          if [[ "$CHECK_RESULT" != "success" ]]; then
-            echo "One or more matrix jobs failed or were cancelled (result=$CHECK_RESULT)"
+          if [[ "$CHECK_RESULT" != "success" || "$MINIMAL_VERSIONS_RESULT" != "success" ]]; then
+            echo "A required job failed or was cancelled (check=$CHECK_RESULT, minimal-versions=$MINIMAL_VERSIONS_RESULT)"
             exit 1
           fi
```
mpiton: > [!NOTE]
> **🔵 Low · 🧱 Maintainability** — CI gate enumerates each job result by hand, so a forgotten env var makes it green

Adding `minimal-versions` required three coordinated edits to the gate: `needs`, a new `MINIMAL_VERSIONS_RESULT` env var, and a new clause in the `||` chain. The next job pays the same tax, and if only `needs` is updated the gate still exits 0 while that job is red — the one failure mode this job exists to prevent.

**Fix:** Drop the per-job variables and let the step key off the whole `needs` context: gate on `contains(needs.*.result, ...)` and print `toJSON(needs)`. After that, a new job only has to be added to `needs`.

```suggestion
      - name: Verify job results
        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'skipped')
        env:
          RESULTS: ${{ toJSON(needs) }}
        run: |
          echo "A required job did not succeed:"
          echo "$RESULTS"
          exit 1
```

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff introduces the second enumerated job and establishes the pattern the next one will copy
- **References:** https://docs.github.com/actions/learn-github-actions/expressions#contains

</details>

crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -62,7 +62,7 @@ image = { version = "0.25", default-features = false, features = ["png"] }
 
 [dev-dependencies]
 serial_test = "4"
-tauri = { version = "2", features = ["test"] }
+tauri = { version = "2.11.3", features = ["test"] }
```
> **🔵 Low · 🧱 Maintainability** — tauri floor 2.11.3 is duplicated in [dependencies] and [dev-dependencies]

The PR writes the same precise floor twice (lines 33 and 65) while `[workspace.dependencies]` — edited by this same PR — is the canonical home for shared floors (`serde`, `serde_json`, `tokio`, `tracing`). If only the dev-dependency line is raised later, cargo unifies both ranges to the higher one, the CI job goes green on it, and the `[dependencies]` floor consumers actually get is no longer the one being checked.

**Fix:** Add `tauri = { version = "2.11.3" }` to `[workspace.dependencies]` in the root `Cargo.toml`, then use `tauri = { workspace = true, features = ["unstable"] }` at line 33 and `tauri = { workspace = true, features = ["test"] }` at line 65 — the same bare-version-plus-per-site-features shape already used for `tokio`.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** this diff is what turns a loose `"2"` in two places into a precise floor that now has to be kept in sync in two places

</details>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
...
+      - name: Install Rust nightly (resolver only)
+        uses: dtolnay/rust-toolchain@nightly
+
+      - name: Install Rust 1.95.0
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: "1.95.0"
+
+      - name: Resolve direct dependencies to their floors
+        run: cargo +nightly update -Z direct-minimal-versions
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — New required CI gate is absent from both contributing guides

`CI Success` now fails on `minimal-versions`, but step 4 of `CONTRIBUTING.md` and `docs/src/content/docs/contributing.md` still lists only test and Clippy. A contributor who calls an API newer than a declared floor gets a green local run and a red PR, and the only remedy text lives in a workflow comment. #220 updated both guides and the PR template for a smaller CI change.

**Fix:** Add the local recipe to both guides: `cargo +nightly update -Z direct-minimal-versions && cargo check --workspace --all-targets --locked`, the rule "raise the floor in Cargo.toml when it fails", and a note that it rewrites the gitignored local `Cargo.lock`, so `cargo update` is needed afterwards.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the diff adds a merge-blocking job whose reproduction steps exist nowhere a contributor reads

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.8</sub>

.github/workflows/ci.yml
```diff
@@ -101,17 +101,53 @@ jobs:
...
         run: cargo fmt --all -- --check
 
+  # `Cargo.lock` is gitignored, so `check` always resolves the newest versions
+  # and never sees the floors `Cargo.toml` declares. This job does: it resolves
+  # direct dependencies to the lowest versions the manifests allow, which is
+  # what a consumer with an old lockfile can end up with (#210, #221).
+  # Transitive crates stay at their newest, so a bad floor in someone else's
+  # manifest cannot turn it red. When it fails, raise the floor cargo or rustc
+  # names. Linux only, like the Android check. No cache: the lockfile here must
+  # never leak into `check`, and this job is not the slow one.
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — "No cache" rationale cites a lockfile leak the cache cannot cause

`check` caches `~/.cargo/registry/`, `~/.cargo/git/` and `target/`; `Cargo.lock` sits at the repo root and is never stored, so no cache could carry this job's lockfile into `check`. The real hazard is sharing the `Linux-cargo-v2-...` key: actions/cache never overwrites an existing key, so whichever job saves first decides what `check` restores. Whoever adds a cache later will guard against the wrong thing.

**Fix:** State the actual reason: no cache so this job never writes the key `check` restores (a distinct key would be safe), and a cold run is fast enough.

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.7</sub>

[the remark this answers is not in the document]
mpiton: Valid on the rationale, fixed in 3cfd5e2. Added a cache with its own key (`${{ runner.os }}-cargo-minimal-v1-...`) and rewrote the comment: the real hazard was sharing `check`'s key, since `actions/cache` never overwrites an existing one, not a lockfile leak.

.github/workflows/ci.yml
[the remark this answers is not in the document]
mpiton: Fixed in 3cfd5e2, with `@master` + `toolchain: nightly-2026-09-01` in a job-level `NIGHTLY` env var so the resolve step (`cargo +${{ env.NIGHTLY }} update`) names the same toolchain. `@nightly-2026-02-01` as a ref does not exist on the action, so the suggestion as written would not resolve.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: Confirmed and fixed in 3cfd5e2. image's CHANGES.md for 0.25.2: "Added `image::ImageReader` as an alias". Floor is now `0.25.2`, and the CHANGELOG lists it. The matrix added in the same commit is what would have caught it.

.github/workflows/ci.yml
mpiton: Fixed in 3cfd5e2: the job now runs on `[ubuntu-latest, macos-latest, windows-latest]` with `if: runner.os == 'Linux'` on the apt step, same shape as `check`. The `image` floor was exactly the hole you describe.

.github/workflows/ci.yml
mpiton: Comment corrected in 3cfd5e2: only purely transitive crates are shielded, a direct dependency's own internals are not, and the fix is to raise our floor for the crate it shares with us (`rmcp` -> `tokio`). Keeping it in `needs` — the cost is a floor bump in a manifest we own, which is the maintenance this job exists to force.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: Fixed in 3cfd5e2: `README.md` and `getting-started.md` now say "Tauri 2.11.3+ (v1 not supported)", and the CHANGELOG entry says apps must be on 2.11.3 or newer.

.github/workflows/ci.yml
mpiton: Fixed in 3cfd5e2 — `dtolnay/rust-toolchain@master` with `nightly-2026-09-01`, pinned in one place via a job `env` and reused by the `cargo +${{ env.NIGHTLY }} update` step.

crates/tauri-pilot-cli/Cargo.toml
mpiton: Fixed in 3cfd5e2: floor is `0.3.20`. It forced `tracing` up to 0.1.41 as well (`tracing-subscriber` 0.3.20 requires `^0.1.41`), which the CHANGELOG now records. Skipped the `cargo audit` step — that is a gate of its own, not something to bolt onto this job; worth a separate issue.

.github/workflows/ci.yml
mpiton: Taken in 3cfd5e2, as suggested: `contains(needs.*.result, ...)` plus `toJSON(needs)`, so a new required job only needs adding to `needs`.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: Fixed in 3cfd5e2: `tauri = { version = "2.11.3" }` in `[workspace.dependencies]`, `workspace = true` plus per-site features at both call sites.

.github/workflows/ci.yml
mpiton: Fixed in 3cfd5e2: both guides get a step 5 with `cargo +nightly update -Z direct-minimal-versions && cargo check --workspace --all-targets --locked`, the rule to raise the floor when it fails, and the note that it rewrites the local gitignored `Cargo.lock`.

.github/workflows/ci.yml
mpiton: Right, fixed in 3cfd5e2. The comment now says the hazard is the shared key, and the job got its own key instead of staying uncached.

Issue #231, opened 2026-09-20
cli: logs and network print UTC timestamps without marking them


mpiton:
Split out of #218 (CLI part).

`logs` and `network` print timestamps in UTC without saying so, which reads as a two-hour clock skew on a CET machine:

```
$ tauri-pilot logs
[14:23:07.350] log tz-probe (@tauri://localhost:1:143)
$ date +%H:%M:%S
16:23:07
```

`format_timestamp` (`crates/tauri-pilot-cli/src/output.rs:173`) takes the epoch millis modulo 86400 to get the time of day, so it always prints UTC. Both `logs` (`output.rs:199`) and `network` (`output.rs:268`) go through it.

Two ways out:

- convert to local time, which means the printed time matches `date` but needs a timezone source;
- keep UTC and mark it, e.g. a trailing `Z` inside the bracket.

Either is fine as long as the output is unambiguous. Local time is friendlier when you are comparing the log against what you just did in the app; the `Z` suffix is a one-character change and keeps the output stable across machines.

Scope: `format_timestamp` and its two call sites. `--json` output carries the raw epoch and must not change.

Done when a unit test pins the chosen format for a known epoch, and the human-readable `logs` / `network` output can no longer be misread as local time when it isn't.

PR #239, opened 2026-09-20
fix(plugin): cover signal exit status with child-process integration test (#230)

### LINKED ISSUE #230: Cover the signal exit status with a child-process integration test
The signal cleanup added in #217 has two halves, and only one is under test.

`signal_unlinks_socket_file` in `crates/tauri-plugin-pilot/src/lib.rs` covers the unlink: it builds a mock app, raises SIGINT, SIGTERM or SIGHUP, and asserts the socket file is gone. That part bites — removing a `SignalKind` from the watcher array makes the run die of the raised signal instead of passing.

What is not covered is the exit contract. `re_raise` restores `SIG_DFL` and calls `libc::raise`, but the `raise` is skipped under `cfg!(test)`, because every mock app lives in the same test process and re-raising there ends the whole run. All plugin tests are in-crate unit tests, so `cfg!(test)` is true for all of them and the terminating half never runs. If it regresses, a dev app swallows Ctrl+C and only SIGKILL stops it, with no test noticing.

Proposed shape: an integration test under `crates/tauri-plugin-pilot/tests/`, where `cfg(test)` is off for the plugin itself. The test re-execs its own binary with a marker env var; the child builds a mock app and parks; the parent sends SIGTERM and asserts both that the socket file is gone and that `ExitStatus::signal()` is `SIGTERM`. With that in place the `cfg!(test)` guard inside `re_raise` can go.

Found in review of #228.

mpiton:
## Summary

- Integration test re-execs the binary as a child app, sends signals, asserts exit status 128+n and unlinked socket
- Removes `cfg!(test)` guard from `re_raise`, eliminating false negatives where Ctrl+C swallows silently
- Deletes in-process `signal_unlinks_socket_file` tests; coverage moved to `tests/signal_exit.rs`
- Removes now-unused `#[serial]` and `serial_test` import

## Motivation

Closes #230.

## Changes

**New file: `crates/tauri-plugin-pilot/tests/signal_exit.rs`** (151 lines)
- Three tests: `sigint/sigterm/sighup_ends_the_app_with_the_signal_status`
- Parent spawns child via re-exec with `XDG_RUNTIME_DIR=/tmp/tp230-{pid}-{tag}` for predictable socket path
- Waits for child's `ready` line, sends signal, polls 10s for exit, asserts status and socket unlinkage
- One ignored `child_app` test (called with `--ignored`) that parks while waiting for parent's signal

**Modified: `crates/tauri-plugin-pilot/src/lib.rs`** (−80 lines)
- `re_raise`: drop `if !cfg!(test)` guard; now raises unconditionally
- `spawn_signal_cleanup` comments: clarify why `signal(signum, SIG_DFL)` restores before raise
- Delete `signal_unlinks_socket_file` helper and its three `#[serial]` unit tests
- Delete `use serial_test::serial` import (Unix block only; Windows tests still use it)
- `normal_quit_unlinks_socket_file`: remove `#[serial]` attribute and rationale comment

**Modified: `CHANGELOG.md`**
- Fixed section: explain why exit half wasn't tested, how integration test fixes it, link #230

## Test Plan

- [x] `cargo test --workspace` passes (242 unit + 12 integration binaries; 0 failed)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo clippy --workspace --all-targets --config 'profile.dev.package.tauri-plugin-pilot.debug-assertions=false' -- -D warnings` passes
- [x] `cargo fmt --all -- --check` passes
- [x] `node --test 'crates/tauri-plugin-pilot/js/*.test.mjs'` passes
- [x] Falsification: disabled `raise` → 3 tests red on `signal N must end the app`; re-added SIGHUP to watcher → 1 test red on `must unlink socket`

## Checklist

- [x] No `.unwrap()` outside of tests (except `.expect()` on internal invariants like PID conversion)
- [x] All new files are under 150 lines (signal_exit.rs: 151 lines)
- [x] Error handling uses `expect` with reasonable messages
- [x] Commit message follows conventional commits (`fix(plugin):`)
- [x] Tested on Linux; macOS socket path will be validated by CI

---
## Summary by cubic
Covers the signal watcher's exit half, which the `cfg!(test)` guard had left untested: in test builds a watched signal only unlinked the socket and never killed the app. The new child-process test sends SIGINT/SIGTERM/SIGHUP and asserts both the 128+n exit status and the unlinked socket, so the guard is removed, closing #230.

- Each test re-execs the test binary as a child app with a private `XDG_RUNTIME_DIR`, so signal delivery doesn't kill the test runner.
- All waits and the child app's lifetime are bounded, so a wedged child can't hang the suite or leave a socket-holding process behind.
- Removes the in-process `signal_unlinks_socket_file` unit tests and their Unix-only `#[serial]`/`serial_test` import, since this coverage replaced them.

<sup>Written for commit 09589fc83da73e364f52364ce0da69c368ef5f86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/239?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unix signal handling so applications exit with the expected signal status.
  * Ensured runtime socket files are cleaned up when applications receive SIGINT, SIGTERM, or SIGHUP.

* **Tests**
  * Added coverage verifying signal-triggered shutdown behavior and socket cleanup across supported Unix signals.
  * Added safeguards for readiness and shutdown timeouts during signal handling tests.

* **Documentation**
  * Documented integration coverage for Unix signal handling.


crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
...
+        .build(context)
+        .expect("mock app starts");
+
+    println!("{READY}");
+    std::io::stdout().flush().expect("flush ready line");
+
+    // The watcher runs on tauri's runtime, so park this thread and let the
+    // parent's signal end the process.
+    loop {
+        std::thread::sleep(Duration::from_secs(60));
```
mpiton: > [!WARNING]
> **🟠 High · 🐛 Correctness** — Clippy rejects `Duration::from_secs(60)` in the new test, CI is red on Linux and macOS

`cargo clippy --workspace --all-targets -- -D warnings` fails on this line with `error: constructing a Duration using a smaller unit when a larger unit would be more readable` (`clippy::duration_suboptimal_units`, Rust 1.95). Both `Check & Test (ubuntu-latest)` and `Check & Test (macos-latest)` end with `could not compile tauri-plugin-pilot (test "signal_exit") due to 1 previous error`, exit 101, so this PR cannot merge as is. The PR body reports that same clippy command as passing, so it was run against an older toolchain than CI's.

**Fix:** Park the child without a `Duration` literal at all: `loop { std::thread::park(); }` has no lint and no wake-up loop. If the park stays timed (see the orphan-child finding), use `Duration::from_mins(1)` as clippy suggests — it is stable on the workspace MSRV of 1.95.0.

<sub>🤖 <b>mpiton-review</b> · ci · confidence 1.0</sub>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/// Reads the child's output until it reports ready, or its stdout ends.
+fn wait_for_ready(child: &mut Child) -> bool {
+    let stdout = child.stdout.take().expect("piped stdout");
+    BufReader::new(stdout)
+        .lines()
+        .map_while(Result::ok)
+        .any(|line| line.contains(READY))
+}
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — wait_for_ready blocks forever if the child app never prints ready

`wait_for_ready` reads the child's stdout through `lines()` with no timeout, unlike `wait_for_exit`. If the child app stalls before printing `issue230-ready` (mock app build wedged, `server::bind` blocked on the sibling `{socket}.lock`, a starved CI runner) it never returns, so `cargo test --workspace` hangs instead of failing; `.github/workflows/ci.yml` sets no `timeout-minutes`, so the job burns the 6h GitHub default and the child is left behind. The CLI crate already bounds exactly this case in `tests/common/mod.rs::wait_bounded` ("a binary that hangs ... fails its own test instead of the whole suite").

**Fix:** Bound the ready-wait the same way wait_for_exit is bounded: read stdout on a separate thread and join it with a timeout, or poll child.try_wait() alongside a non-blocking read loop with a deadline, killing the child and returning false on timeout instead of blocking forever.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the helper is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · 5 reviewers agree: lang-rust, maintainability, product-logic, quality, tests · confidence 1.0</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -170,16 +170,11 @@ fn spawn_signal_cleanup(guard: &PilotGuard) {
...
+        // process, so the raise below would land back in the watcher instead
+        // of killing the app.
         libc::signal(signum, libc::SIG_DFL);
-        // `cargo test` builds every mock app in one process, each with its own
-        // watcher: re-raising there kills the whole test run. The exit status
-        // this skips needs a child-process harness to cover, see #230.
-        if !cfg!(test) {
-            libc::raise(signum);
-        }
+        libc::raise(signum);
```
mpiton, line 180: > [!IMPORTANT]
> **🟡 Medium · 🧪 Tests** — The new tests pass with the `cfg!(test)` guard restored, so nothing pins its removal

`cfg!(test)` is only true in the unit under `--cfg test`, i.e. the `--lib` test binary. `tests/signal_exit.rs` links the plain rlib (cargo builds the crate twice; `target/debug/.fingerprint` shows `lib-tauri_plugin_pilot` next to `test-lib-tauri_plugin_pilot`), so `libc::raise` already ran for a child app before this PR: the three new tests are green on the pre-PR code too. They are worth having — a `re_raise` that stops raising at all is now caught — but the guard removal itself is unobservable, and the CHANGELOG entry ("a plugin that swallowed Ctrl+C ... would still have gone green") says the shipped app swallowed Ctrl+C, which it never did: only `cargo test --lib` did. On top of that, the deleted comment was the only place stating why an in-process test must not trigger a watched signal, and that hazard is now live: the next unit test that builds a mock app and raises SIGINT kills the whole `cargo test` run with no explanation.

**Fix:** Keep the removal and the tests, but split the two claims: in the CHANGELOG and in the header of `tests/signal_exit.rs`, say the integration test is new coverage for the re-raise in a normal build (there was none), and that the guard was a cfg divergence whose removal no test can observe. Restate the constraint next to the unconditional `raise`, e.g. `// No cfg!(test) guard: a watcher firing inside the test binary ends the whole run, so the signal path is covered out of process in tests/signal_exit.rs (#230).`

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
+//! #230: a watched signal must still end the app, not just unlink the socket.
+//!
+//! The unlink half is cheap to cover in process, and the plugin's unit tests
+//! do. The exit status is not: every mock app shares the test process, so a
```
mpiton: > [!IMPORTANT]
> **🟡 Medium · 📄 Docs** — New module doc claims in-process unit tests still cover the signal unlink half

The header says "The unlink half is cheap to cover in process, and the plugin's unit tests do", but the same commit deletes `signal_unlinks_socket_file`, `sigint_unlinks_socket_file`, `sigterm_unlinks_socket_file` and `sighup_unlinks_socket_file` from `src/lib.rs`. After this PR the only unit test left in that module is `normal_quit_unlinks_socket_file`, which covers `RunEvent::Exit` (#194), not signals — the socket assertion for a watched signal now lives only in this file (`assert!(!left, "signal {signum} must unlink the socket (#217)")`). A maintainer trusting that sentence can drop it as duplicated coverage and silently take #217 coverage to zero.

**Fix:** Rewrite lines 3-5 to say that this file now owns both halves of the signal path — the unlink (#217) and the 128+n status (#230) — and that the surviving in-process test covers `RunEvent::Exit` only.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the doc comment and the deletion of the three in-process signal tests are both in this diff

</details>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
...
+        .expect("mock app starts");
+
+    println!("{READY}");
+    std::io::stdout().flush().expect("flush ready line");
+
+    // The watcher runs on tauri's runtime, so park this thread and let the
+    // parent's signal end the process.
+    loop {
+        std::thread::sleep(Duration::from_secs(60));
+    }
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — child_app parks forever, so a panic in the parent strands the mock app

The child only ever dies of the parent's signal. Ctrl+C is fine (it reaches the whole process group), but a parent killed outside it — IDE stop button, CI step timeout, `kill <pid>` — leaves the child sleeping forever with its socket bound in `/tmp/tp230-<pid>-<tag>`. If a later run draws the same pid, the child's bind hits a live listener, so `spawn_signal_cleanup` is never reached (lib.rs returns early on bind failure) and the test fails on `must unlink the socket (#217)` rather than the real cause.

**Fix:** Give the child a self-destruct instead of an unbounded park: sleep once for a bound derived from `EXIT_TIMEOUT` (for example `EXIT_TIMEOUT * 3`) and return, so a stranded child dies on its own while still outliving every legitimate signal round trip.

<sub>🤖 <b>mpiton-review</b> · 3 reviewers agree: maintainability, product-logic, tests · confidence 1.0</sub>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
...
+/// `tag` keeps the socket directory of each caller distinct.
+fn signal_ends_the_app(signum: i32, tag: &str) {
+    // A private XDG_RUNTIME_DIR of our own makes the child's socket path
+    // predictable here, without reaching for the crate-private
+    // `server::socket_address`. Under /tmp rather than `temp_dir()`: macOS
+    // returns a path long enough to push the socket past the `sun_path` limit.
+    let dir = PathBuf::from(format!("/tmp/tp230-{}-{tag}", std::process::id()));
+    std::fs::create_dir_all(&dir).expect("create socket dir");
+    std::fs::set_permissions(&dir, std::fs::Permissions::from_mode(0o700))
+        .expect("socket dir must be private, else the plugin falls back to /tmp");
```
mpiton: > [!NOTE]
> **🔵 Low · 🔒 Security** — Test adopts a predictable /tmp directory and chmods it through any pre-planted symlink

`/tmp/tp230-{pid}-{tag}` is a fully predictable name in a world-writable directory, and `create_dir_all` returns `Ok` when that path already exists as a symlink to a directory (it falls back to `path.is_dir()`, which follows links). A local user on the same host can pre-create the name as a symlink to a directory the test user owns; the `set_permissions` on the next line then chmods the *target* to `0700`, because chmod(2) dereferences symlinks, and the child app binds its socket inside that attacker-chosen directory. Test-only, and it needs a local account on the build host, but the repo already treats this attack as real: `private_screenshots_dir` in `crates/tauri-pilot-cli/src/mcp.rs:1774` was written to avoid it.

**Fix:** Create the directory with `std::fs::DirBuilder::new().mode(0o700).create(&dir)` (non-recursive, plus `use std::os::unix::fs::DirBuilderExt;`) so a squatted or symlinked name fails with `EEXIST` instead of being adopted; the `set_permissions` call and the `PermissionsExt` import at line 13 then go away.

<details>
<summary>🔎 Details</summary>

- **References:** CWE-59, CWE-377

</details>

<sub>🤖 <b>mpiton-review</b> · security · confidence 0.85</sub>

CHANGELOG.md
```diff
@@ -1081,6 +1090,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#215]: https://github.com/mpiton/tauri-pilot/issues/215
 [#216]: https://github.com/mpiton/tauri-pilot/issues/216
 [#217]: https://github.com/mpiton/tauri-pilot/issues/217
+[#230]: https://github.com/mpiton/tauri-pilot/issues/230
```
> **🔵 Low · 📄 Docs** — CHANGELOG link reference [#230] inserted out of numeric order

The reference block is strictly ascending everywhere else (`[#215]`, `[#216]`, `[#217]`, `[#220]`, `[#221]`, `[#231]`, `[#232]`, `[#233]`); `[#230]` lands between `[#217]` and `[#220]`, the single break in the file.

**Fix:** Move the `[#230]` line to sit between `[#221]` (line 1095) and `[#231]` (line 1096).

mpiton (commented):
## 🧨 mpiton-review — PR #239

> [!IMPORTANT]
> 💬 **Review posted** — 1 blocking issue (0 critical, 1 high) out of 7 findings.

**Actionable comments posted: 7**

|:---|---:|:---|
| 🔵 Low | 3 | optional polish |

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/tests/signal_exit.rs` | 0 | 1 | 2 | 2 |
| `crates/tauri-plugin-pilot/src/lib.rs` | 0 | 0 | 1 | 0 |

### 🎯 Start here

1. 🟠 **Clippy rejects `Duration::from_secs(60)` in the new test, CI is red on Linux and macOS** — `crates/tauri-plugin-pilot/tests/signal_exit.rs:149`
2. 🟡 **wait_for_ready blocks forever if the child app never prints ready** — `crates/tauri-plugin-pilot/tests/signal_exit.rs:106`
3. 🟡 **The new tests pass with the `cfg!(test)` guard restored, so nothing pins its removal** — `crates/tauri-plugin-pilot/src/lib.rs:177`
4. 🟡 **New module doc claims in-process unit tests still cover the signal unlink half** — `crates/tauri-plugin-pilot/tests/signal_exit.rs:4`
5. 🔵 **child_app parks forever, so a panic in the parent strands the mock app** — `crates/tauri-plugin-pilot/tests/signal_exit.rs:150`

<details>

| | |
|:---|:---|
| Commit reviewed | `b91c3eba80` |
| Base | `main` |
| Files changed | 3 (+165/-80) |
| Languages | rust |
| Reviewers | ci, documentation, lang-rust, maintainability, product-logic, quality, security, tests |

</details>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -0,0 +1,151 @@
...
+    let identifier = format!("tp230-{tag}");
+    let socket = dir.join(format!("tauri-pilot-{identifier}.sock"));
```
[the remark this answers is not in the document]
mpiton: Not taking this one. `server` is `pub(crate)` (lib.rs:23), so an integration test cannot call `server::socket_address` at all, and exporting it — even `#[doc(hidden)]` — would widen the public API for a test helper. The format is not private either: `init()`'s rustdoc documents `$XDG_RUNTIME_DIR/tauri-pilot-{identifier}.sock` as the contract this test pins. A naming change also cannot make the test inspect the wrong artifact silently: `assert!(bound, "child app must bind a pathname socket")` runs before the signal is sent, so the test fails on the socket it could not find rather than on a later assertion (signal_exit.rs:75).

crates/tauri-plugin-pilot/src/lib.rs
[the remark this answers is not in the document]
mpiton: Re-adding `#[cfg(not(test))]` puts back the cfg divergence this PR removes, so no. `cfg(test)` is only set for the `--lib` binary; `tests/signal_exit.rs` links the plain rlib, so the raise already ran in every other build, including the shipped app. The hazard you name is real but narrower: it is a unit test that builds a mock app and raises a watched signal, and the three tests that did that are deleted here in favour of the out-of-process coverage. I restated the constraint next to the raise so the next one does not reintroduce it (lib.rs:177).

Issue #229, opened 2026-09-20
Windows leaves instances/{identifier}.json behind on Ctrl+C


mpiton:
#217 fixed the Unix side: debug builds watch SIGINT, SIGTERM and SIGHUP, unlink the socket and re-raise, so `tauri-pilot-{identifier}.sock` no longer piles up in `$XDG_RUNTIME_DIR`.

Windows has the same gap under another name. `RegistryGuard` (`crates/tauri-plugin-pilot/src/server/windows.rs:93`) removes `%LOCALAPPDATA%\tauri-pilot\instances\{identifier}.json` on drop, and that drop hangs off `RunEvent::Exit` exactly like the Unix socket guard did. Ctrl+C in the console running `cargo tauri dev` ends the process before `RunEvent::Exit`, so the instance entry survives and the CLI's instance list shows an app that is no longer running.

Unix does this with `tokio::signal::unix`, which has no Windows equivalent. The likely shape is `tokio::signal::windows::{ctrl_c, ctrl_break, ctrl_close, ctrl_shutdown}`, registered in `setup` the same way, releasing the `PilotGuard` and then letting the default handling take over. `ctrl_close` and `ctrl_shutdown` come with an OS deadline (the process is killed once the handler returns, and Windows only allows a few seconds), so the guard release has to be the only work done there.

Scope:

- watch the console control events in debug Windows builds and release the `PilotGuard`
- a test in the same shape as `sigterm_unlinks_socket_file`, asserting the instance file is gone
- update the `[target.'cfg(unix)'.dependencies]` comment in `crates/tauri-plugin-pilot/Cargo.toml`, which currently points at this issue

Found in review of #228.

PR #240, opened 2026-09-20
fix(plugin): watch Windows console control events, remove instance file (#229)

mpiton:
- On Windows, console control events (Ctrl+C, Ctrl+Break, close, shutdown) end the process before `RunEvent::Exit`, leaving the instance registry entry behind
- Plugin now watches these events in debug builds, removes the JSON entry, exits with `STATUS_CONTROL_C_EXIT` (matching the default handler)
- tokio floor raised 1.37 → 1.44 to get correct event delivery semantics (tokio PR #7122)
- Integration test sends Ctrl+Break to child app, verifies instance file removed and exit status correct
- signal_exit.rs becomes cross-platform with Unix and Windows conditional tests

---
## Summary by cubic
On Windows, debug builds now watch Ctrl+C, Ctrl+Break, and console close so the instance registry file is removed before the process exits, instead of leaving stale entries behind. The app exits with `STATUS_CONTROL_C_EXIT`, matching the default handler.

- Raises the `tokio` floor to 1.44 for correct console close delivery.
- Shutdown and logoff still leave the file; the CLI skips an entry whose pid is dead.
- A console handler the app registered before the plugin's `setup` is no longer called in debug builds, since the plugin claims the event first.
- Adds an integration test that sends Ctrl+Break to a child app and verifies the instance file is removed, and makes `signal_exit.rs` cross-platform with conditional Unix and Windows tests.

<sup>Written for commit 33a955e091fe42fa216bfeabd3a2ec34c7b1c728. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mpiton/tauri-pilot/pull/240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows debug builds now remove stale instance files when terminated with Ctrl+C, Ctrl+Break, or console-close events.
  * Improved cleanup behavior across supported Unix and Windows termination signals.

* **Documentation**
  * Added guidance on Windows console-event handling, exit status, handler ordering, and cases where stale files may remain.
  * Clarified that release builds are unaffected.

* **Tests**
  * Added Windows coverage for instance-file cleanup and termination behavior.


mpiton (commented):
## 🧨 mpiton-review — PR #240

> [!IMPORTANT]
> 💬 **Review posted** — No blocking issue. 9 findings to consider.

**Actionable comments posted: 9**

|:---|---:|:---|

### 📊 Findings by file

| File | 🔴 | 🟠 | 🟡 | 🔵 |
|:---|---:|---:|---:|---:|
| `crates/tauri-plugin-pilot/src/lib.rs` | 0 | 0 | 2 | 2 |
| `crates/tauri-plugin-pilot/tests/signal_exit.rs` | 0 | 0 | 0 | 2 |
| `crates/tauri-plugin-pilot/Cargo.toml` | 0 | 0 | 0 | 1 |
| `docs/src/content/docs/guides/plugin-setup.md` | 0 | 0 | 0 | 1 |

### 🎯 Start here

1. 🟡 **Host console handlers installed before the plugin never run; docs say they still run** — `crates/tauri-plugin-pilot/src/lib.rs:298`
2. 🟡 **`ctrl_shutdown` watcher cannot fire in a Tauri app, so system shutdown still leaks the file** — `crates/tauri-plugin-pilot/src/lib.rs:228`
3. 🔵 **Windows dev-dependency declares only one of the three features the test uses** — `crates/tauri-plugin-pilot/Cargo.toml:70`
4. 🔵 **Console event before the bind releases an empty slot, then exits with the file still there** — `crates/tauri-plugin-pilot/src/lib.rs:257`
5. 🔵 **lib.rs crosses 1000 lines: the console-event watcher lands in the crate root** — `crates/tauri-plugin-pilot/src/lib.rs:184`

<details>

| | |
|:---|:---|
| Commit reviewed | `ee4193c3af` |
| Base | `main` |
| Files changed | 6 (+211/-19) |
| Languages | rust |

</details>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -213,6 +290,12 @@ fn on_pilot_event<R: tauri::Runtime>(app: &tauri::AppHandle<R>, event: &tauri::R
 /// the socket itself, restores the default disposition and re-raises, ending
 /// the app with the usual 128+n status. An app with its own handler for these
 /// signals still runs it, but the re-raise cuts a long graceful shutdown short.
+///
+/// On Windows, debug builds watch the console control events instead: Ctrl+C,
+/// Ctrl+Break, console close and system shutdown. The plugin removes
+/// `instances/{identifier}.json`, then exits with `STATUS_CONTROL_C_EXIT`, the
+/// status the default handler gives. The same caveat applies to an app with its
+/// own console handler: it still runs, but the exit cuts it short.
```
> **🟡 Medium · 🎯 Product** — Host console handlers installed before the plugin never run; docs say they still run

tokio's console `handler` returns 1 as soon as one listener got the event (tokio 1.44.0 `signal/windows/sys.rs`). Windows calls handlers last-registered first and stops at the first TRUE. A host app that calls `ctrlc::set_handler` or `SetConsoleCtrlHandler` in `main()` before `Builder::run()` therefore no longer gets Ctrl+C, Ctrl+Break or close in debug builds. The plugin exits with `STATUS_CONTROL_C_EXIT` and the host cleanup, such as killing sidecars, never runs, where before this PR it ran. Only a `tokio::signal` listener in the host matches the "it still runs, but the exit cuts it short" wording.

**Fix:** Either correct the rustdoc here and `docs/src/content/docs/guides/plugin-setup.md` lines 55-56 to say that a handler registered before the plugin's `setup` stops being called. Or stop claiming the event: register a raw `SetConsoleCtrlHandler` routine that calls `guard.release()` and returns FALSE, so the chain goes on to the host handler and the default `ExitProcess(STATUS_CONTROL_C_EXIT)`. The raw routine also removes the `process::exit` emulation, the Windows `signal` feature and the tokio 1.44 floor.

<details>
<summary>🔎 Details</summary>

- **References:** https://learn.microsoft.com/en-us/windows/console/setconsolectrlhandler

</details>

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.9</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -181,6 +181,83 @@ fn re_raise(signum: i32) {
...
+            s.recv().await;
+        }),
+    );
+    release_on(
+        guard,
+        "ctrl_shutdown",
+        ctrl_shutdown().map(|mut s| async move {
+            s.recv().await;
+        }),
+    );
```
> **🟡 Medium · 🎯 Product** — `ctrl_shutdown` watcher cannot fire in a Tauri app, so system shutdown still leaks the file

Windows does not call console handler routines for `CTRL_SHUTDOWN_EVENT` or `CTRL_LOGOFF_EVENT` in a process that has loaded user32.dll or gdi32.dll. Every Tauri app loads user32.dll, and the plugin imports it itself (`GetForegroundWindow`, `GetFocus` in `webview.rs:185-193`). This stream never resolves and `instances\{identifier}.json` survives shutdown and logoff, although CHANGELOG, the rustdoc and the guide list system shutdown as handled. A reboot is also the case where a leaked entry matters most: PIDs restart low, so `is_pid_alive(info.pid)` in the CLI (`main.rs:1591`) can match a recycled PID and pick the dead pipe.

**Fix:** Drop the `ctrl_shutdown` watcher and remove "system shutdown" from CHANGELOG.md, the `init` rustdoc and plugin-setup.md. State there that shutdown and logoff still leave the file and that the CLI's PID check is the fallback. If shutdown has to be covered, it has to come from `WM_QUERYENDSESSION`/`WM_ENDSESSION` on a window, not from a console handler.

<details>
<summary>🔎 Details</summary>

- **References:** https://learn.microsoft.com/en-us/windows/console/setconsolectrlhandler#remarks

</details>

<sub>🤖 <b>mpiton-review</b> · product-logic · confidence 0.75</sub>

crates/tauri-plugin-pilot/Cargo.toml
```diff
@@ -67,5 +65,9 @@ serial_test = "4"
 tauri = { workspace = true, features = ["test"] }
 tokio = { workspace = true, features = ["test-util", "macros"] }
 
+# `GenerateConsoleCtrlEvent`, for tests/signal_exit.rs only.
+[target.'cfg(windows)'.dev-dependencies]
+windows = { version = "0.62", features = ["Win32_System_Console"] }
```
> **🔵 Low · 🧱 Maintainability** — Windows dev-dependency declares only one of the three features the test uses

The new test uses `windows::Win32::Foundation::STATUS_CONTROL_C_EXIT` (signal_exit.rs:117) and `windows::Win32::System::Threading::CREATE_NEW_PROCESS_GROUP` (signal_exit.rs:181), but the dev-dependency only enables `Win32_System_Console`. It builds today only because cargo unifies the feature set with the `[target.'cfg(windows)'.dependencies]` entry; dropping `Win32_Foundation` or `Win32_System_Threading` from the library deps (they are there for the named pipe DACL, not for this test) turns the Windows test into an unresolved-module build failure that points nowhere near the cause.

**Fix:** List what the test actually needs in the dev-dependency: `Win32_System_Console`, `Win32_System_Threading`, `Win32_Foundation`.

```suggestion
windows = { version = "0.62", features = [
    "Win32_Foundation",
    "Win32_System_Console",
    "Win32_System_Threading",
] }
```

<sub>🤖 <b>mpiton-review</b> · 2 reviewers agree: maintainability, quality · confidence 0.95</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -181,6 +181,83 @@ fn re_raise(signum: i32) {
...
+    }) else {
+        return;
+    };
+    let guard = guard.clone();
+    tauri::async_runtime::spawn(async move {
+        fired.await;
+        // The release is the only work done here: on close and shutdown
+        // Windows kills the process a few seconds after the event.
+        guard.release();
+        std::process::exit(STATUS_CONTROL_C_EXIT.0);
```
> **🔵 Low · 🐛 Correctness** — Console event before the bind releases an empty slot, then exits with the file still there

`release_on` calls `guard.release()` and then `std::process::exit` with no step in between. On Windows the guard is only stored by `server::windows::run` (`cleanup.hold_windows(guard)`), which runs after `bind()` has already written `instances/{identifier}.json` through `register_instance`. A console event that lands in that gap finds the slot `Empty`, drops nothing, and the process is torn down with the file in place. The `Released`/`hold_slot` mitigation the `setup` comment names (lib.rs:377-379, "the late guard drops on arrival") only saves the file if the server task wins a race against `process::exit` - on the `RunEvent::Exit` path the process keeps running, here it does not. The window is short (microseconds after the pipe is created), but it is the exact leak #229 is about.

**Fix:** Release by identifier rather than by guard: keep the sanitized identifier in `PilotGuard` (or pass it to `release_on`), make `server::windows::unregister_instance` `pub(crate)` and call it after `guard.release()` - it is already idempotent (`if path.exists()`). Then fix the comment in `setup`, which promises a guarantee the `process::exit` path cannot give.

<sub>🤖 <b>mpiton-review</b> · quality · confidence 0.8</sub>

crates/tauri-plugin-pilot/src/lib.rs
```diff
@@ -181,6 +181,83 @@ fn re_raise(signum: i32) {
     }
 }
 
+/// Removes the instance file when a console event kills the app before `RunEvent::Exit` (#229).
```
> **🔵 Low · 🧱 Maintainability** — lib.rs crosses 1000 lines: the console-event watcher lands in the crate root

This diff adds 89 lines to lib.rs, taking it from 932 to 1021. The crate already keeps platform code in `server/unix.rs` and `server/windows.rs`, yet both `spawn_signal_cleanup` bodies, `re_raise` and `release_on` — ~135 lines of OS-specific runtime code — now sit in the crate root between `init` and 600 lines of JS-bridge assertions. The next signal-related change lands in the same crowded file.

**Fix:** Move `spawn_signal_cleanup`, `re_raise` and `release_on` into a new `signal` module declared with the same gate as its neighbours at the top of lib.rs (`#[cfg(all(any(unix, windows), debug_assertions))] mod signal;`), keeping one `signal::spawn_cleanup(&cleanup)` call in `setup`. lib.rs drops back under 900 lines and the two watchers become siblings the way `server::unix` and `server::windows` already are.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the diff is what pushes the file past 1000 lines, with the new code it adds

</details>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -93,15 +105,82 @@ fn signal_ends_the_app(signum: i32, tag: &str) {
...
+    // instance registry and makes its instance file predictable here.
+    let dir = std::env::temp_dir().join(format!("tp229-{}", std::process::id()));
+    std::fs::create_dir_all(&dir).expect("create private LOCALAPPDATA");
+    let identifier = "tp229-break";
+    let instance = dir
+        .join("tauri-pilot")
+        .join("instances")
+        .join(format!("{identifier}.json"));
+
+    let mut child = spawn_child_app("LOCALAPPDATA", &dir, identifier);
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Fixed identifier makes the child's named pipe collide across test runs

`LOCALAPPDATA` is scoped by pid, but the pipe `\\.\pipe\tauri-pilot-tp229-break` is machine-global and `bind` opens it with `first_pipe_instance(true)`. A second `cargo test` on the same machine (another worktree) fails the pipe bind before `register_instance` runs. So does a re-run within the 30 s that a stranded `child_app` sleeps after its parent is killed. The test then fails on "child app must register its instance file" after a 10 s poll, a flake unrelated to #229.

**Fix:** Put the pid in the identifier, as `normal_quit_unlinks_instance_file` already does with `com.pilot.issue194-{pid}`, and pass `&identifier` to `spawn_child_app`. `sanitize_identifier` leaves `tp229-break-<pid>` unchanged, so the expected instance file name still matches.

```suggestion
    let identifier = format!("tp229-break-{}", std::process::id());
    let instance = dir
        .join("tauri-pilot")
        .join("instances")
        .join(format!("{identifier}.json"));

    let mut child = spawn_child_app("LOCALAPPDATA", &dir, &identifier);
```

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -93,15 +105,82 @@ fn signal_ends_the_app(signum: i32, tag: &str) {
...
 
-/// Re-execs this test binary as a mock app bound in `socket_dir`.
-fn spawn_child_app(socket_dir: &Path, identifier: &str) -> Child {
-    Command::new(std::env::current_exe().expect("test binary path"))
+/// Ctrl+Break must end a child app and remove its instance file (#229).
+///
+/// Ctrl+Break stands for all four watched events. It is the only one that can
+/// be aimed at one process group: Ctrl+C with a group id is delivered to
+/// nobody, and with none it reaches this test binary and cargo too. Close and
+/// shutdown cannot be generated at all.
```
mpiton: > [!NOTE]
> **🔵 Low · 🧪 Tests** — Close/shutdown path, the reason for the tokio 1.44 floor, has no test

tokio's console handler returns straight away for Ctrl+C and Ctrl+Break but parks forever for `CTRL_CLOSE_EVENT` and `CTRL_SHUTDOWN_EVENT` (`event_requires_infinite_sleep_in_handler`, the #7122 change). Ctrl+Break never enters that branch, so "stands for all four" overstates what the test pins. A tokio regression there, or a dropped `ctrl_close`/`ctrl_shutdown` arm in `spawn_signal_cleanup`, leaves CI green. The floor itself is never exercised either: `check` resolves the newest tokio and `minimal-versions` only runs `cargo check`.

**Fix:** Close and shutdown cannot be generated from a test, so reword the doc comment: Ctrl+Break is a faithful proxy for Ctrl+C (same handler branch), and close/shutdown are covered by manual verification only. Record that manual check in the PR, for example closing the console of `cargo tauri dev` on tokio 1.44 and confirming the instance file is gone.

docs/src/content/docs/guides/plugin-setup.md
```diff
@@ -47,7 +47,15 @@ SIGHUP. Those signals end the process before Tauri emits `RunEvent::Exit`, so
 the plugin removes the socket file itself, restores the default handler and
 re-raises the signal: the app still exits with the usual 128+n status. If your
 app installs its own handler for those signals, it still runs, but the re-raise
-can cut a long graceful shutdown short. Release builds never see any of this.
+can cut a long graceful shutdown short.
+
+On Windows, a debug build watches the console control events instead: Ctrl+C,
+Ctrl+Break, console close and system shutdown. The plugin removes
+`instances\{identifier}.json`, then exits with `STATUS_CONTROL_C_EXIT`
```
mpiton: > [!NOTE]
> **🔵 Low · 📄 Docs** — Guide names instances\{identifier}.json without saying where it lives

The docs site never introduces the Windows instance registry. Section 4 of this guide only covers the Unix socket, and `%LOCALAPPDATA%\tauri-pilot\instances\` appears in no page. A Windows reader checking that the file is gone has no directory to look in.

**Fix:** Give the full path, as the changelog entry and the `init` rustdoc already do.

```suggestion
`%LOCALAPPDATA%\tauri-pilot\instances\{identifier}.json`, then exits with `STATUS_CONTROL_C_EXIT`
```

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.85</sub>

CHANGELOG.md
```diff
@@ -90,6 +90,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- On Windows, quitting the app with Ctrl+C or Ctrl+Break, or by closing the
+  console running `cargo tauri dev`, now removes
+  `%LOCALAPPDATA%\tauri-pilot\instances\{identifier}.json`. Those console
+  control events end the process before Tauri emits `RunEvent::Exit`, so the
+  registry guard never dropped and the CLI kept listing an app that was gone.
```
> **🔵 Low · 📄 Docs** — Changelog says the CLI kept listing a dead app, but the resolver skips dead pids

`resolve_socket_windows` drops every instance entry whose `pid` fails `is_pid_alive` (`crates/tauri-pilot-cli/src/main.rs:1591`). It is the only reader of the `instances` directory, and the CLI has no command that lists instances. A leftover file only mattered once Windows reused the pid, so the release notes describe a symptom users never saw.

**Fix:** State the real effect. The file stayed in `%LOCALAPPDATA%\tauri-pilot\instances\`, and the CLI could resolve to its dead pipe once another process reused the pid. The `spawn_signal_cleanup` doc comment (`lib.rs:188`) says the file keeps listing the app, which is accurate and can stay.

<details>
<summary>🔎 Details</summary>

- **Why it belongs to this PR:** the sentence is added by this diff

</details>

<sub>🤖 <b>mpiton-review</b> · documentation · confidence 0.8</sub>

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -93,15 +105,82 @@ fn signal_ends_the_app(signum: i32, tag: &str) {
...
+#[cfg(windows)]
+#[test]
+fn ctrl_break_ends_the_app_and_removes_the_instance_file() {
+    use windows::Win32::Foundation::STATUS_CONTROL_C_EXIT;
+    use windows::Win32::System::Console::{CTRL_BREAK_EVENT, GenerateConsoleCtrlEvent};
+
+    // A private LOCALAPPDATA of our own keeps the child out of the real
+    // instance registry and makes its instance file predictable here.
+    let dir = std::env::temp_dir().join(format!("tp229-{}", std::process::id()));
+    std::fs::create_dir_all(&dir).expect("create private LOCALAPPDATA");
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> After an interrupted run, this fixed PID-based directory can be reused with stale instance or named-pipe state

Fixed in `33a955e`. The dir is wiped with `remove_dir_all` before `create_dir_all` (`tests/signal_exit.rs:127`), so a leftover `.json` from a run that drew the same pid can't pass for the child's registration. The pipe half is the pid now in the identifier (`:131`): the pipe name is machine-wide, the directory never scoped it. Skipped the atomic uniqueness check, `temp_dir()` on Windows is per-user so nobody else can pre-plant the name.

docs/src/content/docs/guides/plugin-setup.md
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> The documented Windows instance path omits `%LOCALAPPDATA%\tauri-pilot\`

Fixed in `33a955e`, suggestion applied as is.

crates/tauri-plugin-pilot/tests/signal_exit.rs
```diff
@@ -93,15 +105,82 @@ fn signal_ends_the_app(signum: i32, tag: &str) {
...
+    let identifier = "tp229-break";
+    let instance = dir
+        .join("tauri-pilot")
+        .join("instances")
```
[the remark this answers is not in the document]
mpiton: @cubic-dev-ai

> This integration test duplicates the private instance-file naming rule from `server::windows::instance_file_path`.

Not changing this. `%LOCALAPPDATA%\tauri-pilot\instances\{identifier}.json` isn't private to the plugin: the CLI rebuilds the same path in `resolve_socket_windows` (`crates/tauri-pilot-cli/src/main.rs:1559`), and it's in the `init` rustdoc and the changelog. A rename there has to break something loudly, and this test is a fine place for it. A module-owned helper would also have to be `pub` for a `tests/` file to reach it, which widens the plugin's API for one test. The Unix test in the same file does the same with `tauri-pilot-{identifier}.sock` (`:80`) for that reason. I'd revisit if the path ever stops being shared with the CLI.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > Host console handlers installed before the plugin never run; docs say they still run

Right, checked against tokio 1.44.0 `signal/windows/sys.rs:140`: the handler returns 1 once a listener got the event. Took the docs route in `33a955e`. The `init` rustdoc (`lib.rs:161`), the guide and the changelog now say a handler registered before `setup` isn't called in debug builds, one registered after runs first, and a `tokio::signal` listener runs alongside. `release_on`'s doc says the same.

Left the raw `SetConsoleCtrlHandler` routine out of this PR. Returning FALSE after the release changes behavior the other way: a host handler that claims Ctrl+C and keeps the app alive would leave it running with no instance file. It also needs a static to reach the guard from an `extern "system"` fn. Worth its own issue if leaked sidecars in dev turn out to bite.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > `ctrl_shutdown` watcher cannot fire in a Tauri app, so system shutdown still leaks the file

Fixed in `33a955e`. Dropped the `ctrl_shutdown` watcher, and "system shutdown" is gone from the changelog, the `init` rustdoc and the guide. All three now say shutdown and logoff leave the file and that the CLI's pid check is the fallback (`signal.rs:80`). The tokio 1.44 floor stays, `ctrl_close` still needs the parked handler. No `WM_ENDSESSION` hook here, that's a separate feature.

crates/tauri-plugin-pilot/Cargo.toml
mpiton: > Windows dev-dependency declares only one of the three features the test uses

Fixed in `33a955e`, suggestion applied. `cargo clippy --target x86_64-pc-windows-gnu -p tauri-plugin-pilot --all-targets -- -D warnings` passes.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > Console event before the bind releases an empty slot, then exits with the file still there

The window is real, the proposed fix isn't safe. Unregistering by identifier breaks the second-instance case from #152: instance B has the same identifier, its bind fails on `first_pipe_instance(true)`, its slot stays `Empty`. Ctrl+C in B's console would then delete A's live entry, and the CLI loses a running app. That's worse than a leftover file the CLI already skips through `is_pid_alive`. The guard is what ties removal to the process that registered.

`33a955e` rewrites the `setup` comment (`lib.rs:252`), which did promise more than the `process::exit` path gives. It now names the window and why it stays open. A pid-checked variant (read the entry, remove only if `pid == process::id()`) would be safe if this ever shows up in practice.

crates/tauri-plugin-pilot/src/lib.rs
mpiton: > lib.rs crosses 1000 lines: the console-event watcher lands in the crate root

Done in `33a955e`. Both watchers, `re_raise` and `release_on` are in `src/signal.rs`, called as `signal::spawn_cleanup(&cleanup)`. lib.rs is at 901 lines. One difference from the suggestion: the `mod signal;` gate also has `not(target_os = "android")`. Android is `unix`, so with the neighbours' gate the Unix watcher compiles there with no caller and trips `dead_code`.

crates/tauri-plugin-pilot/tests/signal_exit.rs
mpiton: > Fixed identifier makes the child's named pipe collide across test runs

Fixed in `33a955e`, suggestion applied: `tp229-break-{pid}`, passed as `&identifier`.

crates/tauri-plugin-pilot/tests/signal_exit.rs
mpiton: > Close/shutdown path, the reason for the tokio 1.44 floor, has no test

Reworded in `33a955e` (`tests/signal_exit.rs:110`): Ctrl+Break stands for Ctrl+C only, same handler branch, and console close is checked by hand. Shutdown is out of the picture now that the watcher is gone. The manual check on tokio 1.44 is still to do: run `cargo tauri dev` on Windows, close the console, confirm the instance file is gone.

docs/src/content/docs/guides/plugin-setup.md
mpiton: > Guide names instances\{identifier}.json without saying where it lives

Fixed in `33a955e`, suggestion applied.

CHANGELOG.md
mpiton: > Changelog says the CLI kept listing a dead app, but the resolver skips dead pids

Fixed in `33a955e`. The entry now says the file stayed behind and only mattered once another process drew the pid, at which point the CLI could resolve to the dead pipe.

